PR #57, opened 2015-01-02
Implement large object support

titanous:
This patch implements [large object](http://www.postgresql.org/docs/current/static/largeobjects.html) support. It is a verbatim port from our pq implementation of the same that we've been using for a while.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -50,6 +50,7 @@ type Conn struct {
 	causeOfDeath       error
 	logger             Logger
 	mr                 msgReader
+	fp                 *fastpath
 }
 
 type PreparedStatement struct {
--- a/fastpath.go
+++ b/fastpath.go
@@ -1,0 +1,110 @@
+package pgx
+
+import (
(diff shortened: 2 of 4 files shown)


titanous:
Thanks for merging!

jackc:
Thank you for your contribution! I haven't had opportunity to use large objects in PostgreSQL yet, so I wasn't going to implement something without having first-hand knowledge of how and why to use it. It's good to have the interface designed by someone who will use it.

I was curious about the fast path code. It looks like it is using the function call sub protocol (http://www.postgresql.org/docs/9.4/static/protocol-flow.html#AEN102867) which is marked as legacy. Have you found it faster than the documentation's suggestion to use prepared statements?

titanous:
I haven't actually tried the prepared statement approach, as I don't think there is any documentation, and all of the implementations I looked at (including libpq) use the fastpath protocol. I currently don't have any time to experiment with it, but I'd be interested in the results.

PR #70, opened 2015-03-27
added GetValue() accessor for nullable structs for db to code generators

silviucm:
This pull request might help any template-based code generation libraries that use pgx as the underlying db foundation 

The problem arises from the fact that in the nullable structs, there is no common "Value" property.
The NullString value container is called String, the NullBool one is called Bool, and so forth.

The GetValue() method, applicable to nullable structs allows database-to-struct code generators using the pgx library to simplify their code generation methodology.

After performing a scan into a &nullableNullTime variable, the generator can simply rely on this code inside the template file:
structInstance.WhateverDateTimeField = nullableNullTime.GetValue()  
(instead of doing complicated if - else sequences at template level)

### DIFF
--- a/values.go
+++ b/values.go
@@ -119,6 +119,8 @@ type NullFloat32 struct {
 	Valid   bool // Valid is true if Float32 is not NULL
 }
 
+func (n *NullFloat32) GetValue() float32 { return n.Float32 }
+
 func (n *NullFloat32) Scan(vr *ValueReader) error {
 	if vr.Type().DataType != Float4Oid {
 		return SerializationError(fmt.Sprintf("NullFloat32.Scan cannot decode OID %d", vr.Type().DataType))
@@ -157,6 +159,8 @@ type NullFloat64 struct {
 	Float64 float64
 	Valid   bool // Valid is true if Float64 is not NULL
 }
+
+func (n *NullFloat64) GetValue() float64 { return n.Float64 }
 
 func (n *NullFloat64) Scan(vr *ValueReader) error {
 	if vr.Type().DataType != Float8Oid {
@@ -197,6 +201,8 @@ type NullString struct {
 	Valid  bool // Valid is true if Int64 is not NULL
 }
 
+func (n *NullString) GetValue() string { return n.String }
+
 func (s *NullString) Scan(vr *ValueReader) error {
 	// Not checking oid as so we can scan anything into into a NullString - may revisit this decision later
 
@@ -230,6 +236,8 @@ type NullInt16 struct {
 	Int16 int16
 	Valid bool // Valid is true if Int16 is not NULL
 }
+
+func (n *NullInt16) GetValue() int16 { return n.Int16 }
 
 func (n *NullInt16) Scan(vr *ValueReader) error {
 	if vr.Type().DataType != Int2Oid {
@@ -270,6 +278,8 @@ type NullInt32 struct {
 	Valid bool // Valid is true if Int64 is not NULL
 }
 
+func (n *NullInt32) GetValue() int32 { return n.Int32 }
+
 func (n *NullInt32) Scan(vr *ValueReader) error {
 	if vr.Type().DataType != Int4Oid {
 		return SerializationError(fmt.Sprintf("NullInt32.Scan cannot decode OID %d", vr.Type().DataType))
@@ -308,6 +318,8 @@ type NullInt64 struct {
 	Int64 int64
 	Valid bool // Valid is true if Int64 is not NULL
 }
+
+func (n *NullInt64) GetValue() int64 { return n.Int64 }
 
 func (n *NullInt64) Scan(vr *ValueReader) error {
 	if vr.Type().DataType != Int8Oid {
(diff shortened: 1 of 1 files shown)


jackc:
The code change seems reasonable enough, but I have a few questions.

These are strongly-typed accessors, so they can't be used as a Go interface. Is this change simply to work around the differences in names? If all the fields were renamed to "Value" would that meet your needs? (Obviously, we can't do that without breaking backwards compatibility, but I am thinking about an eventual v3 release.)

Also, as far as this code:

```go
structInstance.WhateverDateTimeField = nullableNullTime.GetValue()
```
Do you have a larger example of what you are envisioning? I would see an automatic mapping of db to struct having the record struct have a nullable struct as the field if the field is legitimately nullable. If it is not I would directly use the underlying type.

silviucm:
Hi Jack,

Firstly, to answer the "fields renamed to Value" question: yes, that would work for me. I tried to offer those methods as an alternative simply to keep your library intact (compatibility-wise).

For the more general purpose of this, the whole story is revolving around me trying to create a high-productivity pgsql to struct generator for my own purposes (it's not idiomatic "Go", I am making aggressive use of "fake" sub-packages / singleton variables / singleton collection variables, so I am definitely not going to advertise it publicly).
The project is ambitious, in the sense that I am trying to generate as much as possible without using any reflection. I am getting tables, views (including materialized), fields, PKs, FKs, constraints, I am generating Select, SelectAll, Insert, Updates, Transaction wrappers (and rudimentary caching mechanisms).
All this is predicated on keeping heavily-joined queries inside views at db level, and me just hitting the db via a generated view struct via your library with minimal parametric complexity on the Go code side.
 
I am currently using pgx (very impressed by your avoidance of using reflect package in any way, this was the primary reason) in this project pgtogogen (in my repositories) and, after some deliberations, I decided to go with this approach (structure generation, after loading the necessary Postgresql schema info into the template):

type {{.GoFriendlyName}} struct {
	{{range .Columns}}{{if ne .DbComments ""}}/* {{.DbComments}} */
	{{.GoName}} {{.GoType}} // IsPK: {{.IsPK}} , IsCompositePK: {{.IsCompositePK}}, IsFK: {{.IsFK}}
	{{if .Nullable}}{{.GoName}}_IsNotNull bool // if true, it means the corresponding field does not currently carry a null value{{end}}
...
etc
...

I opted to go with "traditional" Go types (string, time.Time, etc) and the interaction with your library would be done by a boolean control variable (fieldName)_IsNotNull.
The primary reason was that your library could be decoupled at any point and replaced with another one. To be honest, at this stage, it's highly unlikely, but I don't want to change what I've done so far.
While there is an element of redundancy in this approach, the secondary reason is that I wanted to keep go-friendly types, and be able to reuse those generated structs in different situations

I am also generating setter methods which, at "Select" time would populate that control boolean variable appropriately, based on your "Valid" bool in the nullable types.
The generator, based on the field's nullability creates a setter method that has one or two parameters (the 2nd one being the Valid-like bool).

I completely understand if you object to strongly-type accessors, and I would just keep these methods in my fork, until you release the next version.
In my opinion those accessors are not too much of a pest, it's just that you'd have to make sure to have one of them in any future nullable types.

Cheers
silviu

jackc:
I think I found a for you to get GetValue() without breaking compatibility, adding to the interface of pgx, or you having to fork pgx.

The trick is that types in Go can be copied/renamed.

```go
package yourpackage

import (
	"github.com/jackc/pgx"
)

type NullString pgx.NullString

func (n *NullString) Scan(vr *pgx.ValueReader) error {
	return (*pgx.NullString)(n).Scan(vr)
}
func (n NullString) FormatCode() int16 {
	return pgx.NullString(n).FormatCode()
}
func (n NullString) Encode(w *pgx.WriteBuf, oid pgx.Oid) error {
	return pgx.NullString(n).Encode(w, oid)
}
``` 

The above gives you a copy of NullString. But since it is defined in your package you can add methods to it.

```go
func (n NullString) GetValue() string { return n.String }
```

Now you can just use your version of Null* types and have GetValue(). 

By the way, I like the idea of statically generating mappings. I've found hand-writing SQL is great for non-CRUD operations, but a real drag for traditional CRUD. Been thinking about approaches to ease the pain for a while.

silviucm:
Your idea is actually better than mine - thanks - I think I will go ahead and incorporate it. In theory, keeping this stuff on my side should give me more flexibility going forward.
Feel free to invalidate this pull request if need be.

Jack, I wanted to ask you something regarding QueryRow (instead of opening another issue / PR).
I currently implemented a Single(condition etc etc) method. 
My particular intention is that Single would return a dedicated error if more rows than just one satisfying the condition are found.
I am not exactly sure, but is your QueryRow more or less a wrapper over Query ?
I currently use Query plus a counter variable, and if that var reaches row number two I close the Rows and exit with that error.

With the specific purpose of returning an error if more rows than one are found: Is there another way to get the rowcount directly after running the query, without doing a secondary call to Next() ? 
(I am assuming "no" for now, as I noticed that "rowcount" field in your Row struct is incremented at the "Next" method only) ?

Cheers, and thanks for the note about my library. At some point in April I'll write some documentation and let you know. If you can hold your nose about its hyper-verbosity and idiosyncrasies, feel free to use / repurpose / optimize it if you have any need for it.
(Next to do for me: try to identify all pg functions inside the database that return a specific "SetOf" - table or view - and create the Go skeleton for them, to return Go structures)

jackc:
QueryRow is just a thin wrapper over Query. I don't believe there is any way to get the row count without actually counting the received rows. This limitation exists all the way down to the PostgreSQL wire protocol. I don't think there is any way around it.

PR #75, opened 2015-04-07
Fix decoding of ARRAY[]::text[]

debackerl:
The original version could not decode ARRAY[]::text[]. When an empty array was detected, decode1dArrayHeader was not reading enough values off the socket, and subsequent values were incorrectly read.

### DIFF
--- a/values.go
+++ b/values.go
@@ -1102,15 +1102,16 @@ func encodeTimestamp(w *WriteBuf, value interface{}) error {
 
 func decode1dArrayHeader(vr *ValueReader) (length int32, err error) {
 	numDims := vr.ReadInt32()
-	if numDims == 0 {
-		return 0, nil
-	}
-	if numDims != 1 {
+	if numDims > 1 {
 		return 0, ProtocolError(fmt.Sprintf("Expected array to have 0 or 1 dimension, but it had %v", numDims))
 	}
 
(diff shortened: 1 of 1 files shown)


jackc:
Do have a failing test for this? I wanted to add a failing test before I merged in your fix, but I was unable to get it to fail. See https://github.com/jackc/pgx/blob/master/values_test.go#L226 for the empty array tests I wrote.

debackerl:
Hi, I'm not sure how to update this pull request with an additional pull request. Anyway, here is a gist with failing test:

https://gist.github.com/debackerl/041e4d986f2fb5c18677

--- FAIL: TestEmptyArrayDecoding (0.02s)
	values_test.go:267: error reading array: Received an invalid size for an int4: 28
FAIL

This test is failing in a nice way. But I had a case where I created a connection pool, started a transaction, ran a Query similar to this one, getting a similar error, but the "defer tx.Rollback()" would loop infinitely, having 100% CPU use. I'll try to reproduce it, but it's going to be hard.

Thx for your support!

jackc:
I merged the test in and then put your fix in on top. Thanks!

debackerl:
Great thx! I'll also try to reproduce the infinite loop issue, but that one is tricky. I believe it may still happen if somebody implements a custom Scanner and reads too little or too many bytes via the reader, but even then, it did not always occur.

PR #77, opened 2015-04-09
Unit test demonstrating infinite loop

debackerl:
Hi, I have just created a unit test demonstrating an infinite loop during cleanup in a defered function. The CPU will just sit at 100% use. I know this is a special case, but it's would be nice if a buggy Scanner doesn't make cleanup functions like rows.Close run indefinitely. Sadly, I'm not sure how to fix/improve this.

It would be awesome if you could do something about this :-)

### DIFF
--- a/values_test.go
+++ b/values_test.go
@@ -223,6 +223,37 @@ func TestArrayDecoding(t *testing.T) {
 	}
 }
 
+type buggyScanner struct {}
+
+func (*buggyScanner) Scan(r *pgx.ValueReader) error {
+	r.ReadInt32()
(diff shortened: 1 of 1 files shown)


jackc:
Thanks for this test case. It was quite an interesting bug to track down. It turns out the problem would only manifest when a value was not entirely read on the last column of the select and there were following rows. The ultimate issue was the ValueReader that stores the number of bytes left in the value was not being cleared when a new row was read.

Fixed in 879ca97228dc4ad2f2901dd2a16ed2bb61dfe6cb.

PR #78, opened 2015-04-10
add Decoder interface (like Scanner)

deoxxa:
this change allows a type to implement a Decode(r *ValueReader) method
instead of a Scan(r *ValueReader) method, in case that type already has
a Scan(v interface{}) method designed to be used by the built-in sql
package.

perhaps the Scanner interface could be deprecated in future, to avoid
possible confusion with the interface of the same name in the built-in
sql package?

### DIFF
--- a/query.go
+++ b/query.go
@@ -280,6 +280,11 @@ func (rows *Rows) Scan(dest ...interface{}) (err error) {
 			if err != nil {
 				rows.Fatal(err)
 			}
+		case Decoder:
+			err = d.Decode(vr)
+			if err != nil {
+				rows.Fatal(err)
+			}
 		default:
 			rows.Fatal(fmt.Errorf("Scan cannot decode into %T", d))
 		}
--- a/values.go
+++ b/values.go
@@ -90,6 +90,16 @@ type Scanner interface {
 	Scan(r *ValueReader) error
 }
 
+// Decoder is an interface used to decode values from the PostgreSQL server.
+type Decoder interface {
+	// Decode MUST check r.Type().DataType (to check by OID) or
+	// r.Type().DataTypeName (to check by name) to ensure that it is scanning an
+	// expected column type. It also MUST check r.Type().FormatCode before
+	// decoding. It should not assume that it was called on a data type or format
+	// that it understands.
+	Decode(r *ValueReader) error
+}
+
 // Encoder is an interface used to encode values for transmission to the
 // PostgreSQL server.
 type Encoder interface {


jackc:
Not sure how I feel about this. I purposely have used the same names for the same concepts as database/sql where possible to make it as easier for people familiar with database/sql to learn and use pgx.

However, that does have the downside you bring up of method name collisions for custom types that try to support both systems...

Could the same goal be achieved by redefining the type for one or the other? e.g.

```go
type NullPoint struct {
	X, Y  float64 // Coordinates of point
	Valid bool    // Valid is true if not NULL
}

func (p *NullPoint) Scan(vr *pgx.ValueReader) error {
// ...
}

type NullPointStdlib NullPoint

func (p *NullPointStdlib) Scan(v interface) error {
// ...
}
```

Then you have a single underlying type, and can type convert to get whichever method set you want.

kdar:
Then could pgx/stdlib have all the Null* types with the sql.Scanner interface? I was confused for a while when I was trying to use sqlx with pgx, and it was failing to scan pgx.NullTime.

jackc:
As of 9f9a9779ac794e885f6fbb04935724f03149f092, pgx can use objects that implement the sql.Scanner interface. This should make it easier to interop with sqlx.

bdittmer:
I'm still having issues with sqlx and the pgx Null* types. When trying to load a float32 into pgx.NullFloat32 using the sqlx StructScan method I'm getting:

```
panic: sql: Scan error on column index 3: unsupported Scan, storing driver.Value type float32 into type *pgx.NullFloat32
```

Some simplified code:
```
type DummyData struct {
       ID int `db:"id"`
       One int `db:"one"`
       Two int `db:"two"`
       Three pgx.NullFloat32 `db:"three"`
}

type dataStore struct {
	db *sqlx.DB
}

func NewDataStore() (dataStore, error) {
	db, err := sqlx.Connect("pgx", config.DbConnString())
	if err != nil {
		panic(err)
	}

	return &dataStore{
		db: db,
	}, nil
}

func (s *dataStore) Get(id int) DummyData {
	rows, err := s.db.Queryx(myQuery, id)
	if err != nil {
		panic(err)
	}

        var dd DummyData
	for rows.Next() {
		if err := rows.StructScan(dd); err != nil {
			panic(err)
		}
	}

	return dd, nil
}
```

jackc:
I don't think sqlx understands pgx Scanner types. But pgx does understand database/sql Scanner types. Try using the database/sql.NullFloat instead of pgx.NullFloat32.

PR #79, opened 2015-04-13
Support nil driver.Value

archSeer:
If the driver.Valuer ever returns nil for a custom type, `valueToInterface` will return a panic: `interface conversion: interface is nil, not interface {}`. This patch adds support for nil values.


jackc:
Thanks!

PR #80, opened 2015-04-18
Support using a custom dialer

lmars:
For example I may want to use a dialer which retries transient network errors (e.g. DNS issues).


jackc:
Thanks!

PR #83, opened 2015-07-14
format type added (time); json decode function added; and flexible scan to values 'Int*' format;

caanvaga:


jackc:
Thanks for the PR. I hope to merge in some of these features, but before I need to make sure I understand the changes and ensure that the test suite is updated and passes.

I made some line by line notes, but as far as the overall PR, it's hard to understand how the changes relate to each other because there seem to be multiple entirely separate features mixed together and there are no tests to document the intended use.

The PostgreSQL time type is a time of day OR an amount of time. Parsing it into a Go time.Time is suspect. What happens when the PostgreSQL time was "36:00:00"? A Go time.Duration would be better suited, but even that is not necessarily a good match semantically.

If I understand correctly, the JSON support is designed to read the JSON from PostgreSQL directly into a PostgreSQL struct. That's a really nice feature. However, I'd like to see some tests around that functionality to: 1. Ensure the new feature works and 2. Ensure old behaviour of reading into a string or []byte still works.

PR #84, opened 2015-08-19
Add support for specifying sslmode in connection strings

RickSnyder:
- In conn.go, setting the TLSConfig, UseFallbackTLS, and FallbackTLSConfig on ConnConfig has been moved into a separate function called configSSL, and ParseEnvLibpq has been updated to call configSSL.
- In conn.go, ParseURI now allows ability to specify sslmode parameter on the connection URL, leveraging configSSL method.
- In conn.go, ParseDSN now allows ability to specify sslmode parameter on the connection string, also leveraging configSSL method.
- Tests have been updated/added to accordingly.
- Tweaked a few comments to illustrate ability to use sslmode parameter.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -296,6 +296,11 @@ func ParseURI(uri string) (ConnConfig, error) {
 	}
 	cp.Database = strings.TrimLeft(url.Path, "/")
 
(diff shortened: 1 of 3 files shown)


jackc:
Thanks!

PR #86, opened 2015-08-26
Support decoding of TimestampOid in stdlib driver

samuel:
Seems safe as it's pretty much equivalent to TimestampTzOid


jackc:
Thanks!

PR #88, opened 2015-09-04
Add JSON/JSONB support

josephglanville:
Somewhat inspired by #83

### DIFF
--- a/query.go
+++ b/query.go
@@ -284,7 +284,14 @@ func (rows *Rows) Scan(dest ...interface{}) (err error) {
 				rows.Fatal(err)
 			}
 		default:
-			rows.Fatal(fmt.Errorf("Scan cannot decode into %T", d))
+			switch vr.Type().DataType {
+			case JsonOid:
+				decodeJson(vr, &d)
+			case JsonbOid:
+				decodeJson(vr, &d)
+			default:
+				rows.Fatal(fmt.Errorf("Scan cannot decode into %T", d))
+			}
 		}
 
 		if vr.Err() != nil {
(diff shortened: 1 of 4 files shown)


query.go
```diff
@@ -284,7 +284,14 @@ func (rows *Rows) Scan(dest ...interface{}) (err error) {
 				rows.Fatal(err)
 			}
 		default:
-			rows.Fatal(fmt.Errorf("Scan cannot decode into %T", d))
+			switch vr.Type().DataType {
```
josephglanville: Is there a better way to do this? It seems hacky but I couldn't find a reasonable way to do it other than matching on interface{} and then falling through into the default case above (which was even more hacky).

josephglanville:
I would like to add JSON/JSONB array support too and look for how these types work in binary encoding.
Any pointers very welcome.

josephglanville:
Will also need a means of disabling the JSON/JSONB tests on older PG versions that don't support either/or those types.

jackc:
I refactored the the Scan method a bit, but it is still complicated. By changing the precedence of JSON decoding vs. normal array decoding we can now decode JSON arrays directly into Go slices.

I added filtering to skip JSON/JSONB tests so it now passes on Travis.

I don't think binary support for JSONB makes sense at this time. The binary format is simply a 1 byte version header, then the document in text.

josephglanville:
Wow awesome.

Yeah I wasn't sure about the binary encoding format, I hadn't looked at the pg sources for how that was done for JSONB yet.

Learnt a few things from how you refactored these changes, thanks very much.

PR #89, opened 2015-09-06
Add support for pointers to pointers

titanous:
Using types like `**string` allows the inner pointer to be `nil`’ed out, avoiding the need for `NullX` types.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"os/user"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


jackc:
I was initially concerned that pointer to pointer would perform more poorly than `NullX`. So I wrote some benchmarks, and it appears that while it does cause additional mem allocs, overall performance is unaffected or even improved slightly (presumably due to avoiding the Scanner interface function call).

Rebased and merged. Thanks.

titanous:
Neat, thanks! I was wondering about the performance but didn't have time to benchmark.

PR #90, opened 2015-09-06
Add support for encoding to UUID from string types

josephglanville:
Decoding works out of the box as the default is to try decode to string.
However encoding requires handling as a string like type.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -776,7 +776,7 @@ func (c *Conn) sendPreparedQuery(ps *PreparedStatement, arguments ...interface{}
 				err = encodeFloat4(wbuf, arguments[i])
 			case Float8Oid:
 				err = encodeFloat8(wbuf, arguments[i])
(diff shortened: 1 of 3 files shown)


jackc:
I tried running the test you wrote, but without the code changes, and the test still works. So the test is not demonstrating any new behavior. What new behavior are you trying to allow?

josephglanville:
Yeah my test case is wrong then. The error I was hitting was "Cannot encode *string into oid 2950 - *string must implement Encoder or be converted to a string" which is fixed by this change.
I will have another go at writing a minimal test case.

josephglanville:
@jackc I think that demonstrates the behaviour I need, specifically being able to encode from *string into uuid.

jackc:
Thanks. Merged. I built upon this a little bit more to support encoding *string to any non-varchar/text PostgreSQL type.

PR #91, opened 2015-09-08
Fix feature attribution in CHANGELOG

titanous:
Just noticed this while reading through.


jackc:
Sorry about that. The commits were right next to each other. Must have copy and pasted the wrong one.

titanous:
No worries, thanks for maintaining this library!

PR #92, opened 2015-09-10
ConnPool.Begin/BeginIso will retry if they Acquire a dead connection

josephglanville:
This implements a naive retry in Begin/BeginIso which will attempt the exhaust the connection pool trying to find a live connection.

The loop exits when:
* A live connection is found and a transaction successfully created
* Attempting to create the transaction returns an error which is not ErrDeadConn
* Attempting to acquire a connection from the pool fails (possibly due to ErrDeadConn if the target server is still down)


jackc:
I added a test and made a few tweaks to this. The main change is checking connection IsAlive instead of looking for ErrDeadConn on Conn.Begin error. The reason is ErrDeadConn is the error returned when calling a connection that is already known to be dead. But when the connection actually dies, then the real error is returned (something like: write unix @->/var/run/postgresql/.s.PGSQL.5432: write: broken pipe).

Issue #93, opened 2015-09-14
Possible rxMsg bug


josephglanville:
Hi,

I have been trying to track down a mysterious hang and I think it may be a pgx bug.
Here is a stack trace of the goroutine that is hung:

```
goroutine 97 [IO wait, 18 minutes]:
net.(*pollDesc).Wait(0xc20813a990, 0x72, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc20813a990, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/net/fd_poll_runtime.go:89 +0x43
net.(*netFD).Read(0xc20813a930, 0xc20819a000, 0x2000, 0x2000, 0x0, 0x7f028754edb0, 0xc20800b8b0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/net/fd_unix.go:242 +0x40f
net.(*conn).Read(0xc208036e08, 0xc20819a000, 0x2000, 0x2000, 0x0, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/net/net.go:121 +0xdc
bufio.(*Reader).Read(0xc20813e900, 0xc20819a000, 0x2000, 0x2000, 0x5e, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/bufio/bufio.go:164 +0x13a
io.(*LimitedReader).Read(0xc2081b8060, 0xc20819a000, 0x2000, 0x2000, 0x5e, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/io/io.go:408 +0xce
io/ioutil.devNull.ReadFrom(0x0, 0x7f0287550bd8, 0xc2081b8060, 0x1ea, 0x0, 0x0)
        /home/ubuntu/go/src/github.com/flynn/flynn/util/_toolchain/go/src/io/ioutil/ioutil.go:151 +0xb1
io/ioutil.(*devNull).ReadFrom(0xc20800a2f0, 0x7f0287550bd8, 0xc2081b8060, 0x7f0287550c28, 0x0, 0x0)
        <autogenerated>:9 +0xbb
[...]

PR #95, opened 2015-09-17
Track Listen/Unlisten and clear subscriptions on conn release

josephglanville:
This avoid releasing connections back into the pool that have active listen subscriptions.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -52,6 +52,7 @@ type Conn struct {
 	config             ConnConfig        // config used when establishing this connection
 	TxStatus           byte
 	preparedStatements map[string]*PreparedStatement
+	channels           map[string]struct{}
 	notifications      []*Notification
 	alive              bool
 	causeOfDeath       error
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -572,6 +574,36 @@ func (c *Conn) Deallocate(name string) (err error) {
 // Listen establishes a PostgreSQL listen/notify to channel
 func (c *Conn) Listen(channel string) (err error) {
 	_, err = c.Exec("listen " + channel)
+	if err != nil {
+		return
+	}
+	var s struct{}
+	c.channels[channel] = s
```
titanous: Can replace `s` with `struct{}{}`

jackc:
I'm not convinced that this is the underlying cause of #93, but I think this is a reasonable idea. Though to some degree, whenever you are using a connection pool, the client is responsible for not damaging / altering the state of the connection. There are all sorts of ```set``` commands that could be done that would have negative affects that we are unable to detect. Still, given that this is simple to do, may as well prevent foot-guns when we can.

I think the implementation can be simplified.

```sql
unlisten *; -- unlistens from all
```

Still would need to track if there was a listener, because we don't want the performance impact of calling it on every connection release if unneeded.

jackc:
Merged via cherry pick.

Issue #110, opened 2015-12-04
Means for dealing with session level state problems.


josephglanville:
Hi @jackc,

I encountered a problem with prepared queries and migrations that led me to think about how to deal with this class of problems with pgx.
Specifically my problem is we pre-prepare all queries on all connections using the `ConnPool` `AfterConnect` hook.
If then another instance of my program runs a database migration that alters the type of a column these prepared queries will become invalid if they make use of this column and return:
```
ERROR: cached plan must not change result type (SQLSTATE 0A000)
```

The issue here is that this is a session state error. To correct this I would need to either deallocate that query and re-prepare it or close that connection and allow the connection pool to establish a new one.
pgx currently doesn't provide any good interfaces for accomplishing this for queries that occur within a transaction. I have wrapped the `ConnPool` and can handle it on standard `Exec`/`Query` but the Tx object takes full control over the connection and wants to call `Release` on it itself.

Do you have any ideas on how we could incorporate a hook or something that would be called on `Exec`/`Query` on a `Tx` that would then store a flag to call `Close` on the connection before `Release`? (because `Rollback` will probably be called after returning the `Error` anyways)

josephglanville:
I pushed up a proposed solution here: #111

I then use this hook to check if the error is a one which I consider to be a session state problem and if so call `conn.Close`, which will cause new connections to be spawned as the query is retried.

jackc:
This is a tricky problem.

A connection-wide on error handler has a few downsides. 1. We need to be sure to call it every place an error could occur not just Query and Exec (like ListenForNotification?) 2. It moves the error handling away from where the error occurred -- it feels like an exception handler -- not very Go-like.

That said, I don't have any ideas that I really like better either. Here's some of the things I've been considering though:

1. Prepared statements on ConnPool. This was requested back in #68. If this were in place it might be a simple solution to re-prepare statements that have changed.
2. Exposing Conn on Tx. Don't really like this because the Conn and Tx state could really get confused.
3. Exposing a CloseConn method on Tx. Better than the above idea, but still feels a little off.
4. Adding a Reset method to ConnPool that would close all connections and reestablish them as needed. This might be fairly clean.

Also, just out of curiosity, what sort of column type changes can be fixed by a re-prepare, but don't require a different Go type to be Scan'ed or Exec'ed? I would expect you would need to change the Go program as well.

josephglanville:
[...]

PR #111, opened 2015-12-04
Add OnError hook to Conn

josephglanville:
Proposed solution to #110


josephglanville:
Decided this is not useful in it's current form, may explore solutions in this space later.

PR #121, opened 2016-03-04
Added uint32 and uint64 decoders

utrack:
There are `uint32` and `uint64` encoders, but no decoder.
While there's no distinct `uint`-like type in the Postgres, this decoder would do the 'best-effort' conversion and work right for the majority of cases (except negative values ofc, but I guess it's safe to assume that the user knows what he's doing if he plays with `uint`).

### DIFF
--- a/values.go
+++ b/values.go
@@ -687,25 +687,35 @@ func Decode(vr *ValueReader, d interface{}) error {
 	case *int32:
 		*v = decodeInt4(vr)
 	case *uint32:
+		var valInt int32
 		switch vr.Type().DataType {
 		case Int2Oid:
-			*v = uint32(decodeInt2(vr))
+			valInt = int32(decodeInt2(vr))
 		case Int4Oid:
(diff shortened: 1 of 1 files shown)


jackc:
Given that we are already handling auto-converting any Go int type to any PostgreSQL int type, I think that doing the same for decoding seems like a reasonable idea. That said, I don't think it is fair to allow the possibility of silent data loss. The existing encode path checks to see if there would be data loss in the conversion and returns an error. I think we need to do the same for decoding.

utrack:
@jackc Good point! Could you show me what error gets returned if there's data loss (if there's any standard one) so I could use it with `uint` as well?

jackc:
At the moment it is just a `fmt.Errorf`. See https://github.com/jackc/pgx/blob/master/values.go#L943 for example.

utrack:
@jackc Thanks! Checks enabled.

jackc:
Thanks!

PR #124, opened 2016-03-08
Add ConnAllocTimeout and QueryExecTimeout

konstantin-dzreev:


konstantin-dzreev:
P.S. I do not think that QueryExecTimeout works for queries because a query does not seem to run until the very first Next() call.

jackc:
Thanks! I think this will be a good addition to pgx. I do have a few thoughts to go over before merging though.

To be consistent with the `Acquire` method name `ConnAllocTimeout` should be `ConnAcquireTimeout` (or even just `AcquireTimeout` -- I don't think there would be any ambiguity).

I haven't written a test case to verify, but I think there would be a goroutine/memory leak when `Acquire` has a deadline, but completes successfully. The `defer timer.Stop()` would execute and leave the goroutine hung trying to read from `timer.C` (according to the docs `Stop()` does not close the channel). I _think_ that `time.AfterFunc` would be a better choice.

These should be pretty straight forward tweaks.

`QueryExecTimeout` is a bit more subtle. At the moment any error that bubbles up from the `*net.Conn` causes the `pgx.Conn` to be killed. But the server still will keep going. It'd be nice to cancel the request on the server. It might be interesting to instead have a `Conn.Kill()` method that would establish another connection to the server and kill the original connection (with pg_terminate_backend; possibly also setting an immediate deadline to cause the original request to return). With this scheme the client code could have complete control of timeouts -- which could allow it to work with Query as well.

What do you think?

konstantin-dzreev:
Sounds like I have a nice little project. Thanks for the feedback!

jackc:
Cool. I thought a little more last night, and `Kill()` might not work for cancelling queries started directly from a ConnPool. Might need some combination of `Kill()` and a timeout.

It does look like a fun feature, let me know if you want any help.

konstantin-dzreev:
Hi Jack,
Sorry, It took a while to get back to the story. I updated (and squashed) the commit.
Could you give it another look?

P.S. Query() and rows.Next() both seem to respond to the timeout now.

konstantin-dzreev:
objections? ;)

jackc:
Sorry, I'm at Ancient City Ruby. Haven't had a chance to look at it yet.

Konstantin Dzreev wrote:

konstantin-dzreev:
Oh, np. I thought you just missed the update.

jackc:
Sorry I took so long to get back to you on this. This touches some pretty critical and tricky portions of the code so it took me a few hours to review and think through the proposed change. 

The ConnPool Acquire timeout implementation looks good. If you want to extract just that change to it's own commit I will merge it.

I'd like to have some way to override the acquire timeout on demand but that can be done later -- I think coming up with the right interface might be tricky.

The query timeout  implementation doesn't work though. Rather than set a timeout for the query as a whole it sets a timeout per row. See test below where timeout is 5 seconds -- the query takes 10 seconds but never times out.

```go
func TestExecWithQueryExecTimeoutSetQuery(t *testing.T) {
	t.Parallel()

	config := *defaultConnConfig

	config.QueryExecTimeout = 5 * time.Second
	conn := mustConnect(t, config)
	defer closeConn(t, conn)

	fmt.Println("Starting test\n\n\n")
	rows, err := conn.Query("SELECT repeat('asdf', 500) from generate_series(1,40) n where pg_sleep(0.25) is not null")
	if err != nil {
		t.Fatal(err)
	}
	n := 0
	for rows.Next() {
		var tmp []byte
		rows.Scan(&tmp)
		n++
		fmt.Printf("%d - %v: Got row\n", conn.Pid, time.Now())
	}

	t.Errorf("n:%d", n)

	if rows.Err() == nil {
		t.Fatal("Expected Query to fail with 'conn is dead', instead it did not")
	}
	if !strings.Contains(rows.Err().Error(), "conn is dead") {
		t.Fatalf("Expected Query to fail with timeout, instead it failed with '%v'", rows.Err())
	}
}
```

This actually brings up some more complications with query timeouts. There are a number of different things someone may want to timeout on -- time to first row, time for entire query, time per row, total time for acquire AND entire query. (For a similar example from the Go stdlib see https://golang.org/pkg/net/http and search for Timeout -- there are a lot of different timeout types).

One possibility is we provide an interface that the client code can use to cancel a query at any arbitrary time (maybe Kill or Cancel). While that would be perfectly flexible, in most cases it would be more convenient to have a query timeout attribute in the style you already implemented. Not really sure what is best here.

konstantin-dzreev:
Hi Jack,
Thanks for the review.

I completely agree that what was implemented was "row ready timeout". I was not sure what a "full query exec timeout" would mean: would it be a time when a query started until the very last row read? But there could be a time consuming client side code that could do things between rows reads... So I decided to go with the "row ready" one, even though the name of the timeout is a bit confusing.

Let us re-think what we would want from the timeouts and I'll come back to bug you again ;)

Thanks!

P.S. and I'm creating another pull request for the AcquireTimeout part

jackc:
I think merged PR #163 superceded this.

PR #126, opened 2016-03-17
[]uint16, []uint32, and []uint64 encoding and decoding

m1kc:
As the title says. Every operation is checked so it doesn't allow reading negative values into uints or writing uints that exceed maximum values of PostgreSQL integer types. Tests included.


jackc:
Thanks!

PR #128, opened 2016-03-22
Handling big numbers for ports

javo05:
Whe you use docker or just a different port
(not default), port numbers are int32 numbers.
This change just modify uint16 by uint32 in
ConnConfig.Port struct


josephglanville:
I'm fairly certain 65535 is the max port number, which makes it a uint16.
What port number are you trying to use that isn't uint16 but also works?

jackc:
AFAIK @josephglanville is correct, port numbers max out at 65535. Can you give an example of what you are trying to do that doesn't work?

c2h5oh:
`uint16` is correct in this case: both TCP headers have 16-bit port numbers https://www.ietf.org/rfc/rfc793.txt

While technically it is possible to have a higher port count - it's up to transport layer - none of the transports supported by PostgreSQL does that.

PR #133, opened 2016-03-25
Enable decoding to a *interface

robarchibald:
I'm creating a library that provides the ability for any database library to query directly to a json string or a struct.  In order for this to work though I need the ability to Scan into an array of interface{} values.  Pgx only allows scanning into "core types and the Scanner interface", so I created this simple patch to enable decoding to an interface.  Since the case statement is all of the types Postgres supports, I don't know if you'd want this in a separate function like DecodeToInterface, but I just left it inline for simplicity.  Also, I couldn't find a decode for uuid.  Does pgx not handle the uuid type?

### DIFF
--- a/values.go
+++ b/values.go
@@ -767,6 +767,65 @@ func Decode(vr *ValueReader, d interface{}) error {
 		*v = decodeInet(vr)
 	case *[]net.IPNet:
 		*v = decodeInetArray(vr)
+	case *interface{}:
+		switch vr.Type().DataType {
+		case BoolArrayOid:
+			*v = decodeBoolArray(vr)
+		case BoolOid:
+			*v = decodeBool(vr)
+		case ByteaOid:
+			*v = decodeBytea(vr)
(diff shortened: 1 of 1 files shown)


josephglanville:
pgx decodes UUID into either `string` or `[]byte`, whatever you prefer. As Go doesn't have a UUID type in the stdlib this is the most portable approach.

jackc:
By chance would `func (rows *Rows) Values() ([]interface{}, error)` already do what you want?

For uuid, as @josephglanville said, Go does not have a uuid type in stdlib. You can read it to `string` or `[]byte`. However, it does support database/sql custom types (as well as its own), so you could use https://github.com/satori/go.uuid if you wanted a library.

robarchibald:
Thanks to both @josephglanville and @jackc  for your replies.  I took a look at the Values() method and that can work for me.  Thanks for the help.  Always good to get something to work without extra code.

PR #141, opened 2016-05-10
Lazy load the inet constants

j16r:
This is a small change to the connection routine to lazy load the inet
constants to optimize the connection time slightly.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -63,8 +63,8 @@ type Conn struct {
 	logLevel           int
 	mr                 msgReader
 	fp                 *fastpath
-	pgsql_af_inet      byte
-	pgsql_af_inet6     byte
+	pgsql_af_inet      *byte
+	pgsql_af_inet6     *byte
 	busy               bool
 	poolResetCount     int
 	preallocatedRows   []Rows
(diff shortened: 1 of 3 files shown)


j16r:
This breaks a lot, might need a rethink.

jackc:
If we want to do this I think the proper place would be in Prepare. It would have to check the argument types and returned column types and see if a inet or cidr was among them, if so it could do the load. That said, I wonder if the connection time improvement is worth the additional complexity.

j16r:
Trying something new in most recent commit - thoughts?

I agree the prepare idea seems to add a little bit of complexity, so I think this design is a little simpler.

jackc:
What happens if the connection is not established through a ConnPool?

j16r:
If you choose not to use a connection pool, you get the old behavior - it loads the values each time.

jackc:
That makes sense, somehow I missed that when I looked at the code the first time. I agree we might want to tweak the interface for the shared objects at some point in the future, but it doesn't expose any public interface, so it can easily be changed when and if necessary.

I did make a couple of changes. First, sharing the PgTypes map between connections potentially allows a race condition, because that map is writable on a per connection basis (not that it is a good idea when using a connection pool -- but it is exposed). So instead of sharing the map it is copied to each connection.

I did the same for for the inet constants even though they aren't publicly exposed -- it just felt cleaner not to share memory between connections.

Thanks!

PR #142, opened 2016-05-10
Reduce the set of types loaded on connect [Fixes #140]

### LINKED ISSUE #140: Connections are slow on databases with a large number of tables
Hi there!

I'm using both pgx version 2.5.0 and master (master seems to be getting stuck and I don't have any idea why yet), and I'm having a problem using it against my particular database. When pgx first connects, it appears to do this query:

    select t.oid, t.typname from pg_type t where t.typtype='b'

Problem is that on my database, there are 249811 types. Tables seem to be included in this list of types, so each connection takes around 5s to finish this query.

There doesn't appear to be a good property on the table to distinguish tables from the normal types you'd want to do conversions from, so for now I've just set this to only get up to `3927 _int8range`. Any thoughts on how to reduce the number of types reduced here in a sensible / portable way?

Also each connection in a pool needs to do this query, is it possible perhaps to share the list of types at the pool level to avoid having to repeat this? I could write a pull request for this if it'd be acceptable.

Anyway, thanks for your hard work.

j16r:
Modify the pg_types query so that it does not load the table array types
that are created when new tables are created.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -303,7 +303,7 @@ func (c *Conn) connect(config ConnConfig, network, address string, tlsConfig *tl
 }
 
 func (c *Conn) loadPgTypes() error {
-	rows, err := c.Query("select t.oid, t.typname from pg_type t where t.typtype='b'")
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

Issue #144, opened 2016-05-13
Prepare should bind param types


willtrking:
PGX should bind the param types when preparing statements, not just executing them. There are edge cases where if this doesn't happen, postgres won't be able to figure out the proper types. Consider the following query

```UPDATE my_table AS t SET expiration=c.c_3 FROM(VALUES ($1,$2),($3,$4),($5,$6),($7,$8),($9,$10))  AS c(c_3,c_4) WHERE t.id=c.c_4 RETURNING *```

Attempting to prepare this statement returns ```ERROR: operator does not exist: bigint = text (SQLSTATE 42883)``` This is because Postgres can't figure out the types of the bound parameters, and assumes their all strings. When using the raw libpq C library included with postgres, this works as it properly binds the OIDs (in the 9.5.3 library this happens in ```PQsendPrepare```)

In the above, expiration is a ```timestamp without timezone``` and id is a ```biginteger```. The types of the values being bound are properly set in Go, but because the ```Prepare``` function isn't sending the params like it does in ```sendPreparedQuery```, it fails.

This would be incredibly useful for me, so I'm definitely willing to put the effort in to get it done if you'd like it included.

jackc:
It really surprises me that this would work with libpq. I took a quick look at the source for `PQsendPrepare` and I didn't see anything that would auto-identify types. It only uses parameter OID's if provided by the client (https://github.com/postgres/postgres/blob/master/src/interfaces/libpq)/fe-exec.c#L1251).

I'm interested in this, but I'm curious how it could work.

willtrking:
Ah you're right, and yes that is what I was doing. Well I think a good in between would be a "PrepareWithTypes" that would allow the client to provide the types if they are known.

jackc:
Do you find it easier to pass in OIDs to prepare rather than just supply the appropriate cast in the SQL (e.g. `UPDATE my_table AS t SET expiration=c.c_3 FROM(VALUES ($1::timestamp,$2::bigint),($3::timestamp,$4::bigint),($5::timestamp,$6::bigint),($7::timestamp,$8::bigint),($9::timestamp,$10::bigint)) AS c(c_3,c_4) WHERE t.id=c.c_4 RETURNING *`

Assuming so, I'm open to a advanced variant of `Prepare`. I'd probably want to have the advanced options (initially just the OID list) be a struct so if necessary we could add additional options (such as choosing between binary or text protocol) later without changing the method signature.

willtrking:
[...]

PR #146, opened 2016-05-18
Preparex

willtrking:
In reference to https://github.com/jackc/pgx/issues/144

Adds the `Preparex` function to both `Conn` and `ConnPool`. Adds the `PreparexOptions` struct to support `Preparex`

Sample usage:
```
pgxPool, _ := pgx.NewConnPool(pgx.ConnPoolConfig{
		ConnConfig: pgx.ConnConfig{
			Host:     "my_postgres_host",
			Port:     5432,
			User:     "pg_user",
			Password: "super_secure",
			Database: "tester",
		},
		MaxConnections: 1,
})

prepOpts := pgx.PreparexOptions{
	ParameterOids: []pgx.Oid{
		pgx.TimestampOid,
		pgx.Int8Oid,
		pgx.TimestampOid,
		pgx.Int8Oid,
	},
}

pgxConn, _ := pgxPool.Acquire()

pgxConn.Preparex("tester", `UPDATE my_table AS t 
SET expiration=c.c_3 
FROM(VALUES ($1,$2),($3,$4)) AS c(c_3,c_4) 
WHERE t.id=c.c_4 RETURNING *`, prepOpts)

timeOne, _ := time.Parse(time.RFC3339Nano, "2016-05-10T22:22:20.6666Z")
timeTwo, _ := time.Parse(time.RFC3339Nano, "2016-04-28T10:10:11.4444Z")

pgxConn.Exec("tester", timeOne, 253, timeTwo, 254)
```

Using the standard `Prepare` with the above query will fail with `ERROR: operator does not exist: bigint = text (SQLSTATE 42883)`, as `Query` and `Exec` will.

I would still propose  `Queryx` and `Execx` functions to make accessing the `Preparex` functionality more straightforward.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -78,6 +78,11 @@ type PreparedStatement struct {
 	ParameterOids     []Oid
 }
 
+// PreparexOptions is an option struct that can be passed to Preparex
+type PreparexOptions struct {
+	ParameterOids []Oid
+}
+
 // Notification is a message received from the PostgreSQL LISTEN/NOTIFY system
 type Notification struct {
 	Pid     int32  // backend pid that sent the notification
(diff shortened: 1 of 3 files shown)


willtrking:
Also now includes #147

renak:
Is there a way to dynamically generate the OID types to pass in based on the field types in my struct model? I assume your library must already have a function that maps go types to oid that I can utilize?

willtrking:
@renak It would be possible to do this with reflection (if you wanted to attempt it I'd recommend `reflectx` to make your life easier). However it should be noted that in many cases there isn't a 1 to 1 mapping between Go types and PG OID's. Is a Go `string` a PG `text` field? A `varchar` field? PG doesn't have unsigned integers, so what does `uint32` or `uint64` go to?

I ran into a similar issue of having to specify OID's throughout my codebase, so instead I wrote a basic global cache of OID's that pulls them directly from Postgres instead of attempting to map Go types.

Here's a gist of the copy pasted file, note it uses one of my personal projects instead of `pgx` for connections, so anywhere you see `pgsqlxx.Connxx` just replace it with `pgx.Conn`. After it's setup, it will cache OID's on a per-table basis (you specify the tables in Go) and has a timer to recache them.

Here's the cache code: https://gist.github.com/willtrking/0a351054676c352bb244aa2e60f131b6
And here's a sample if it being setup for use (I'd run this wherever you initialize `pgx`): https://gist.github.com/willtrking/3844dd93e7870d3b12c7e4562a240f73

Hope this is useful!

renak:
@willtrking Thanks for your help and for the quick response. That's definitely a creative solution that I will consider!

PR #147, opened 2016-05-18
Add Prepare to Tx

willtrking:
Even though prepared statements are bound to a connections session and are not isolated by transaction, we need this functionality as there is no way to access the transactions underlying connection.

EDIT:

Just realized there IS a way to access the underlying `Conn` in a `Tx`, my fault. Leaving this here as I think it would be a nice convenience function.


jackc:
This was merged in #146.

PR #149, opened 2016-06-01
Modification makes it run on 32-bit linux system (kali_sana)

anil1596:
In `func encodeUInt()`, `case Int8Oid:`, changed `value` to `int64(value)` and `math.MaxInt64` to `int64(math.MaxInt64)`


jackc:
Thanks!

PR #159, opened 2016-06-28
Copy sql query from prepared statement to rows.

krzysztofdrys:
This enables proper logging of sql query when using stdlib. As `sql` argument is an empty string, `rows.sql` is empty as well. Therefore in `rows.close()` empty string is logged as a sql query. This problem is  fixed by coping sql query from the prepared statement.


jackc:
Thanks!

PR #162, opened 2016-07-05
Introduced interface for the returning value of QueryRow func(s) that allows to mock DB connection

Alma-media:
Introduced Scanable interface for using as returning value instead of *Row. It allows testing code without establishing real database connection:

```

type someScanner struct {
    args []interface{}
}

func (ss *someScanner) Scan(dest ...interface{}) error {
    // do something with dest here
    return nil
}

type anotherScanner struct {
    args []interface{}
}

func (as *anotherScanner) Scan(dest ...interface{}) error {
    // do something with dest here
    return nil
}

type defaultScanner struct {
    args []interface{}
}

func (ds *defaultScanner) Scan(dest ...interface{}) error {
    // do something with dest here
    return nil
}

type mockConn struct {}

func (mc *mockConn) QueryRow(query string, args ...interface{}) pgx.Scanable {
    switch query {
    case someQuery:
        return &someScanner{args}
    case anotherQuery:
        return &anotherScanner{args}
    default:
        return &defaultScanner{args}
    }
}

```

Let's imagine some func that depends on db connection and has to be tested:

```
func myfunc(conn pgx.Conn) []byte {
    var res []byte
    conn.QueryRow(
        someQuery,
        someArg1,
        someArg2,
    ).Scan(&res)
    return res
}
```

So, now we can use mock in our tests:

```
func TestSomething(t *testing.T) {
    conn := &mockConn{}
    res := myfunc(conn)
    // check result here
}
```

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -308,7 +308,7 @@ func (p *ConnPool) Query(sql string, args ...interface{}) (*Rows, error) {
 // QueryRow acquires a connection and delegates the call to that connection. The
 // connection is released automatically after Scan is called on the returned
 // *Row.
-func (p *ConnPool) QueryRow(sql string, args ...interface{}) *Row {
+func (p *ConnPool) QueryRow(sql string, args ...interface{}) Scannable {
 	rows, _ := p.Query(sql, args...)
 	return (*Row)(rows)
 }
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -307,8 +307,8 @@ func (p *ConnPool) Query(sql string, args ...interface{}) (*Rows, error) {
 
 // QueryRow acquires a connection and delegates the call to that connection. The
 // connection is released automatically after Scan is called on the returned
-// *Row.
-func (p *ConnPool) QueryRow(sql string, args ...interface{}) Scannable {
+// *Row wich implements Scanable interface.
+func (p *ConnPool) QueryRow(sql string, args ...interface{}) Scanable {
 	rows, _ := p.Query(sql, args...)
 	return (*Row)(rows)
 }
--- a/query.go
+++ b/query.go
@@ -7,9 +7,9 @@ import (
 	"time"
 )
 
-// Scannable interface provides ability to mock DB connection for testing code
+// Scanable interface provides ability to mock DB connection for testing code
 // without establishing real DB connection.
-type Scannable interface {
+type Scanable interface {
 	Scan(...interface{}) error
 }
 
@@ -482,8 +482,8 @@ func (c *Conn) getRows(sql string, args []interface{}) *Rows {
 
 // QueryRow is a convenience wrapper over Query. Any error that occurs while
 // querying is deferred until calling Scan on the returned *Row which implements
-// Scannable interface. That *Row will error with ErrNoRows if no rows are returned.
-func (c *Conn) QueryRow(sql string, args ...interface{}) Scannable {
+// Scanable interface. That *Row will error with ErrNoRows if no rows are returned.
+func (c *Conn) QueryRow(sql string, args ...interface{}) Scanable {
 	rows, _ := c.Query(sql, args...)
 	return (*Row)(rows)
 }
--- a/query.go
+++ b/query.go
@@ -7,6 +7,12 @@ import (
 	"time"
 )
 
+// Scannable interface provides ability to mock DB connection for testing code
+// without establishing real DB connection.
+type Scannable interface {
+	Scan(...interface{}) error
+}
+
 // Row is a convenience wrapper over Rows that is returned by QueryRow.
 type Row Rows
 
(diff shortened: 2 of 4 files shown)


jackc:
Hmm. This is interesting. I think the idea of moving to an interface is reasonable, but it is a breaking change. I don't explicitly guarantee it, but I try to preserve Go-style compatibility. Any code with a *pgx.Rows type directly specified instead of via type inference would break. OTOH, it's unlikely much code is doing that... Not really sure what is best here.

Alma-media:
Hi Jackc. Thanks for your answer. You are right, it could be a breaking change for those who use *pgx.Rows type directly. Probably you'll use more interfaces in the next version of pgx. For now I have to use my "forked" version to have ability to write unit tests for my code. If you found it braking for the current version, please remove PR. Thank you.
Regards,
Alex

GildasCh:
First of all, thanks for that pull request, testing pgx without a real database connection has been my major concern with pgx. A few comments:
1. `Scanner` would be a more go-style name for the interface
2. `Rows`(with at least `Next` and `Scan`) would have to be supported too, right?
3. What about providing a simple constructor for `Row` and `Rows` types? Then you would be able to override the query functions (using a custom interface for example)

jackc:
I've been thinking of starting a v3-experimental branch for a while to experiment with some possible breaking changes. This would be the type change that could go there.

However, after mulling on this for a while, I do have another concern about returning an interface. With a normal type, adding a method is not a breaking change. But adding a method to an interface is a breaking change. If we go this route it means the only way to safely add a method is to rely on the caller to type assert to an additional interface or the underlying type.

I'd like to take a step back and consider what are other options to make pgx easier to test before making a difficult to reverse decision.

GildasCh:
@jackc What about providing a simple constructor for `Row`and `Rows` types?

This would allow testing in that fashion:

1. Create a custom pool interface to be used in code
```
type readPool interface {
	Query(sql string, args ...interface{}) (*pgx.Rows, error)
	Close()
}
```

2. In tests, create a Mock struct that satisfies that interface
```
type MockReadPool struct {}

func (m MockReadPool) Query(sql string, args ...interface{}) (*pgx.Rows, error) {
	return pgx.NewRows(...), nil
}

func (m MockReadPool) Close() {}
```

Alma-media:
Hi! It is exactly what I have now in my app
```
// ConnInterface is the connection interface.
type ConnInterface interface {
	Listen(channel string) error
	Unlisten(channel string) error
	Exec(string, ...interface{}) (pgx.CommandTag, error)
	QueryRow(string, ...interface{}) pgx.Scanable
	WaitForNotification(time.Duration) (*pgx.Notification, error)
}

// PoolInterface is the connection pool interface
type PoolInterface interface {
	Acquire() (ConnInterface, error)
	Release(ConnInterface)
	Exec(string, ...interface{}) (pgx.CommandTag, error)
	QueryRow(string, ...interface{}) pgx.Scanable
}
```
But the point is that I need an interface for the returning value of the `QueryRow` method. I'm unable to create a proper mock object of `type Rows ` (it has a lot of private fields).
Thanks for your attention to my issue

jackc:
I'm not sure how a public constructor for Row/Rows would work. Rows doesn't actually contain the result set directly. It reads it from the underlying connection as needed. So as far as directly doing something in pgx we would need to go the convert Rows to interface route or change Rows substantially to basically be a wrapper that can switch between the real rows implementation or a mock. Of the two an interface would be far simpler. But as we've discussed above there are some issues with that as well... Sigh...

Here's a few other possibilities for tackling the problem at a different level.

In my apps I typically do not call pgx directly. I define higher level methods like GetUser. Perhaps it would make for sense to mock at the data access layer instead of the database connection.

Or to go the other direction, what if we had a mock PostgreSQL server that pgx could connect to? I've thought of doing this to make certain edges of pgx easier to test.

tejasmanohar:
I'm running into the same issue. I _think_ if you use the db.Sql interface to pgx, you can get away with https://github.com/DATA-DOG/go-sqlmock.

terinjokes:
@tejasmanohar which is fine until you find you need some of the pgx-only behavior.

julianvmodesto:
Like mocking pgx's `CopyFrom` :(

denghejun:
You guys also can take look at this repo, to mock pgxpools, it works for me: https://github.com/driftprogramming/pgxpoolmock

```
func TestName(t *testing.T) {
	t.Parallel()
	ctrl := gomock.NewController(t)
	defer ctrl.Finish()

	// given
	mockPool := pgxpoolmock.NewMockPgxPool(ctrl)
	columns := []string{"id", "price"}
	pgxRows := pgxpoolmock.NewRows(columns).AddRow(100, 100000.9).ToPgxRows()
	mockPool.EXPECT().Query(gomock.Any(), gomock.Any(), gomock.Any()).Return(pgxRows, nil)
	orderDao := testdata.OrderDAO{
		Pool: mockPool,
	}

	// when
	actualOrder := orderDao.GetOrderByID(1)

	// then
	assert.NotNil(t, actualOrder)
	assert.Equal(t, 100, actualOrder.ID)
	assert.Equal(t, 100000.9, actualOrder.Price)
}
```

bfontaine:
6 years later, is this PR still relevant?

bfontaine:
No response; I’m closing this. If you think it’s still relevant, please reopen a PR with no conflict with the base branch.

PR #163, opened 2016-07-06
Non blocking ConnPool.Acquire()

konstantin-dzreev:


konstantin-dzreev:
Hi. We tried to make ConnPool.Acquire less blocking. When there was a connection issue the old code blocked other goroutines on ConnPool.Acquire method even when had our own dialer with OpenTimeout set. Instead of taking about one OpenTimeout secs to fail on bad connection, N goroutines failed after N*OpenTimeout seconds. 

The commit is not perfect because it breaks the atomicity of the connection acquisition. But we did not find any better way yet.

jackc:
I may not be understanding the issue in it's entirety, but couldn't the existing AcquireTimeout be used for this? If you set the ConnConfig.AcquireTimeout to the same value as Dialer.OpenTimeout I think that would give the same effect.

jackc:
I've merged this in and a I added a few improvements on top of it.

I changed the test that connected to microsoft.com to use a custom dialer that returns the same error as a normal timeout. 7477020

I also simplified the acquire logic a bit as well. The old logic created a placeholder Conn and put it in the pool. A problem with that is during the time the pool is unlocked for the connect call, someone could check that invalid connection out. I replaced this logic with tracking and checking the number of in-progress connect attempts. This also solves the problem of doing a connect only to find that too many connections are in the pool and having to immediately close our new connection and the problem of a concurrent ResetPool. ecf158b

konstantin-dzreev:
Thanks for merging this in!

P.S. and for fixing typos

alaska:
This is a great change. Thanks for the hard work, guys.

PR #169, opened 2016-08-21
Updates test instructions in README

manniwood:
Lets the user know about extra packages that need to be installed for the tests to run, and that connection_settings_test.go.example has been renamed to conn_config_test.go.example.


jackc:
Thanks!

PR #174, opened 2016-09-02
Switches Oid from int32 to uint32

manniwood:
According to https://www.postgresql.org/docs/9.5/static/datatype-oid.html, "The oid type is currently implemented as an unsigned four-byte integer."

In src/include/postgres_ext.h,

```
typedef unsigned int Oid;
```

And in a psql sesson:

```
# create table t (o oid);
CREATE TABLE
# insert into t (o) values (1);
INSERT 0 1
# insert into t (o) values (4294967295);
INSERT 0 1
# insert into t (o) values (4294967296);
ERROR:  22003: OID out of range
```

In the current implementation, when we try this:

```
  var minOid pgx.Oid
  var maxOid pgx.Oid
  err := conn.QueryRow("select 1::oid, 4294967295::oid").Scan(&minOid, &maxOid)
```

maxOid comes out as -1 instead of 4294967295. This patch fixes that.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -53,6 +53,10 @@ func (s *startupMessage) Bytes() (buf []byte) {
 	return buf
 }
 
+// Oid (Object Identifier Type) is, according to https://www.postgresql.org/docs/current/static/datatype-oid.html,
+// used internally by PostgreSQL as a primary key for various system tables. It is currently implemented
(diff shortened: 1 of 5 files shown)


jackc:
Thanks!

PR #175, opened 2016-09-05
Add Xid/NullXid type

manniwood:
In later versions of PostgreSQL, the `pg_stat_activity` system view has two new columns: `backend_xid` and `backend_xmin`.

This PR introduces the Xid/NullXid type so that one can select these columns without having to do any casting in SQL first.

This will also avoid casting if one wants to read the hidden/system `xmin`/`xmax` columns on any table in PostgreSQL:

```SQL
select xmin, xmax, * from some_table;
```

It is currently implemented as an unsigned for byte integer. (See `TransactionId` in src/include/postgres_ext.h in the PostgreSQL sources.)

### DIFF
--- a/values.go
+++ b/values.go
@@ -340,7 +340,7 @@ func (n NullInt32) Encode(w *WriteBuf, oid Oid) error {
 //
 // it is the data type of the xmin and xmax hidden system columns.
 //
-// It is currently implemented as an unsigned for byte integer.
+// It is currently implemented as an unsigned four byte integer.
 // Its definition can be found in src/include/postgres_ext.h as TransactionId
 // in the PostgreSQL sources.
 type Xid uint32
--- a/values.go
+++ b/values.go
@@ -329,9 +329,23 @@ func (n NullInt32) Encode(w *WriteBuf, oid Oid) error {
 	return encodeInt32(w, oid, n.Int32)
 }
 
+// Xid is PostgreSQL's Transaction ID type.
+//
+// In later versions of PostgreSQL, it is the type used for the backend_xid
+// and backend_xmin columns of the pg_stat_activity system view.
+//
+// Also, when one does
+//
+// 	select xmin, xmax, * from some_table;
+//
+// it is the data type of the xmin and xmax hidden system columns.
+//
+// It is currently implemented as an unsigned for byte integer.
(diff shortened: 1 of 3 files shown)


values.go
```diff
@@ -327,6 +329,61 @@ func (n NullInt32) Encode(w *WriteBuf, oid Oid) error {
...
+// Its definition can be found in src/include/postgres_ext.h as TransactionId
+// in the PostgreSQL sources.
+type Xid uint32
+
+// NullXid represents a Transaction ID (Xid) that may be null. NullXid implements the
+// Scanner and Encoder interfaces so it may be used both as an argument to
+// Query[Row] and a destination for Scan.
+//
+// If Valid is false then the value is NULL.
+type NullXid struct {
```
james-lawrence: is it even possible for Xid to be null? I can only think of a few spots where it may make sense, maybe temporary tables/materialized views.

manniwood: Yes, definitely. Here's an example of me logged on to my personal copy of PostgreSQL 9.5, selecting both xid-typed columns from pg_stat_activity. The backend_xid column is null.

```
# \pset null '[NULL]'
# select backend_xid, backend_xmin from pg_stat_activity;
┌─────────────┬──────────────┐
│ backend_xid │ backend_xmin │
├─────────────┼──────────────┤
│      [NULL] │      1090671 │
└─────────────┴──────────────┘
(1 row)
```

One thing that prompted this PR was a desire to easily read every column from pg_stat_activity. This would be useful to a lot of PostgreSQL tool writers, as well as non-novice users looking to monitor the state connections to their database.

james-lawrence: ha nice, yeah only places I could think of is where it is in a result but uninitialized, and there it is pg_stat_activity. =)

manniwood: Another easy way it could happen in practice is if an xid-typed column is
involved in a left join where no row can be joined to on the right-hand
side of the join: that column will obviously satisfy the criteria: it will
be type xid, and it will be null.

manniwood: But, yes, you inspired me to try to solve one of your root problems! :-)

jackc:
Thanks!

manniwood:
My pleasure! Awesome project.

PR #179, opened 2016-09-13
Add Cid/NullCid Type

manniwood:
When one does

```SQL
select cmin, cmax, * from some_table;
```

The cmin and cmax system columns are of type cid.

cid is currently implemented as an unsigned four byte integer. Its definition can be found in src/include/c.h as CommandId in the PostgreSQL sources.


jackc:
Thanks!

PR #181, opened 2016-09-16
Add Tid/NullTid Type

manniwood:
When one does

```SQL
select ctid, * from some_table;
┌───────┬───┐
│ ctid  │ x │
├───────┼───┤
│ (0,1) │ 1 │
└───────┴───┘
(1 row)
```

The ctid system column is of type tid (Tuple Identifier).

tid is currently implemented as a pair of numbers: an unsigned four byte integer and an unsigned two byte integer. Its conversion functions can be found in src/backend/utils/adt/tid.c in the PostgreSQL sources.

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -136,6 +136,12 @@ func (wb *WriteBuf) WriteInt16(n int16) {
 	wb.buf = append(wb.buf, b...)
 }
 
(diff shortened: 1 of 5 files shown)


jackc:
Thanks!

PR #185, opened 2016-09-20
Add Char/NullChar Type

manniwood:
Adds the PostgreSQL "char" type, which is used in a lot of the system catalog tables.

Fun note: the "char" type always has double-quotes around it when used in a psql session:

```
-- we are not casting to the "char" type
# select 'a'::char; rollback;
┌────────┐
│ bpchar │
├────────┤
│ a      │
└────────┘

-- we are casting to the "char" type
# select 'a'::"char"; rollback;
┌──────┐
│ char │
├──────┤
│ a    │
└──────┘
```

The "char" type can hold one byte, and is pretty much just the C language's char type translated into PostgreSQL. It's used in a lot of system table columns to hold a single character. (The pg_class.relkind column is a prime example.) It's more akin to Go's byte type than it is to the rune type, given its limited range of 0 through 255.

### DIFF
--- a/values.go
+++ b/values.go
@@ -69,7 +69,8 @@ const minInt = -maxInt - 1
 // or binary). In theory the Scanner interface should be the one to determine
 // the format of the returned values. However, the query has already been
 // executed by the time Scan is called so it has no chance to set the format.
-// So for types that should be returned in binary th
+// So for types that should always be returned in binary the format should be
+// set here.
 var DefaultTypeFormats map[string]int16
 
 func init() {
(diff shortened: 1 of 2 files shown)


manniwood:
I see the Travis CI build failed. I was developing/testing using go 1.7, so I installed go 1.6.2 to test against my Pg 9.5.2 install (to match that pair of versions in one of the Travis CI failures) and the TestConnQueryDatabaseSQLDriverValuer test still passed for me. Unsure what to try next.

jackc:
I'm getting that error on master now, so I don't think it is anything with this PR. I can reproduce error locally by updating `github.com/shopspring/decimal`. Still looking into what's going on.

manniwood:
Oh, interesting! I would not have thought of that. Note to self. :-)

jackc:
https://github.com/shopspring/decimal/commit/54efccb61ffe0a31b6a1908bdc8a35f491da01ea was the problem. Need to use pointer to value instead of value. Fixed on master now.

manniwood:
Cool. Will do a git pull next chance I get.

On Thu, Sep 22, 2016 at 10:12 AM, Jack Christensen <notifications@github.com

jackc:
Thanks!

PR #191, opened 2016-09-28
Added schema to CopyTo().

davidsonff:


tejasmanohar:
@jackc Are you +1 on merging this? We have a fork just for this :\

jackc:
@tejasmanohar I didn't want to merge a breaking change. But as it turns out `CopyTo` is actually the wrong name, it really should be `CopyFrom`. See #243. So we can introduce a a new method and deprecate the old one without making a breaking change. I also prefer the generalization of an identifier type rather separate schema and column string parameters. I put together my approach in #246. Barring any negative feedback I'll merge that in a few days.

jackc:
#246 is now merged -- schemas can be used with the new `CopyFrom` method.

PR #192, opened 2016-09-29
Adds Name/NullName types

manniwood:
The name type is used a lot in the system tables, such as the `relname` column of `pg_class`.

Under the hood, it's pretty much just a C-string that can be 63 bytes long, not counting the '\0' terminator.

This PR also accompanies the pgx.Name type with a little convenience method that says whether a name is short enough. PostgreSQL will silently truncate longer names, so the convenience method can be used by those who want to know beforehand.

### DIFF
--- a/values.go
+++ b/values.go
@@ -287,7 +287,7 @@ func (n Name) LengthOK() bool {
 // If Valid is false then the value is NULL.
 type NullName struct {
 	Name  Name
-	Valid bool // Valid is true if Char is not NULL
+	Valid bool // Valid is true if Name is not NULL
 }
 
 func (n *NullName) Scan(vr *ValueReader) error {
--- a/values.go
+++ b/values.go
@@ -19,6 +19,7 @@ const (
 	BoolOid             = 16
 	ByteaOid            = 17
 	CharOid             = 18
+	NameOid             = 19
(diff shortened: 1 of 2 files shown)


jackc:
IIRC length of name is configurable at compile time. I can't imagine many people actually do that, but that might render the length check error prone.

manniwood:
Yeah, definitely. If someone custom compiles postgresql to support longer
identifiers, do we also assume they don't mind forking / custom compiling
pgx to match? Should my documentation around the helper method and constant
make that more clear? Is there a more elegant solution? Is the cleanest
solution just to remove the helper method and constant altogether?

jackc:
I think the helper method should either be correct or be removed.

It'd be possible to query the server at connection time to find out the maximum length (something like what is done for inet/cidr in Conn.loadInetConstants). But it feels like such a rare use that I think that documentation explaining the the silent truncation would be sufficient.

manniwood:
OK, cool. I have done another push to my branch, removing the constant and the helper method, and expanding the docs around the Name type. Let me know if this seems better.

manniwood:
Oh, but let me merge master to my branch to resolve the conflict...

jackc:
Thanks!

PR #193, opened 2016-10-03
Fix test failure when DB and client are not in the same time zone

waisbrot:
To reproduce the test failure, create a database where the Postgres server's local time is not your machine's local time. I made mine inside a Docker container where it was UTC while my machine was in EDT.

This patch explicitly sets the time zone to UTC in the database and in the test expectation, then compares the two times in the client-local time zone.


jackc:
Thanks!

PR #197, opened 2016-10-11
Fixes TestNullX test of NullName

manniwood:
Fixes my copy-paste error when adding the test for NullName in TestNullX.


jackc:
Thanks!

PR #203, opened 2016-10-27
Simplify QueryExecTimeout

danielrife:
When postgres accepts connections but does not respond to a request, QueryExecTimeout causes client to hang.
Simplify QueryExecTimeout to not make an extra request to postgres but just close current connection


danielrife:
Looking at travis failure

PR #204, opened 2016-10-28
Add AclItem/NullAclItem

manniwood:
The aclitem datatype is another PostgreSQL-specific data type used in the system catalogs. This one is kind of tricky for a few reasons: 1) it only goes over the wire in text format, not binary; 2) an aclitem is valid if the user/role after the slash in the aclitem actually exists in PostgreSQL; 3) the user/role part of an aclitem is a SQL identifier, with all the crazy formatting rules that go with that: within double quotes, pretty much anything goes.

The pg_class system table's relacl column actually uses the aclitem[] type (not aclitem), which also goes over the wire in text format only, so if this PR gets accepted, the next thing I'd like to figure out is how to get aclitem[] working too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,8 @@ env:
     - PGVERSION=9.2
     - PGVERSION=9.1
 
+# The tricky test user, below, has to actually exist so that it can be used in a test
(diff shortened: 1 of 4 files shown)


jackc:
Thanks!

PR #210, opened 2016-11-18
Add aclitem array

manniwood:
Adds the aclitem[] / []pgx.AclItem type. This will allow users to read the relacl column of the pg_class system table.


jackc:
Thanks!

PR #220, opened 2016-12-16
Adds regproc OID synonym type

manniwood:
OID synonym types appear in the system tables (https://www.postgresql.org/docs/current/static/datatype-oid.html#DATATYPE-OID-TABLE) to conveniently represent OIDs as human-readable names.

The PR adds the regproc type (which gets used for a lot of the columns in pg_type, for instance).

If this PR gets approved, I'll add the other OID synonym types too, one PR at a time.

### DIFF
--- a/values.go
+++ b/values.go
@@ -24,6 +24,7 @@ const (
 	Int8Oid             = 20
 	Int2Oid             = 21
 	Int4Oid             = 23
+	RegProcOid          = 24
 	TextOid             = 25
 	OidOid              = 26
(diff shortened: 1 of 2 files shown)


bfontaine:
@manniwood I know this PR is very old, but if you think the changes are still relevant, could you please rebase it on top of the latest code so it can be reviewed and merged? Thanks!

manniwood:
@bfontaine , wow, this PR is indeed quite old. I don't have time to review it right now, so I will just close it.

PR #222, opened 2016-12-31
Try add support for []record type

RHavar:
This doesn't actually work (not sure why), but I added a (failing) test case and some code which I think would be a useful start.

So this is probably best treated as an issue, but it would be nice if pgx could support []record types


bfontaine:
If this doesn’t work and nobody is actively working on this, I believe we can close this 6-years-old PR.

PR #224, opened 2017-01-05
Add pg 9.4+ logical decoding / logical replication support

kriswehner:
This adds logical decoding support to pgx, including explicit start replication support and functionality for sending StandbyStatus messages to the master to track replication positions.

WalMessage payload is exposed as an unstructured []byte, as the content of the payload itself depends on the logical decoding plugin in use. Testing directions are updated, and rely only on the example test_decoding support that's shipped bundled with postgres.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,11 @@ before_install:
   - echo "host      all         pgx_md5     127.0.0.1/32          md5"      >> /etc/postgresql/$PGVERSION/main/pg_hba.conf
   - echo "host      all         pgx_pw      127.0.0.1/32          password" >> /etc/postgresql/$PGVERSION/main/pg_hba.conf
   - echo "hostssl   all         pgx_ssl     127.0.0.1/32          md5"      >> /etc/postgresql/$PGVERSION/main/pg_hba.conf
-  - echo "host      replication pgx_ssl     127.0.0.1/32          md5"      >> /etc/postgresql/$PGVERSION/main/pg_hba.conf
+  - echo "host      replication pgx_replication 127.0.0.1/32      md5"      >> /etc/postgresql/$PGVERSION/main/pg_hba.conf
   - sudo chmod 777 /etc/postgresql/$PGVERSION/main/postgresql.conf
-  - echo "wal_level='logical'"     >> /etc/postgresql/$PGVERSION/main/postgresql.conf
-  - echo "max_wal_senders=5"       >> /etc/postgresql/$PGVERSION/main/postgresql.conf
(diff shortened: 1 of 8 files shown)


kriswehner:
My apologies for the test failures; the replication code relies on additional users and replication permission. I'd be happy to do whatever legwork is necessary to get these passing, but I'm not clear on the travis config.

jackc:
This is really cool!

Two things should be able to fix the tests.

First, we should add a t.Skip if the replication user isn't set up. See https://github.com/jackc/pgx/blob/master/conn_test.go#L62 for an example of how the Unix domain socket connection is skipped if not set up (necessary for Windows). That will avoid test failures for people who don't set up all the pg_auth stuff.

That actually should make the test suite pass on Travis, but only by skipping the test. So to get the test to work on Travis we would need to add the pg_hba.conf and postgresql.conf stuff to the .travis.yml file in the before_install hook. I think it should just be adding a few more `echo ... >>` lines to append the relevant data. Then we need to add a check to the test for a high enough version of PostgreSQL and skip the test if it is too old for logical replication.

I'm pretty tied up today, but I might have a chance to look into it further tomorrow.

kriswehner:
@jackc Apologies for the stream of consciousness commits there as I tried to get travis working. I think I got both travis and the skips done correctly now. 

Just to be clear: The functionality (should) work on 9.4+, but my tests rely in a 9.6 specific column to verify that the StandbyStatus updates are being received, so I've got the tests set to run only on 9.6.

josephglanville:
Does `flush_location` in `pg_stat_replication` work in this context?
I am not too familiar with logical replication slots yet but this seems awesome!

kriswehner:
My testing says it does not; ```flush_location``` appears to be expose internal LSNs that don't reliably update when the standby status is sent. I spent quite a while trying to stabilize the tests using the ```flush_location``` and couldn't get reproducible results until I found the new ```confirmed_flush_lsn```

josephglanville:
I think to make `flush_location` work reliably `synchronous_standby_names` would have to be used or tests would need to be aware of the async nature of the table. However that setting is server wide so I am not sure about implications regarding logical decoding yet.

jackc:
Thanks for pushing through and getting the Travis support working. I know that can be a real pain sometimes.

I merged in the changes, but I do have an API change to suggest. As I understand it, a replication connection cannot do any non-replication things, and a regular connection cannot do any replication things. Perhaps it would make sense to split `Conn` into `Conn` and `ReplicationConn`. This would avoid either type of connection having methods that could never be called successfully. Presumably `ReplicationConn` would simply wrap `Conn` or even just be a renamed type. This would also let a `ConnectReplication` function handle assigning the `replication=database` connection parameter.

PR #226, opened 2017-01-09
Break out the replication connection into its own connection type

kriswehner:
This puts all the replication functionality in its own connection class, to avoid confusion by the end user using functionality that is unavailable on replication type connections.

### DIFF
--- a/helper_test.go
+++ b/helper_test.go
@@ -13,7 +13,23 @@ func mustConnect(t testing.TB, config pgx.ConnConfig) *pgx.Conn {
 	return conn
 }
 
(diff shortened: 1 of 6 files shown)


kriswehner:
Please hold off on merging this, I want to fully verify that it's ready to go with my client project before we call this done.

kriswehner:
@jackc This should be good to go. We've got it merged in our mainline.

The big changes are:
- I dropped the StopReplication() function, as there was no way to use the connection after a stop, so it was a poor API. You instead just need to Close() the connection instead.
- I added calls to fetch the system metadata around timelines and timeline history, which is necessary to write a full-fledged logical replication client.

jackc:
Looks great! Thanks!

PR #233, opened 2017-02-01
support for uuid[]

alicebob:
#Add support for UUID arrays. It works, but the change is rather pervasive.

Unfortunately # 1: This switches UUIDs over to the binary protocol, with UUIDs encoded over the wire as 16 bytes, not as text, so we need to add the suitable en/decoding.

Unfortunately # 2: , since a uuid in Go is just a string we can't switch on the type of what we decode to, since that's a `[]string` or a `*string`, so we need to check the OID of the column everywhere we deal decode into a string.

I guess we could add a case for a `[16]byte`, so we can skip the en/decoding step, if the user prefers that format.

Maybe you have an idea to solve this in a nicer way, @jackc?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -920,7 +920,7 @@ func (c *Conn) sendPreparedQuery(ps *PreparedStatement, arguments ...interface{}
 			wbuf.WriteInt16(TextFormatCode)
 		default:
 			switch oid {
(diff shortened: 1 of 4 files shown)


jackc:
Between this and #234 and I think I prefer this one. Even though this change has some unfortunate complications internally it avoids adding a new type for what Go considers `[]string` or `[16]byte`.

That said, in either case `ReadUuid` should definitely not be public on `ValueReader` and probably shouldn't be on `ValueReader` at all.

Another alternative, I wonder if it is worth adding parsing for the text protocol arrays. If that had been available uuid[] would be much easier to support.

I'm not sure what the best direction here is. Let me think about this for while.

alicebob:
Thanks for the reply, no hurry.
 
Text arrays parsing sounds good. Are there more datatypes which would be easier to implement having that?

jackc:
> Text arrays parsing sounds good. Are there more datatypes which would be easier to implement having that?

Long term, I'm considering trying to support text and binary protocols for all types. There are enough odd cases that crop up due to the occasional text/binary mismatch that it would be nice if the core pgx types could all handle both.

bfontaine:
Almost 6 years later, does it make sense to keep this PR open?

alicebob:
not alt all

PR #234, opened 2017-02-01
Alternative `uuid[]`

alicebob:
This is an alternative to #233. You do need to use `pgx.UUIDs(yourStringSlice)` when querying or scanning, but the change is smaller overall. Single UUIDs are still simple strings.


alicebob:
And finally there is https://github.com/alicebob/pgx/tree/uuidsagain which adds `UUID`, `[]UUID`, and `UUIDs`, and also still works with simple strings, but that got a bit too silly.

bfontaine:
I’m closing this in favor of #233 since Jack [wrote](https://github.com/jackc/pgx/pull/233#issuecomment-279247867) that he prefers the other PR.

PR #237, opened 2017-02-11
allow a firstelem at 0 - int2vector

alicebob:
I'm reading [pg_catalog.pg_index](https://www.postgresql.org/docs/9.6/static/catalog-pg-index.html) for a [project](https://github.com/alicebob/schemaspy). It has a `int2vector` column which pgx doesn't seem to support. 
Query:
```
SELECT indexrelid, indrelid, indisunique, indisprimary, indkey::int4[]
FROM pg_catalog.pg_index
```
pgx error:
```
	int_test.go:20: can't scan into dest[4]: Expected array's first element to start a index 0 or 1, but it is 0
```

This patch seems to fix it nicely, but I see no way to add a proper test case, since I don't know how to generate a new `int2vector` type :(

### DIFF
--- a/values.go
+++ b/values.go
@@ -2406,8 +2406,8 @@ func decode1dArrayHeader(vr *ValueReader) (length int32, err error) {
 	length = vr.ReadInt32()
 
 	idxFirstElem := vr.ReadInt32()
-	if idxFirstElem != 1 {
-		return 0, ProtocolError(fmt.Sprintf("Expected array's first element to start a index 1, but it is %d", idxFirstElem))
+	if idxFirstElem < 0 || idxFirstElem > 1 {
+		return 0, ProtocolError(fmt.Sprintf("Expected array's first element to start at index 0 or 1, but it is %d", idxFirstElem))
 	}
 
 	return length, nil


jackc:
`int2vector` is a restricted type of `int2[]`. Most arrays in PostgreSQL are 1-based and allow nulls. `int2vector` is special in that it is 0-based and does not allow nulls.

I'm a little wary of making a change to the core array parsing. If we want native support for `int2vector` it probably would be best to handle it on it's own path rather than through the normal array handling.

That said, the simplest thing would be to cast the `int2vector` to `int2[]` -- with the correct slice offset.

```
SELECT indexrelid, indrelid, indisunique, indisprimary, indkey, indkey[0:]::int2[]
FROM pg_catalog.pg_index
;
```

That should give you a normal `int2[]` that pgx should be able to decode.

Also, if you need to create int2vector values directly you can do it like the following:

```
jack=# select '1 3 8'::int2vector;
 int2vector 
------------
 1 3 8
(1 row)
```

alicebob:
Thanks for the reply Jack.

Leaving out the upper bound turns out to be introduced in 9.6, but this works fine:
```indkey[0:array_length(indkey, 1)]::int4[]```

Great, everything works!

Btw, the SQL cast needs to be to `int4[]`, and the Go type must be `[]int32`.
The various decoding errors I get with other combinations:
`int2[]` -> `[]int32`:
int_test.go:20: can't scan into dest[4]: Cannot decode oid 1005 into []int32
`int2[]` -> `[]int`:
int_test.go:20: can't scan into dest[4]: Scan cannot decode into *[]int
`int2[]` -> `[]int16`:
works
`int[]` -> `[]int`:
int_test.go:20: can't scan into dest[4]: Scan cannot decode into *[]int
`int4[]` -> `[]int`:
int_test.go:20: can't scan into dest[4]: Scan cannot decode into *[]int

Issue #255, opened 2017-03-31
OpenFromConnPool contains results in a memory leak.


james-lawrence:
everytime OpenFromConnPool is invoked it leaks memory.

```
openFromConnPoolCountMu.Lock()
name := fmt.Sprintf("pgx-%d", openFromConnPoolCount)
openFromConnPoolCount++
openFromConnPoolCountMu.Unlock()

sql.Register(name, d)
```

the invocation to sql.Register leaks the name + the driver + the pool.

jackc:
I'd expect sql.Register to hold a reference to name, driver, and pool by necessity. But I also did not anticipate `OpenFromConnPool` typically being called more than once per program execution. What's the usage pattern for multiple calls to `OpenFromConnPool`?

At the time `OpenFromConnPool` was written, I didn't see any other way to connect a pre-existing pgx pool to a *sql.DB aside from registering a new driver. Looking at it with fresh eyes, it might be possible to avoid the multiple sql.Register call by doing something with db.Driver, but I'm not totally sure.

james-lawrence:
There are a couple problems with OpenFromConnPool.

1) it was the only way to go from a ConnConfig to a sql.DB using ParseDSN. I do not think there is anything that can be done about this short of my other PR.

2) It doesn't document the fact it leaks memory.

3) not sure it really has a use case. sql.DB already does connection pooling. Why would we need to add a pool for something that already does connection pooling?

james-lawrence:
Regarding the usage pattern: It was the only way to open a sql.DB connection using pgx and a DSN connection string. =) I naively assumed OpenFromConnPool was doing some internal conversion to sql.DB.

And the resulting sql.DB violated some basic promises of the sql.DB structure (such as calling db.Close() would shut down all the connections)

jackc:
`v3` removes `OpenFromConnPool` and replaces it with `RegisterDriverConfig`. This entirely eliminates the structural problem with `OpenFromConnPool`. If golang/go#20268 gets resolved it should be simple to support that as well.

PR #260, opened 2017-04-11
retain the registered driver name

hsyed:
Small change to enable getting at the driver name so that the control from OpenFromConnPool can be leveraged when integrating with other libraries,

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -71,7 +71,7 @@ var (
 var databaseSqlOids map[pgx.Oid]bool
 
 func init() {
-	d := &Driver{}
+	d := &Driver{name: "pgx"}
 	sql.Register("pgx", d)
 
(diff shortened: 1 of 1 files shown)


jackc:
Does this have anything to do with #255? I haven't come across a good solution issue to that yet -- but if there is one, it would probably involve altering `OpenFromConnPool` such that it does not register a new driver and name each time it is called. Just want to make sure this change wouldn't make fixing that issue harder in the future.

hsyed:
No, I don't think so at least. 

I was trying out pgx with Gorm -- gorm does not take a *sql.DB handle and In order to be able to use and control multiple pgx pools the driver name needs to be used as the discriminator. stdlib only manages one strongly named driver pool "pgx" transparently currently -- and the user has no control over this one.

bfontaine:
Almost 6 years later, does it make sense to keep this PR open?

PR #261, opened 2017-04-11
add decoders for *uuid.UUID and *decimal.Decimal so row.Scan() works

matthewmueller:
Hey there, this PR adds support for two additional decoder types:

- *uuid.UUID
- *decimal.Decimal

Putting these values into the database works fine, just when you want to get the values out via Scan, you get errors. This PR fixes that.

### DIFF
--- a/values.go
+++ b/values.go
@@ -13,6 +13,9 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	uuid "github.com/satori/go.uuid"
+	"github.com/shopspring/decimal"
 )
 
 // PostgreSQL oids for common types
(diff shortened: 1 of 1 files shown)


matthewmueller:
@jackc I'm in the process of upgrading to v3. Everything has been fantastic so far, but I'm still running into the issues that this PR resolves.

I'm happy to define a custom type in my own repo, but I'm still not sure how to do that with pointer types.

Any chance you could unstick me a little? Thanks!

matthewmueller:
woops! i was doing something wrong in my code, works like a charm!

PR #285, opened 2017-06-30
Add missing `pgx.Identifier` to `CopyFrom` example

meowgorithm:
Noticed a the column name needed be wrapped in type `pgx.Identifier` which was missing in the example documentation.

Excellent work with the `CopyFrom` feature, by the way.


jackc:
Thanks!

PR #291, opened 2017-07-22
ConnPool.Query(): fix Release() right after query.

soulne4ny:
subj

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -367,8 +367,8 @@ func (p *ConnPool) Query(sql string, args ...interface{}) (*Rows, error) {
 	}
 
 	rows, err := c.Query(sql, args...)
+	p.Release(c)
 	if err != nil {
-		p.Release(c)
 		return rows, err
 	}
 


jackc:
In the case of `Query` the connection is still in use until all rows are read. The connection is released in the `Rows` struct. That is why we only `Release` on error in the connection pool.

soulne4ny:
@jackc, you are right. Then `ConnPool` `QueryRow()` and `QueryRowEx()`, better release the connection right before return.
https://github.com/jackc/pgx/blob/8f4178b3d3fc8922384e0989a683b112fb5d1066/conn_pool.go#L403

jackc:
It gets closed by Rows.Next() in the ordinary case -- https://github.com/jackc/pgx/blob/master/query.go#L152

Error paths call close as well.

soulne4ny:
Using QueryRow() there is no `Rows.Next()` as only one row is returned, or, at least, it unlikely to be called.

jackc:
`QueryRow` returns a `Row` which is a type rename of `Rows`. When `Row.Scan` is called it internally calls `Rows.Next` and `Rows.Close` which then would release the connection. See [Row.Scan](https://github.com/jackc/pgx/blob/master/query.go#L22).

soulne4ny:
`QueryRow()` returns `*Row`.
```
row.Next undefined (type *pgx.Row has no field or method Next)
```

soulne4ny:
@jackc, thanks for clarification. However should not Row.Scan() call `rows.Close()` in [error case.](https://github.com/jackc/pgx/blob/8f4178b3d3fc8922384e0989a683b112fb5d1066/query.go#L26)

jackc:
The line you reference is checking if the row already has hit an error. That can only happen if the error occurs while sending the query. In that case the query is already closed and released.

PR #304, opened 2017-08-18
Fix numeric EncodeBinary bug

crvv:
There is a bug in `(*pgtype.Numeric).DecodeBinary`.
To reproduce, run the following code and the result is `1.0234567890000004`
```
package main

import (
	"fmt"
	"github.com/jackc/pgx"
)

func main() {
	conn, err := pgx.Connect(pgx.ConnConfig{})
	if err != nil {
		panic(err)
	}
	rows, err := conn.Query("SELECT $1::NUMERIC", 1.0000023456789)
	if err != nil {
		panic(err)
	}
	defer rows.Close()
	for rows.Next() {
		var x float64
		err = rows.Scan(&x)
		if err != nil {
			panic(err)
		}
		fmt.Println(x)
	}
}
```

This pull request fixes this.

### DIFF
--- a/pgtype/numeric.go
+++ b/pgtype/numeric.go
@@ -16,6 +16,7 @@ import (
 const nbase = 10000
 
 var big0 *big.Int = big.NewInt(0)
+var big1 *big.Int = big.NewInt(1)
 var big10 *big.Int = big.NewInt(10)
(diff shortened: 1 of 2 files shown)


jackc:
Ouch. That was an unfortunate bug. Thanks for the find and fix.

PR #305, opened 2017-08-18
fix logrus updated package name (lowercase)

dvic:
See https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276


jackc:
Thanks!

PR #306, opened 2017-08-19
Add more ColumnType support

cyberdelia:
This allow this driver to return valid information for [``ColumnType``](https://golang.org/pkg/database/sql/#ColumnType), at the exception of ``Nullable``.


jackc:
Thanks!

PR #309, opened 2017-08-28
Add UUIDArray type

kelseyfrancis:
Also change UUID.Set() to convert nil to NULL in order for
UUIDArray.Set() to support converting [][]byte slices that contain nil.


jackc:
Merged with a couple tweaks to hook it into the array code generation system. Thanks!

PR #311, opened 2017-08-29
Fix some invalid one round trip execs failing to return non-nil error

kelseyfrancis:
Prior to this commit, execEx() would write the one round trip exec to
the connection before first calling ensureConnectionReadyForQuery, which
ultimately caused any errors to be suppressed if the exec followed a
valid query, because the receive message processing would finish
successfully as soon as it received the ReadyForQuery that actually
belonged to the preceding query. So, the exec would never actually
receive the error message that it caused, leaving it to be incorrectly
received by the first subsequent query sent.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1458,6 +1458,10 @@ func (c *Conn) execEx(ctx context.Context, sql string, options *QueryExOptions, 
 			return "", err
 		}
 	} else if options != nil && len(options.ParameterOIDs) > 0 {
+		if err := c.ensureConnectionReadyForQuery(); err != nil {
+			return "", err
+		}
+
 		buf, err := c.buildOneRoundTripExec(c.wbuf, sql, options, arguments)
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -1458,6 +1458,10 @@ func (c *Conn) execEx(ctx context.Context, sql string, options *QueryExOptions,
 			return "", err
 		}
 	} else if options != nil && len(options.ParameterOIDs) > 0 {
+		if err := c.ensureConnectionReadyForQuery(); err != nil {
+			return "", err
+		}
```
kelseyfrancis: This block could go near the top of either `execEx` or `ExecEx` instead (similar to `QueryEx` or `prepareEx`, but I believe this is the only case where `ensureConnectionReadyForQuery` isn't always called in some other place before writing to the connection.

jackc:
Thanks!

PR #312, opened 2017-08-29
Return ErrClosedPool when Acquire() with closed pool

MaerF0x0:
*Note:* I didnt couldnt get local tests to run . Got a err `query.go:11:2: use of internal package not allowed` 

What: 
  * Use a named Err for when we attempt Acquire with a closed pool

Why: 
 * I can test err = pgx.ErrClosedPool in my consuming code 😀


jackc:
Thanks!

PR #315, opened 2017-08-31
Always return non-nil *Rows from Query to fix QueryRow

kelseyfrancis:
Since QueryRow delegates to Query, it needs Query to always return
non-nil *Rows to prevent a nil pointer deference when the QueryRow
caller calls Scan(). This commit fixes the few returns in QueryEx that
return nil on errors rather than *Rows with its err field set.


jackc:
Thanks!

PR #319, opened 2017-09-08
Skip certificate verification for sslmode "require"

### LINKED ISSUE #282: Why sslmode = require checks SSL certs.
Hello, I am wondering about this.

Why pgx treats `require` like `verify-full` ?

Would you accept a pull request with the changes ?
 
Thanks.

johanbrandhorst:
Fixes #282.

This more appropriately aligns the behaviour of the library with
that advertised by the postgres documentation.

According to the table on the official documentation page
https://www.postgresql.org/docs/current/static/libpq-ssl.html,
the "require" mode should be used when:

"I want my data to be encrypted, and I accept the overhead. I trust that the network will make sure I always connect to the server I want."

This maps reasonably well to a TLS config that skips certificate verification.


jackc:
Thanks. I updated the tests and docs and merged.

PR #328, opened 2017-09-20
Changed CreateReplicationSlot to return the consistent_point and snapshot_name

wingedpig:
When using logical replication, the snapshot name is needed to be able to do a full dump of the existing database when first setting up a replicated system.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -435,7 +435,13 @@ func (rc *ReplicationConn) StartReplication(slotName string, startLsn uint64, ti
 }
 
 // Create the replication slot, using the given name and output plugin.
-func (rc *ReplicationConn) CreateReplicationSlot(slotName, outputPlugin string) (consistentPoint string, snapshotName string, err error) {
+func (rc *ReplicationConn) CreateReplicationSlot(slotName, outputPlugin string) (err error) {
+	_, err = rc.c.Exec(fmt.Sprintf("CREATE_REPLICATION_SLOT %s LOGICAL %s", slotName, outputPlugin))
+	return
+}
+
+// Create the replication slot, using the given name and output plugin, and return the consistent_point and snapshot_name values.
+func (rc *ReplicationConn) CreateReplicationSlotEx(slotName, outputPlugin string) (consistentPoint string, snapshotName string, err error) {
 	var dummy string
 	var rows *Rows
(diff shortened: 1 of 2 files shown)


jackc:
Is there a way this could be structured to not be a breaking change? I don't want to intentionally introduce a breaking change on a non-major release. I suppose it might need to add a new function and deprecate the old.

What are the two ignored columns? If they might be useful to anyone at some point it might be best to include them. 

Maybe a `ReplicationSlot` struct of those 4 fields combined into a `CreateReplicationSlotEx()` function that returns a `(ReplicationSlot, error)`.

I don't have much experience with logical replication so I could be off-base on this.

Thanks.

wingedpig:
Completely understand about wanting a non-breaking change. The two ignored fields are: slot_name and output_plugin, both of which we already know, since they're passed in as parameters to the function in the first place.

It feels like creating a new struct just to hold two result values might be overkill. But I'm happy to make whatever changes you think are best. Just let me know what you'd like the new func name to be and whether to create a new struct, and I'll submit a new PR.

jackc:
I guess a struct isn't necessary in that case. Let's call the new function `CreateReplicationSlotEx()`.

Thanks.

wingedpig:
Ok, changes made. I've never updated a PR before, but it looks like the changes made it. Let me know if you need anything else. Thanks for your work!

jackc:
Merged. Thanks!

sidmutha:
Hi
How do I use the snapshot returned using CreateReplicationSlotEx in a transaction? I would like to run select queries on this snapshot.


The [`SET TRANSACTION SNAPSHOT`](https://www.postgresql.org/docs/9.3/static/sql-set-transaction.html) command is there but I cannot find a way to use it via pgx? Should I use the tx.Query method in pgx?

jackc:
You would need to have a separate normal connection to PostgreSQL. In that connection start a transaction and first thing inside that transaction use Exec to run the `set transaction snapshot`.

That said, `set transaction snapshot` was originally designed for consistent parallel backups (outside of replication). I have no idea how or if it is expected to work with replication (it may work fine of course, I just don't know).

sidmutha:
Hey I tried this and it worked! Thanks.
You also need the transaction isolation level to be `repeatable read`.

PR #329, opened 2017-09-21
Fix QueryRow() left over messages

plopik:
Bug previously hidden be the use of ensureConnectionReadyForQuery.

QueryRow scan now properly read all messages including the last CommandComplete before closing.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -476,3 +476,52 @@ func TestConnBeginBatchQuerySyntaxError(t *testing.T) {
 		t.Error("conn should be dead, but was alive")
 	}
 }
+
+func TestConnBeginBatchSelectInsert(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jackc:
Thanks for finding that. It was a very subtle bug. When I looked into in I found it affected more than `QueryRow`. It also affects `Query` where the `Rows` are closed early. I kept the test but changed the solution.

The underlying issue is that `ensureConnectionReadyForQuery` isn't hiding the bug in the non-batch case -- it is a necessary feature to keep connections usable after a context cancellation interrupts a connection at an arbitrary point. That is why Row.Scan cannot eagerly consume all rows.

To handle the broader bug it necessary to special case `CommandComplete` message handling for batches. I documented the details in 53e5d8e34.

PR #330, opened 2017-09-23
Adds `go get github.com/pkg/errors` to README.md

manniwood:
Following the test setup instructions and then running `go test` results in

```
internal/sanitize/sanitize.go:11:2: cannot find package "github.com/pkg/errors"
```

This patch adds an instruction to the README.md for the user to `go get github.com/pkg/errors`.


jackc:
Thanks!

PR #332, opened 2017-09-25
Removes duplicate if statements in test

manniwood:
While reading TestTimestampTzTranscode in values_test.go, I noticed a setup and its if statements were run twice. This removes the duplication.


jackc:
Thanks.

Issue #324, opened 2017-09-18
Send multiple times with batch queries


plopik:
I have lots (30 for now, more are coming) of small queries that take input values from the preceding query. I was hoping that Batch might help lower the cost of connection and the transactionality.

I'd like to use Batch.Send() multiple times for this. Here is how it might look:
```
batch := db.BeginBatchNoCommit()
batch.Queue("insert (value) in table1 ($1) returning id", "foo")
batch.Send()
batch.QueryRowResults().Scan(&ID)
batch.Queue("insert (id,value2) in table2 ($1,$2)", ID, "bar1")
batch.Queue("insert (id,value2) in table3 ($1,$2)", ID, "bar2")
batch.Send()
batch.Commit()
batch.Close()
```

I'm about to fork and suggest a PR for this. What do you think? Stop me if that's a bad idea ;-)

plopik:
Other solutions would be to create a BeginBatch on the transaction object, or add a methode "SendWithoutCommit" to the batch object. any hindsight ?

jackc:
Let me make sure I understand the use case.

Do all the queries have to be in a single transaction? If not you could use multiple batches.

Depending on the exact nature of the queries, a writable CTE might be appropriate for chaining them all together in a single statement.

But assuming neither of those will work that leads to changing the batch transaction strategy. The reason batches implicitly use transactions is the difficulty in determining what queries took affect on the server if a batch is interrupted. I suppose with suitable warning labels this could be disabled.

I think I would prefer an API that uses a new batch for each send rather than reusing the batch. Your idea of adding BeginBatch to Tx seems promising. Then the only change to the Batch interface is to ignore the txOptions arg for Send when it was created by a Tx.

plopik:
```Do all the queries have to be in a single transaction``` 
yes that the use case.

A CTE would be possible but in some cases they are hard to read maintain and test and a batch of multiple queries feels better.

```I think I would prefer an API that uses a new batch for each send rather than reusing the batch. Your idea of adding BeginBatch to Tx seems promising. Then the only change to the Batch interface is to ignore the txOptions arg for Send when it was created by a Tx.``` 
after further thinking I also think this is better I'll send you a PR soon.

jackc:
#308 has a similar issue. Ideally, whatever we do could handle both cases, but at least let's ensure that solving either one doesn't hinder the other.

jackc:
Resolved by @plopik in PR #333.

PR #333, opened 2017-09-25
New beginBatch on transaction object

plopik:
A batch on a tx object does not open and close a transaction itself and
instead use the tx object to ensure the transactionality of the batch

remove unused boolean 'sent' in batch struct

see #324


jackc:
Thanks!

Issue #320, opened 2017-09-09
Support CrateDB


felixge:
Hi,

thanks for this great library :).

I'm currently working on implementing [CrateDB support for influxdata/telegraf](https://github.com/influxdata/telegraf/pull/3210) and was asked by the maintainers if I could use your library instead of `lib/pq`.

Unfortunately it doesn't work because [pgx queries the postgres catalog](https://github.com/jackc/pgx/blob/master/conn.go#L387-L393) when creating a new connection, and CrateDB doesn't implement the catalog tables.

Please let me know how you feel about supporting CrateDB (or other databases pretending to be postgres on the protocol level). If you're interested, I might be able to contribute a patch. If not, feel free to simply close this issue.

Cheers
Felix

PS: I'm a huge fan of your writing. Your post [Materialized View Strategies Using PostgreSQL](https://hashrocket.com/blog/posts/materialized-view-strategies-using-postgresql) gave me the inspiration to migrate a large project (> 1 billion rows) from NoSQL to PostgreSQL. The migration was a huge success, and the analytics queries are blazing fast thanks to to some eager materialized views :).

jackc:
I'm in favor of supporting PostgreSQL-like servers provided the changes required can be cleanly implemented. This has recently come up with RedShift SSL issues as well. In both these cases, it appears that the changes required would be localized to the connection code so it seems promising.

> PS: I'm a huge fan of your writing. Your post Materialized View Strategies Using PostgreSQL gave me the inspiration to migrate a large project (> 1 billion rows) from NoSQL to PostgreSQL. The migration was a huge success, and the analytics queries are blazing fast thanks to to some eager materialized views :).

Thanks! So glad to hear it was helpful. I might have to write an update after PG10 is released. Transition table triggers have the potential to be much more efficient for certain operations.

jackc:
Resolved by #336.

chetananand:
I tried the version of pgx with this patch - I still run into a similar error on https://github.com/jackc/pgx/blob/master/conn.go#L430 from initConnInfoEnumArray(). That query also is using the pg_type table without the schema pg_catalog and therefore initConnInfo() will still fail.

felixge:
[...]

PR #336, opened 2017-09-27
Support CrateDB

felixge:
See https://github.com/jackc/pgx/issues/320

@jackc I spend a decent amount of time trying to come up with a minimally invasive patch and added some detailed in-line comments explaining my approach. Let me know what you think.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -41,10 +41,11 @@ var minimalConnInfo *pgtype.ConnInfo
 func init() {
 	minimalConnInfo = pgtype.NewConnInfo()
 	minimalConnInfo.InitializeDataTypes(map[string]pgtype.OID{
-		"int4": pgtype.Int4OID,
-		"name": pgtype.NameOID,
-		"oid":  pgtype.OIDOID,
-		"text": pgtype.TextOID,
+		"int4":    pgtype.Int4OID,
+		"name":    pgtype.NameOID,
+		"oid":     pgtype.OIDOID,
+		"text":    pgtype.TextOID,
+		"varchar": pgtype.VarcharOID,
 	})
 }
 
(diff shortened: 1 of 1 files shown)


jackc:
Thanks! Looks like as clean and narrow a change as we could hope for.

felixge:
Thanks a lot! I'll be around to help if anybody runs into any unexpected issues with this.

mfussenegger:
Nice to see CrateDB support in pgx.

One question to the `pg_type` query: Would it be possible to change this in
`pgx` to query explicitly on `pg_catalog.pg_type`?
I'm not sure if this could break anything with older postgres versions. Afaik
9+ should be fine.

It would be fairly simple for us to add the `typtype` column, but extending the
schema lookup logic is a bit more work.

The reason I'm asking is that this way we could add `typtype` and then no
longer have to worry about breaking anything here if we change something with
the error codes / error messages.

felixge:
@mfussenegger not sure about @jackc, but from my POV changing the query to explicitly target the pg_catalog schema should be a reasonable change. As far as I can tell the pg_catalog schema has been around and can be explicitly specified in queries since at least [Postgres 7.3](https://www.postgresql.org/docs/7.3/static/ddl-schemas.html#AEN2160).

That being said, I think the workaround code should probably still need to stay in place to support older CrateDB versions for a while.

jackc:
No objections from me for explicitly qualifying the `pg_catalog` schema.

felixge:
@mfussenegger let me know when `typtype` gets implemented in CrateDB (is there a GH issue to subscribe to?). Then I'll do the PR and testing for pgx.

mfussenegger:
@felixge We've added the column (see https://github.com/crate/crate/pull/6582) - will be part of 2.3 (or in the nightly builds starting tomorrow)

felixge:
@mfussenegger cool. I'm traveling for the next three weeks, but I'll try to find some time to make a PR for this ASAP.

PR #340, opened 2017-10-10
Use logrus.FieldLogger instead of *logrus.Logger

ligustah:
This allows supplying a logrus logger that already has fields configured


jackc:
Thanks!

ligustah:
Thanks for merging (and your fantastic library!) 👍

PR #346, opened 2017-10-26
CopyFromReader functionality and basic tests

silviucm:
Hi Jack,

We conversed a few years ago, regarding a [structures convenience generator](https://www.cmscomputing.com/articles/programming/generate-go-entities-from-postgres-tables-views) I built on top of your excellent library.

Numerous use cases of bulk import provide data as a csv stream, be it a network service or a local file.
The current pull request attempts to make it easy to perform bulk copy directly from an io.Reader interface input, obviating the need to go through the [][]interface{} payload.

The input format is csv-based, at times relaxed, at times opinionated (particularly in regards to date and timestamp formats). This can be further extended if needed, for example creating a pg interval type parser from, given the range limitations of the Go time.Duration type.

Alternatively, if you're ok with the code in principle, but you're not comfortable with the location, I can move them into a copy_from_extras.go 

Cheers,
Silviu

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -2,10 +2,15 @@ package pgx
 
 import (
 	"bytes"
+	"encoding/csv"
 	"fmt"
+	"io"
+	"strings"
+	"time"
 
 	"github.com/jackc/pgx/pgio"
 	"github.com/jackc/pgx/pgproto3"
+	"github.com/jackc/pgx/pgtype"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


jackc:
Is the source CSV compatible with PostgreSQL's built-in COPY / CSV functionality? As I understand it this code reads in a CSV stream, then parses it and formats it as binary values suitable for the binary copy protocol.  Is that because pgx doesn't expose the text-based CSV copy protocol or your source CSV is incompatible with PostgreSQL's CSV importer?

If the former, then maybe we should add CSV text protocol copy support to pgx. That could make this nearly as simple as copying from the io.Reader to pgx's net.Conn. If the latter, we'd have to think carefully about any differences with what someone would expect from doing the import with a `\copy` in `psql`.

silviucm:
@jackc you are correct: The code attempts to "guess" the closest possible binary encoding representation of each parsed value, then use the binary copy. This pull request was indeed a corollary of specific problems I had to solve, so you can say it's a very opinionated subset of PostgreSQL's CSV format.

For example, as opposed to the official one-and-only null placeholder from the [PostgreSQL docs](https://www.postgresql.org/docs/9.6/static/sql-copy.html):
	

> NULL
> 		Specifies the string that represents a null value. The default is \N (backslash-N) in text format, and an unquoted empty string in CSV format. You might prefer an empty string even in text format for cases where you don't want to distinguish nulls from empty strings. This option is not allowed when using binary format.

I am allowing several, defined inside a []string (e.g. {"[NULL]", "[NIL]"})

I felt that it would serve a lot of other users, for generic import use cases, but it is not adhering to the COPY specs page per se.

In the end, I ended up [integrating the proposed code inside the generated Go code that my package emits](https://www.cmscomputing.com/wiki/pgtogogen/bulk-copy), so I have less of an urgency, for my own stuff.

Now that I think of it, you are right, in the sense that we should turn this proposal on its head, and see how it could benefit the larger majority of pgx users.

I am only going to spew out my preliminary thoughts regarding going the csv -> binary encoding path, because I spent some effort in that direction.

- The pgType nameValues map is unexported:	https://github.com/jackc/pgx/blob/master/pgtype/pgtype.go#L208

- I ended up with a local function but maybe we can consider an exported function inside pgtype:
	
	`func ToPgtype(dbtype string, status pgtype.Status, val ...interface{}) (pgtype.Value, error)`

  The status parameter would allow to be very explicit about the status and remove any nullability ambiguity, which later on, cannot be set with the Set method on the Value interface.
  The variadic val would allow setting the value on the spot, or (if not provided) just getting a skeleton Value instance that can subsequently populate with Set

  Example of usage for the first column of: "abc",,"",etc"

  	pv, err := pgType.ToPgtype("character varying", pgtype.Present, "abc")

  For the second column:

  	pv, err := pgType.ToPgtype("character varying", pgtype.Null, "abc")

For the csv options, I started with a few distinct parameters but in the end I ended with a [*CopyFromReaderOptions struct](https://www.cmscomputing.com/wiki/pgtogogen/bulk-copy), which can contain relevant properties such as nullability treatment, the date time format, the separator character, and where I can keep adding options.

(Needless to say, I am aware that, as a derivative project of pgx, I can sort of afford to gently keep breaking compatibility, whereas in pgx, you need to at least keep it stable throughout versions)

Shall we close this pull request for now, or keep it open and continue the discussion towards eventually implementing something in here ?

Cheers
Silviu

jackc:
I think it makes sense to keep this external to pgx at least for now.

With regards to `pgtype.nameValues` not being exported, it is just the initial, hard-coded set of PostgreSQL type names to pgtype.Value. But it is not all inclusive. Other values can be registered automatically as part of the connection process and users can manually register types. I think a better solution reading from the `ConnInfo` on a `Conn` to get the final values. It lets you look up data types by OID, PostgreSQL type name, or `pgtype.Value`. I think that would do what you need. It doesn't allow enumeration of all types, but we could easily add that if it was needed. Open a separate issue for this if ConnInfo doesn't suffice.

For the proposed function `ToPgtype`, I'm not sure I understand the overall intent. `Set` should accept `nil` and set `Status` appropriately. I suppose it's not guaranteed by the interface, but in practice that is how it (should) work. If there's more to it, open an issue for this proposed func and we can work on it there.

PR #354, opened 2017-11-12
travis: add connection test coverage for cratedb

felixge:
@jackc let me know what you think. This makes sure the connection code + a simple query works against CrateDB. This PR should also simplify the process of adding more supported databases in the future.


jackc:
LGTM - thanks!

Issue #331, opened 2017-09-24
add ability to disable implicitly prepared queries.


james-lawrence:
it looks like pgx implicitly prepares all queries particularly within the stdlib interface, which is both surprising and can sometimes be undesirable. I understand that pgx gets its performance boost from using them implicitly.

Would we be open to making this behaviour configurable?

Details:
ran into this while debugging pgbouncer queries returning a `ERROR: unsupported pkt type: 80 (SQLSTATE 08P01)` when querying pgbouncer's internal state. I'm assuming the type is the ascii code for message format's first byte, which would make it a [parse message](https://www.postgresql.org/docs/current/static/protocol-message-formats.html).
```golang
func (c *Conn) Query(query string, argsV []driver.Value) (driver.Rows, error) {
	if !c.conn.IsAlive() {
		return nil, driver.ErrBadConn
	}

	ps, err := c.conn.Prepare("", query)
	if err != nil {
		return nil, err
	}

	restrictBinaryToDatabaseSqlTypes(ps)

	return c.queryPrepared("", argsV)
}
```

jackc:
I think that is a reasonable option.

There is already support for using the simple protocol / not using prepared statements via the `QueryEx` method. To do this by default we would need an additional option in `ConnConfig` -- something like `DisableImplicitPrepare`. Then `QueryEx` and `execEx` would need to to consider that flag in addition to the `options.SimpleProtocol`.

To use this new functionality, the stdlib connection would need to be opened with that configuration option. That can be done with `stdlib.RegisterDriverConfig`.

james-lawrence:
k I'll take a look and see about generating a PR at somepoint.

james-lawrence:
I started looking into this, not sure how to handle the call to restrict binary types as it is attached to a prepared statement. any suggestions?

james-lawrence:
this is the current query function for stdlib:
```golang
func (c *Conn) Query(query string, argsV []driver.Value) (driver.Rows, error) {
	if !c.conn.IsAlive() {
		return nil, driver.ErrBadConn
	}

	ps, err := c.conn.Prepare("", query)
	if err != nil {
		return nil, err
	}

	restrictBinaryToDatabaseSqlTypes(ps)

	return c.queryPrepared("", argsV)
}```

[...]

Issue #344, opened 2017-10-19
pgbouncer transaction pooling mode


fmstephe:
Our application produces errors on every query when connected to pgbouncer which is set to use `transaction` pooling. Is it expected that pgx can communicate with pgbouncer in this pooling mode?

fmstephe:
A further update to this issue. We determined that the likely reason for failure was that when calling `(*ConnPool).Query(queryStr, args...)` pgx will use prepared statements by default.

I have tried to switch to using `(*ConnPool).QueryEx(context.Background(), queryStr, &pgx.QueryExOptions{SimpleProtocol: true}, args...)` but then we get these kinds of error. 

`failed: Scan received wrong number of arguments, got 0 but expected 2 trace:`

fmstephe:
@galdor currently we have given up on transaction/session pooling modes. But I would really like to use them. If I get the time to dive back in and try to find a solution I will post it here.

galdor:
We also went back to session mode. I'll keep the ticket updated if we find a better solution.

kelseyfrancis:
The issue might be that that [pgx itself makes a few queries after it connects (to look up types)](https://github.com/jackc/pgx/blob/master/conn.go#L385), and those would get prepared first, like any other query without its parameter oids supplied explicitly or not using the simple protocol.

fmstephe:
Thanks for the pointer @kelseyfrancis when I get to this I will look there first.

james-lawrence:
duplicate of #331

jackc:
Not sure if this is still relevant to the OP, but this can be handled with `PreferSimpleProtocol` in `v3`. In `v4` a describe mode statement cache can be used to still be able to use the extended protocol.

PR #359, opened 2017-12-02
support querying without generating prepared statements

james-lawrence:
work related to #331, as a starting point for discussion around how to achieve making a query in pgx without passing through a prepared statement.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -40,12 +40,12 @@ func TestConnBeginBatch(t *testing.T) {
 	batch.Queue("select id, description, amount from ledger order by id",
 		nil,
 		nil,
-		[]int16{pgx.BinaryFormatCode, pgx.TextFormatCode, pgx.BinaryFormatCode},
+		[]int16{pgtype.BinaryFormatCode, pgtype.TextFormatCode, pgtype.BinaryFormatCode},
 	)
 	batch.Queue("select sum(amount) from ledger",
 		nil,
 		nil,
-		[]int16{pgx.BinaryFormatCode},
+		[]int16{pgtype.BinaryFormatCode},
 	)
 
 	err := batch.Send(context.Background(), nil)
@@ -167,7 +167,7 @@ func TestConnBeginBatchWithPreparedStatement(t *testing.T) {
 		batch.Queue("ps1",
 			[]interface{}{5},
 			nil,
-			[]int16{pgx.BinaryFormatCode},
+			[]int16{pgtype.BinaryFormatCode},
 		)
 	}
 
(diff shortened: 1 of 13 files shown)


james-lawrence:
this doesn't get us entirely to being able to connect to PG bouncer due to how the connection setup works. but I have a follow up PR that brings it the rest of the way.

james-lawrence:
@jackc this is ready for a review.

fixes #331 probably fixes #344, and allows basic use with pgbouncer.

also lays a better foundation for supporting pg-like databases.

jackc:
First of all, thanks for working on this.

There's a lot of changes here, and I'm not sure I get how they all fit together. It feels like there might really be several distinct PRs / changes here that are bundled together.

With that said, here are my thoughts so far.

1. Regarding moving pgx.TextFormatCode and pgx.BinaryFormatCode to pgtype package: this would be a breaking change for any code relying on it existing in the pgx package. So if we need those constants available in pgtype we would need to copy them and leave the originals alone. That could also reduce the size of the changeset as all the existing internal references to them could remain unchanged. I would probably make them unexported constants of pgtype so as not to increase the surface area of the package. In addition, it appears the reason for this was to add the preferred format code to DataType. While this might be an interesting refactor I didn't see anywhere that it was a prerequisite to any of the other features. So if I am correct that it isn't required by the other features I would prefer to consider this as a separate PR. (Another possibility -- [Preferred]FormatCode could be a method of DataType -- it would subsume the DetermineFormatCode reducing the interface of the package as well)

2. I haven't used PgBouncer, so I might be missing something, but wouldn't `NewPGBouncerConnInfo` render PGBouncer connections unable to use any types that do not have hard-coded OIDs (hstore for example). I would expect it still to be possible to query the server for a list of types even when being proxied through PGBouncer.

3. I'd prefer not to expose a complete `*pgx.QueryExOptions` in `DriverConfig`. Every field except `SimpleProtocol` is query specific. I think it would be confusing to expose config options that could break the connection if they were used. But at a more fundamental level, forcing the entire connection to default to the simple protocol strikes me as something that should be implemented in the core pgx package and enabled / configured from the stdlib package. I just did a quick proof-of-concept in branch prefer-simple-protocol (ce7014d372f12f3e4533fa8b84a1c2e6f395f82d).

4. I like the idea of an extension point for PG-like connections. I _think_ I like the `CustomConnInfo` callback function idea, but this would also be a lot easier to reason about in isolation.

james-lawrence:
>There's a lot of changes here, and I'm not sure I get how they all fit together. It feels like there might really be several distinct PRs / changes here that are bundled together.

Totally agree! I was originally thinking about splitting into 3 distinct parts. I just decided to dump everything on you to see what you liked vs didn't like. I prefer smaller PRs myself.

tl;dr; I'll split up the PR into distinct parts with your comments taken into consideration.

> Regarding moving pgx.TextFormatCode and pgx.BinaryFormatCode ....

you are correct, I noticed it while working on this work, if we're okay with leaving unexported constants in the pgtype package I can do the rest without it.

> haven't used PgBouncer, so I might be missing something...

the distinction is connecting to PGBouncer vs connecting through PGBouncer, this isn't about connecting through pgbouncer, which as you stated you would still be able to query the list of types.
its about connecting directly to pgbouncer to query information about the process from its internal metrics.

>  I'd prefer not to expose a complete *pgx.QueryExOptions

fair enough, this particular change was there just to get a working example, I expected some push back there but didn't want to guess on the direction you'd have preferred.

> I like the idea of an extension point for PG-like connections....

I'll pull it out as a separate PR, its causing the errors in travis I have the fallback logic to cratedb incorrectly implemented apparently and need to revisit it.

PR #364, opened 2017-12-16
extension point for PG-like connections

james-lawrence:
allows for overriding initConnInfo behaviour, useful for crate,pgpool, pgbouncer.

see #359 for originating PR.


jackc:
Thanks! Sorry it took so long to get to this.

james-lawrence:
NP, we all have real lives. =)

PR #365, opened 2017-12-16
support query without prepared statements

james-lawrence:
allows querying without prepared statements from stdlib driver.

minimal PR. work extracted from #359

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -74,6 +74,17 @@ type ConnConfig struct {
 	Dial              DialFunc
 	RuntimeParams     map[string]string // Run-time parameters to set on connection as session default values (e.g. search_path or application_name)
 	OnNotice          NoticeHandler     // Callback function called when a notice response is received.
+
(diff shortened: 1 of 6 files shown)


jackc:
I decided to go a bit different direction on this. At the expense of a slightly more awkward construct in the stdlib package it was possible to avoid needing to change any of the other packages. I was still able to use your test.

Simple protocol / no automatic prepared queries should work in stdlib now.

james-lawrence:
@jackc I'll retest against pgbouncer, but it is unlikely to have worked.

the changes around `readFieldDescriptions` were critical for it to work properly for pgbouncer IIRC.

and the format code stuff I think actually makes the code easier to reason about. it also avoids doing the type switch for each query, which can be a relatively expensive operation all things said.

james-lawrence:
well I take the previous comment back, can't reproduce atm.

PR #366, opened 2017-12-17
Handle timeout parameters

cyberdelia:
Handle `connect_timeout` and `PGCONNECT_TIMEOUT` parameter and pass it to the default `Dialer`. This is written to mimic libpq behaviour as well.


jackc:
Thanks. Merged with the change of parsing into a Dial func with embedded Timeout instead of a Timeout field on ConnConfig.

PR #367, opened 2017-12-17
Postgres 9.2 is not supported since September 2017

cyberdelia:
Postgres 9.2 has reached the end of its life, so don't skip testing compatibility for it.

See: https://www.postgresql.org/support/versioning/


jackc:
Thanks

PR #368, opened 2017-12-19
Add support for bpchar type

aka-rider:
Hi. First of all: kudos for the great library.

Here is support for the bpchar type -- blank padded, fixed-length (`character(n)`, `char(n)`)


jackc:
Thanks. Sorry for the delay getting to this.

PR #369, opened 2017-12-20
Return error on unknown oid while decoding record instead of panic

aka-rider:


jackc:
Thanks. Merged with addition of test case.

PR #385, opened 2018-02-02
Make TLS parsing compatible with lib/pq...

sean-:
This PR fixes this so that `jackc/pgx` can be used by `database/sql` for TLS-enabled PostgreSQL servers (either PostgreSQL or CockroachDB).

Previously `stdlib.DriverConfig().ConnectionString()` would fail to use any of `lib/pq`'s TLS parameters in either a DSN or URI, rendering the `stdlib` shim useless for TLS workloads (note: `jackc/pgx` worked for TLS-enabled servers, just not the `stdlib` shim).

See `conn_config_test.go.example` for the example to test the new `sslrootcert`, `sslcert`, and `sslkey` keywords for PostgreSQL DSNs and URIs (these keywords are the same as [lib/pq](https://github.com/lib/pq/blob/88edab0803230a3898347e77b474f8c1820a1f20/conn.go#L1067)).

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -703,43 +703,15 @@ func ParseURI(uri string) (ConnConfig, error) {
 		cp.Dial = d.Dial
 	}
 
-	err = configTLS(url.Query().Get("sslmode"), &cp)
+	tlsArgs := configTLSArgs{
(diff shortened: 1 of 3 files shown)


jackc:
Thanks!

PR #389, opened 2018-02-05
Fix syntax error in replication protocol msg issued by StartReplication()

steampunkcoder:
Fix syntax error in replication protocol msg issued by StartReplication().
According to https://www.postgresql.org/docs/9.6/static/protocol-replication.html
pluginArguments should be separated by commas and surrounded by parantheses.
Manually tested against Postgres 9.6 server.


jackc:
Thanks!

PR #400, opened 2018-03-20
Add zerolog logging adapter

cheapRoc:
This PR introduces a new logging adapter for [github.com/rs/zerolog](https://github.com/rs/zerolog).


cheapRoc:
@jackc Looks like the build is passing now. Hope this is useful, it has been in our application.

jackc:
Thanks!

cheapRoc:
One thing to note, that I'm working on, is that long integers are displaying as floats. This is probably not an issue for anyone outside using CockroachDB which generates long integers for serial primary keys.

Just to note that the issue is open over at rs/zerolog#46.

PR #401, opened 2018-03-26
implement json.(Un)Marshaler for primary type

eruca:
I think json is very popular in go, so if the primary type implement the json.(Un)Marshaler, it will very convenient to Marshal or Unmarshal without writting a lot code to make it. I did not know the array type, could you implement them?

### DIFF
--- a/pgtype/bytea.go
+++ b/pgtype/bytea.go
@@ -154,3 +154,25 @@ func (src *Bytea) Value() (driver.Value, error) {
 		return nil, errUndefined
 	}
 }
+
+func (src *Bytea) MarshalJSON() ([]byte, error) {
+	switch src.Status {
+	case Present:
+		return src.Bytes, nil
+	case Null:
+		return nil, nil
+	default:
+		return nil, errUndefined
+	}
+}
+
+func (dst *Bytea) UnmarshalJSON(b []byte) error {
+	if b == nil {
+		dst.Status = Null
+		return nil
+	}
+
+	dst.Bytes = b
+	dst.Status = Present
+	return nil
+}
--- a/pgtype/bytea.go
+++ b/pgtype/bytea.go
@@ -154,25 +154,3 @@ func (src *Bytea) Value() (driver.Value, error) {
 		return nil, errUndefined
 	}
 }
-
-func (src *Bytea) MarshalJSON() ([]byte, error) {
-	switch src.Status {
-	case Present:
-		return src.Bytes, nil
-	case Null:
-		return nil, nil
-	default:
-		return nil, errUndefined
-	}
-}
-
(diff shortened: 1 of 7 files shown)


pgtype/bytea.go
```diff
@@ -154,3 +154,25 @@ func (src *Bytea) Value() (driver.Value, error) {
 		return nil, errUndefined
 	}
 }
+
+func (src *Bytea) MarshalJSON() ([]byte, error) {
+	switch src.Status {
+	case Present:
+		return src.Bytes, nil
```
jackc: Raw bytes don't necessarily have a valid JSON representation, so I'm not sure if it makes sense to implement JSON marshaling.

pgtype/float4.go
```diff
@@ -195,3 +196,27 @@ func (src *Float4) Value() (driver.Value, error) {
 		return nil, errUndefined
 	}
 }
+
+func (src *Float4) MarshalJSON() ([]byte, error) {
+	switch src.Status {
+	case Present:
+		buf := make([]byte, 4)
+		binary.BigEndian.PutUint32(buf, math.Float32bits(src.Float))
+		return buf, nil
```
jackc: I don't think this is right. Shouldn't this return the text of the number, not the raw bits?

pgtype/float8.go
```diff
@@ -185,3 +186,27 @@ func (src *Float8) Value() (driver.Value, error) {
 		return nil, errUndefined
 	}
 }
+
+func (src *Float8) MarshalJSON() ([]byte, error) {
+	switch src.Status {
+	case Present:
+		buf := make([]byte, 8)
+		binary.BigEndian.PutUint64(buf, math.Float64bits(src.Float))
+		return buf, nil
```
jackc: Same here. I think this should be text, not raw bits.

pgtype/json.go
```diff
@@ -159,3 +159,24 @@ func (src *JSON) Value() (driver.Value, error) {
...
+
+func (dst *JSON) UnmarshalJSON(b []byte) error {
+	if b == nil {
+		*dst = JSON{Status: Null}
+		return nil
+	}
+
+	*dst = JSON{Bytes: b, Status: Present}
+	return nil
+}
```
jackc: I'm not sure what the semantics of this are / should be. What does it mean to have JSON inside of JSON?

eruca: thanks for your review; the input bytes should be the json byte array, need check; I should consider that. I think the b represent the JSON's content.
I sometimes build my struct like:
```go
type Foo struct {
   Name string `json:"name"`
   Meta pgtype.JSONB `json:"meta"`
}
```
and the struct will be serialize to json to interaction with the client, so if the struct implement the json interface, I do not write the json.Marshal; I think I have no power to contribute, sorry about that

PR #410, opened 2018-04-12
stdlib: allow nested database/sql/driver.Drivers

jbowens:
database/sql/driver.Driver implementations can be nested, with each
layer adding additional functionality. If pgx/stdlib.Driver is wrapped
in another driver.Driver implementation, AcquireConn will error,
detecting that the *sql.DB's driver is not (directly) pgx.Driver.

I'd like AcquireConn to support retrieving a connection from a *sql.DB
that wraps pgx.Driver with another driver.Driver implementation.
It looks like it should be possible, but it requires using a global map of
fakeTxConns instead of a per-Driver map of fakeTxConns.

This change also means that in the ErrNotPgx case, a SQL transaction
will be begun and rolled back. This seems reasonable to me. I can't
think of a use case where calling AcquireConn on non-pgx *sql.DBs
is a part of a frequent code path.

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -99,9 +99,9 @@ var ErrNotPgx = errors.New("not pgx *sql.DB")
 
 func init() {
 	pgxDriver = &Driver{
-		configs:     make(map[int64]*DriverConfig),
-		fakeTxConns: make(map[*pgx.Conn]*sql.Tx),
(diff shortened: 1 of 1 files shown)


jbowens:
PTAL

jackc:
I wasn't aware that nesting of database/sql/driver.Driver was a thing. But the change seems acceptable.

PR #414, opened 2018-04-24
uint should be parsed using strconv.ParseUint

regeda:
If I have uint64 bigger than int64 (for example "9926611508173733888"), pgx returns an error: strconv.ParseInt: parsing "9926611508173733888": value out of range.

### DIFF
--- a/pgtype/convert.go
+++ b/pgtype/convert.go
@@ -185,6 +185,98 @@ func underlyingSliceType(val interface{}) (interface{}, bool) {
 	return nil, false
 }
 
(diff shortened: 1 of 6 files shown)


jackc:
As far as I know PostgreSQL does not support unsigned integers -- the largest integer is signed 64-bit. Where is the unsigned 64-bit int coming from or going to?

regeda:
I can define an array of numeric and hold unsigned numbers. Then I want to scan it in `[]uint64`.

PR #415, opened 2018-04-24
numeric array `AssignTo` supports both types int64 and uint64

regeda:
I want to assign `NumericArray` to `[]uint64`.


regeda:
@jackc Sorry for my previous pull request. After deeper diving into `pgtype`, I have found that`NumericArray` is appropriated type for `[]uint` scanning.

regeda:
@jackc Would you merge the request in near time? It's urgent for my project and I want to use the upstream repository.

jackc:
Thanks!

PR #424, opened 2018-05-17
Allow setting nil to pgtype.Bool

### LINKED ISSUE #423: Why pgtype.Bool forbids setting to nil?
**Problem:**
I can't set `Bool` type to `nil`. It returns `cannot convert <nil> to Bool` error.

Why is it not allowed? If there is no particular reason, I can contribute it with something similar to Int8.
https://github.com/jackc/pgx/blob/master/pgtype/int8.go#L20

Example code:
```
r := &pgtype.Bool{}
err := r.Set(nil)
So(err, ShouldBeNil)
```

tarikdemirci:
Fixes #423

### DIFF
--- a/pgtype/bool.go
+++ b/pgtype/bool.go
@@ -13,6 +13,11 @@ type Bool struct {
 }
 
 func (dst *Bool) Set(src interface{}) error {
+	if src == nil {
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #426, opened 2018-06-04
Add the type of the value in all decode error messages

dvic:
This improves the decode error message, e.g., from

`can't scan into dest[3]: cannot decode &{xyz 2} into string` to

`can't scan into dest[3]: cannot decode &pgtype.Text{String:"xyz", Status:0x2} into string`

(this is an example of me forgetting to take the address of a variable when using `Scan` 😆)

### DIFF
--- a/pgtype/aclitem.go
+++ b/pgtype/aclitem.go
@@ -70,7 +70,7 @@ func (src *ACLItem) AssignTo(dst interface{}) error {
 		return NullAssignTo(dst)
 	}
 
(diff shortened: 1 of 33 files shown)


dvic:
Fails with a data race on PG 9.4? Is this a known issue?

jackc:
Thanks.

Data race was due to altering a tls.Config in a test. Should be fixed as of 39bbc98d99d7b666759f84514859becf8067128f.

PR #438, opened 2018-07-30
Add CopyFromReader and CopyToWriter methods

ikitiki:
Add methods to allow copy from/to using textual format

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -2,6 +2,7 @@ package pgx
 
 import (
 	"context"
+	"io"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


jackc:
Functionality looks good. I'd like you to consider another name though. 

The fact that the existing copy code only uses binary is just an implementation detail. It may support the text format in the future. And the new code you wrote actually works with the binary format as well as text format.

The big difference with this new copy functionality vs. the old is this works with streams / blobs of copy data without interpreting it.  My initial thought for improved names is `CopyFromReader` and `CopyToWriter`.

One other minor question: Does `CopyFrom*` need the `args` param? I realize `CopyTo*` would because you can pass an entire query as part of the copy statement. But I can't see any way that `CopyFrom*` needs escaped values.

ikitiki:
>I'd like you to consider another name though.

Totally agree, I've changed the names of the methods to the proposed `CopyFromReader` and `CopyToWriter`

> Does `CopyFrom*` need the args param?

Actually it doesn't. I've removed it.



Besides, I've added Copy method to the Tx struct.

jackc:
Thanks!

PR #444, opened 2018-08-03
fix: support driver.Valuer on pointer receivers

### LINKED ISSUE #443: `driver.Valuer` not handled properly on pointer receivers
Because the `encodePreparedStatementArgument` [dereferences the pointer](https://github.com/jackc/pgx/blob/master/values.go#L165) before checking for the `driver.Valuer` interface, it will never be used for types which define `Value()` on a pointer receiver.

i.e. this works:
```go
type MyType struct {}
func (m MyType) Value() (driver.Value, error)
```
but this doesn't
```go
type MyType struct {}
func (m *MyType) Value() (driver.Value, error)
```

johngibb:
Fixes #443.

### DIFF
--- a/values.go
+++ b/values.go
@@ -158,6 +158,12 @@ func encodePreparedStatementArgument(ci *pgtype.ConnInfo, buf []byte, oid pgtype
 		buf = pgio.AppendInt32(buf, int32(len(arg)))
 		buf = append(buf, arg...)
 		return buf, nil
+	case driver.Valuer:
+		v, err := callValuerValue(arg)
+		if err != nil {
+			return nil, err
+		}
+		return encodePreparedStatementArgument(ci, buf, oid, v)
 	}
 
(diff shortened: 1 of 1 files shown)


jackc:
The test failure appears to be because this change causes types that pgx can natively encode in binary, but also implement driver.Valuer to use the driver.Valuer implementation to build the data to send over the wire. But `chooseParameterFormatCode` still sees the type as something that should be encoded into binary. Hence, we get a mismatch.

bfontaine:
@johngibb I know this PR is old, but would you mind fixing the merge conflicts as well as the failing test so we can merge it?

kbrgl:
@jackc Is this going to make it into the library? The fact that pgx doesn't support pointer receivers for Valuer is a pretty major footgun. It would be great to have a note in the docs about this. (My apologies if there's already one, but I couldn't find it.)

PR #455, opened 2018-09-09
Pool should work like a queue 

regeda:
Currently, the pool works like a stack. It leads to using a stolen connection from a head when a spike of requests came.
I think [the stdlib implementation](https://golang.org/src/database/sql/sql.go#L1096) maintains available connections better. The queue holds connections in an actual state all the time.

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -116,10 +116,14 @@ func (p *ConnPool) acquire(deadline *time.Time) (*Conn, error) {
 	}
 
 	// A connection is available
-	if len(p.availableConnections) > 0 {
-		c := p.availableConnections[len(p.availableConnections)-1]
+	// The pool works like a queue. Available connection will be returned
+	// from the head. A new connection will be added to the tail.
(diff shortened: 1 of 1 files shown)


jackc:
What you mean by a "stolen connection"?

regeda:
We have a plenty of Postgres instances in "read-only" setup. All of them connected to a balancer (AWS ELB). We have a pool with 5 max idle connections. Just 1-2 connections are in-use state most of the time.
But when a spike of requests came we got "EOF" from a balancer. Hence old connections already disconnected and the balancer requested nothing.

jackc:
So if I understand correctly, what is happening in the existing system is the connections at the bottom of the stack are idle long enough during slow periods that they are disconnected. When a busy period occurs those silently dead connections cause errors.

Changing the pool usage to a queue instead of a stack means that usage is distributed among all the connections making it less likely that a connection will be idle long enough to die.

Am I understanding this correctly?

regeda:
Correctly.

PR #456, opened 2018-09-15
Call context.WithTimeout cancel function

tehbrut:
Call cancel function returned by context.WithTimeout, to avoid a memory leak.


jackc:
Thanks!

PR #463, opened 2018-10-08
Handle (n > 0 and err == io.EOF) in CopyFromReader

fpetrini:
The CopyFromReader method does not correctly handle the case when (n > 0 and err == io.EOF). This is a valid case as per the io.Reader documentation and should be handled.

When reading a gzipped file using gzip.Reader to uncompress data on the fly the last chunk of data is often omitted due to the CopyFromReader method breaking out of the read loop on io.EOF and not processing the last bit of data. The result is that the COPY fails or that not all rows are inserted.
I've added a test case reading from a gzip.Reader which fails without the fix.


jackc:
Good catch. Thanks!

PR #464, opened 2018-10-09
implement TextDecoder interface for the qchar type

mkabilov:


jackc:
Did you see the comments on this type?

```go
// QChar is for PostgreSQL's special 8-bit-only "char" type more akin to the C
// language's char type, or Go's byte type. (Note that the name in PostgreSQL
// itself is "char", in double-quotes, and not char.) It gets used a lot in
// PostgreSQL's system tables to hold a single ASCII character value (eg
// pg_class.relkind). It is named Qchar for quoted char to disambiguate from SQL
// standard type char.
//
// Not all possible values of QChar are representable in the text format.
// Therefore, QChar does not implement TextEncoder and TextDecoder. In
// addition, database/sql Scanner and database/sql/driver Value are not
// implemented.
```

I'm trying to remember why I originally skipped the text format... without success. Looking at the PG source `src/backend/utils/adt/char.c` it appears that the PG server does support it. Oh well, maybe I was just wrong. If it does make sense to support the text format for this, would you mind adding `EncodeText` and correcting the comment as well?

bfontaine:
Almost 5 years later, does it make sense to keep this PR open?

PR #466, opened 2018-10-11
Add NewConnConfig callback to ConnPoolConfig

RaduBerinde:
This is a proposal - if you can think of bettter alternatives, I'd love to hear them.

-----
In CockroachDB test workloads, we want to have a pool of connections
against multiple database servers. With `sql.Database` and `lib/pq`,
we use our own driver which chooses one of multiple targets (in
round-robin fashion) on every `Open`.

In `pgx`, the same could be achieved using a custom `Dialer`, but this
would be pretty hacky:
 - we would need to encode the multiple server information into a
   single "address",
 - more importantly, it would break the cancellation logic which
   relies on creating a separate connection to the same server.

Instead, this commit adds a field `ConnPoolConfig.NewConnConfig` which
is a function that returns a `ConnConfig`. When this field is
specified, `ConnPoolConfig.ConnConfig` is ignored and this function is
used every time the pool creates a new connection. It's not great that
we have to keep both fields but it's necessary to maintain
compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/jackc/pgx/466)

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -16,13 +16,19 @@ type ConnPoolConfig struct {
 	MaxConnections int               // max simultaneous connections to use, default 5, must be at least 2
 	AfterConnect   func(*Conn) error // function to call on every new connection
 	AcquireTimeout time.Duration     // max wait time when all connections are busy (0 means no timeout)
+
+	// If set, the ConnConfig above is ignored and this function is used to obtain
+	// a ConnConfig every time a new connection is made. This allows using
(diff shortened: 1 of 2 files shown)


RaduBerinde:
CC @nvanbenschoten

RaduBerinde:
Retracting. We just wrote a thin layer on top of multiple `ConnPool`s to achieve the same thing.

PR #470, opened 2018-10-23
Replication: Add NOEXPORT_SNAPSHOT, GetConnInfo() function

blind-oracle:
* The NOEXPORT_SNAPSHOT option when creating a replication slot skips creating a snapshot on PostgreSQL side which is needed only when we're seeding the slave DB, which is not a use case of this library

* GetConnInfo() on ReplicationConn returns a connection info from the underlying connection which can be used in decoding the tuples that we get from DB


repl-samuel:
The NOEXPORT_SNAPSHOT option is only support for Postgresql 10+. for 9.x this option will cause CREATE_REPLICATION_SLOT failed with syntax error.

blind-oracle:
The 9.x PostgreSQL anyway does not support logical replication in a useful manner. I've spent quite some time trying to get pgx+pgoutput work with it, but failed. Then we've upgraded to 10.x.

But if somebody anyway using pgx with 9.x - then I agree that this will be a regression. Maybe we can wait for some bug reports and revert in this case.

jackc:
We already have a connection when running CreateReplicationSlot -- so presumably we know or can get the version of the server. Would it be worth branching on server version?

PR #473, opened 2018-10-26
Add path query parameter to support unix sockets

Mic92:
Currently there is no way to specify unix sockets in the connection
url. This patch adds a `path` query parameter that allows to set the path.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -782,6 +782,11 @@ func ParseURI(uri string) (ConnConfig, error) {
 			continue
 		}
 
+		if k == "host" {
+			cp.Host = v[0]
+			continue
+		}
+
 		cp.RuntimeParams[k] = v[0]
 	}
 	if cp.Password == "" {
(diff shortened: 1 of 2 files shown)


Mic92:
Ready to merge!

Mic92:
ping.

Mic92:
ping

jackc:
Sorry for the delay. pgx doesn't get much time at the day job at the present.

Is there some precedent or reason to prefer "path" instead of "host" as the keyword argument? In psql you would use host. e.g.

```
psql postgresql:///pgx_test?host=/var/run/postgresql
```

Mic92:
I found `path` more appealing but for consistency we should the probably also use `path=`.

jackc:
I think I'd prefer it to only be `host`. Rather not have two ways of doing the same thing. If that gets changed I'll merge it.

Mic92:
Yes I meant `host` actually this was a typo.

Mic92:
Should be ok now.

jackc:
Thanks

Issue #321, opened 2017-09-09
RedShift doesn't support ssl_renegotiation_limit


jacobmarble:
(Originally mentioned in #282)

pgx can't connect to a RedShift cluster when the cluster is configured to require SSL. The patch below illustrates a workaround, not a fix; there should be a more thoughtful way to approach this problem.

As RedShift inevitably continues to diverge from Postgres, it might make sense to add a new query parameter to the data source name parser. For example, `redshift=true`. This could be translated to a boolean field of the `ConnConfig` struct, made available for future RedShift-specific workarounds. If that sounds like a good idea, I'll be happy to write a PR.

RedShift have so far not committed to a fix, but at least someone there is aware of the problem.
https://forums.aws.amazon.com/thread.jspa?threadID=229990

```
$ git diff
diff --git a/conn.go b/conn.go
index f549e03..d0b4f4b 100644
--- a/conn.go
+++ b/conn.go
@@ -321,7 +321,7 @@ func (c *Conn) connect(config ConnConfig, network, address string, tlsConfig *tl
        // Go does not support (https://github.com/golang/go/issues/5742)
        // PostgreSQL recommends disabling (http://www.postgresql.org/docs/9.4/static/runtime-config-connection.html#GUC-SSL-RENEGOTIATION-LIMIT)
        if tlsConfig != nil {
-               startupMsg.Parameters["ssl_renegotiation_limit"] = "0"
+               //startupMsg.Parameters["ssl_renegotiation_limit"] = "0"
        }
```

jackc:
I'd prefer to avoid needing named modes for "almost PostgreSQL" servers. It's not the end of the world, but if possible I'd prefer to expose settings that can configure the connection to be compatible with various servers. See also #320 for more discussion of connecting to "almost PostgreSQL" servers.

In RedShift's case there are 2 issues. The first is the inability to connect due to the `ssl_renegotiation_limit` parameter. But with that removed, the first time RedShift sends a renegotiate request the connection will die. When I last looked into this that was at 512 MB of transit.

At the time Go did not support renegotiation at all. But now Go at least has partial opt-in support. I don't know if it is sufficient. If not, there's nothing we can do. But if it is, then there is a clean solution. The only reason we send `ssl_renegotiation_limit` is to avoid mysterious connection errors when connecting to servers that use renegotiation. But PostgreSQL defaults to 0 for all supported versions, and removed the setting for 9.6.

[...]

PR #476, opened 2018-10-30
tls: stop sending ssl_renegotiation_limit in startup message

tejasmanohar:
This addresses https://github.com/jackc/pgx/issues/321 with the
fix @jackc proposed there--

>The only reason we send ssl_renegotiation_limit is to avoid mysterious connection errors when connecting to servers that use renegotiation. But PostgreSQL defaults to 0 for all supported versions, and removed the setting for 9.6. So I guess all we need to do is remove the default startup parameter message. Anyone who really needs it can easily add it back with ConnConfig.RuntimeParams.

Redshift users that need to connect
w/ SSL currently fork the library to delete this parameter, e.g.

https://github.com/segmentio/pgx/commit/8e0028d742358e966669c28193fa65460a85ea69

And, that's annoying to keep up-to-date :)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -337,14 +337,6 @@ func (c *Conn) connect(config ConnConfig, network, address string, tlsConfig *tl
 		Parameters:      make(map[string]string),
 	}
 
-	// Default to disabling TLS renegotiation.
-	//
-	// Go does not support (https://github.com/golang/go/issues/5742)
-	// PostgreSQL recommends disabling (http://www.postgresql.org/docs/9.4/static/runtime-config-connection.html#GUC-SSL-RENEGOTIATION-LIMIT)
-	if tlsConfig != nil {
-		startupMsg.Parameters["ssl_renegotiation_limit"] = "0"
-	}
-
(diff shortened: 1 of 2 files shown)


tejasmanohar:
As expressed in `doc.go`, I _think_ this fix should go in `v3.3.0`. While I doubt this breaks connections for many, I suppose it could if a user of this library ignores warnings and has a non-zero value for this parameter set in `postgresql.conf` in Postgres < 9.6.

jackc:
Thanks. I agree it is reasonable to make this change in a minor release. It's really hard to imagine a case where it would cause breakage.

tejasmanohar:
@jackc Thanks for merging this. Would you please publish a `v3.3.0` tag? Or, are you waiting on something else to do that? The docs change merged in this PR mentions this being available as of `v3.30`, and I'd like to pin my `vendor` to a release of `pgx`.

jackc:
Okay, I released v3.3.0.

tejasmanohar:
Appreciate it

PR #477, opened 2018-11-02
Added option to customize query cancellation and wait for ready

fastest963:
CockroachDB doesn't support `80877102` and doesn't plan to, so instead allow
the user to customize the cancellation with their own function. In our
function, we call CANCEL QUERY with the query_id based on the LocalAddr().

The WaitForReady method can be used by a pool to not put a connection back
in the pool until it is finished cancelled and ready for a new query.

We've been using this in production and seems to be working well for us. We have our own pooling behavior that is using the new `WaitForReady` so that connections don't get put back in the pool that are actually going to be closed (because the cancel failed, for example). I did have to change the `cancelQueryInProgress` so that multiple things could listen on `cancelQueryCompleted` channel, which meant that I needed to close it not send a single thing over it.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
@@ -78,6 +77,7 @@ type ConnConfig struct {
 	RuntimeParams     map[string]string                     // Run-time parameters to set on connection as session default values (e.g. search_path or application_name)
 	OnNotice          NoticeHandler                         // Callback function called when a notice response is received.
 	CustomConnInfo    func(*Conn) (*pgtype.ConnInfo, error) // Callback function to implement connection strategies for different backends. crate, pgbouncer, pgpool, etc.
+	CustomCancel      func(*Conn) error                     // Callback function used to override cancellation behavior
 
 	// PreferSimpleProtocol disables implicit prepared statement usage. By default
 	// pgx automatically uses the unnamed prepared statement for Query and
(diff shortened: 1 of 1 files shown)


fastest963:
I'm not sure why that only single job failed. Can you retry it?

jackc:
I reran it. It looks like some behavior has changed when queries are canceled.

fastest963:
Looks like they failed in `TestStressConnPool` and given that different jobs failed the second time versus the first, I'm imaging it's just a flaky test. I'll try to see if I can improve it. I can't get that test to fail locally, however.

jackc:
That might be challenging to figure out. Because in a certain sense it's designed to be a flaky / timing sensitive test. It's purposely running a bunch of queries in parallel and cancelling some of them. So the exact state of the query being cancelled is non-deterministic.

fastest963:
Yeah, I'm going to add some debug logs to see if I can see what's happening when travis runs it.

fastest963:
I'm still debugging this. If I add print statements it seems to happen less often but the latest time:
```
cancelling query: 127.0.0.1:51566
finished cancelling query: 127.0.0.1:51566: <nil>
stress_test.go:93: canceledQueryExContext: should never receive row on 127.0.0.1:51566 after 2.118077
```

So it does seem like, despite the inner `doCancel` not erroring, it didn't cancel the query because after 2 seconds it still returned.

fastest963:
I think I've figured out the issue. I moved the resetting of the deadline into `cancelQuery` but this caused a race condition where it would attempt to cancel the query and the deadline would be reset before it applied to the query results. I put the resetting back into `waitUntilPreviousCancelQuery`, even though I originally moved it because it's kinda confusing that it's in there.

PR #479, opened 2018-11-09
remove PostgreSQL 10 exclusive argument

repl-samuel:
NOEXPORT_SNAPSHOT is added in PostgreSQL 10+
it causes CreateReplicationSlot() failed with syntax error in PostgreSQL 9.x

the fix partially reverts a recent commit https://github.com/jackc/pgx/commit/04bead7c8a8d6a647336fa74c7a31cd3d36336bf

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -441,7 +441,7 @@ func (rc *ReplicationConn) StartReplication(slotName string, startLsn uint64, ti
 
 // Create the replication slot, using the given name and output plugin.
 func (rc *ReplicationConn) CreateReplicationSlot(slotName, outputPlugin string) (err error) {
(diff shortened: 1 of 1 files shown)


repl-samuel:
if it's intended to break PostgreSQL 9.x support, please close this pull request.

repl-samuel:
closed the PR as no one is reviewing it.

jackc:
Sorry for lack of response, thought I had posted a reply. I guess it was actually on #470. The replication support in pgx was written by other contributors and I haven't personally used it in production. So especially on an issue like this where an improvement for PG10 causes problems for PG9 it is difficult for me to have an informed opinion.

tejasmanohar:
https://github.com/jackc/pgx/pull/470#issuecomment-438475791

I think it would be worth branching code as @jackc suggests here

PR #482, opened 2018-11-14
Added LastStmtSent and use it to retry on errors if statement was not sent

### LINKED ISSUE #427: Feature request: connection testing on checkout
Hi,

I was looking at ways to make sure that connections handed from the pool are not dead.

I'm aiming to support situations where the DB server might have been restarted, or the connection has been lost for whatever reason.

I have read the discussion in #74.  I see that transaction `Begin()` does retry the connection - and so far, wrapping all server-side DB interaction into a transaction works for us as an interim workaround.

Other DB pools I have worked with (in the Java world) support setting up a test query to be run each time a connection is checked out from the pool (or alternatively returned to the pool).  The test query might be just `SELECT 1;`, or it might be specific to the database driver.

For Apache Commons DBCP, the options configuring this behavior are called `validationQuery`  / `testOnBorrow` / `testOnReturn`: 
https://commons.apache.org/proper/commons-dbcp/configuration.html

For the C3P0 project, these are called `preferredTestQuery` / `testConnectionOnCheckout` / `testConnectionOnCheckin`:
https://www.mchange.com/projects/c3p0/#testConnectionOnCheckout

Would adding similar options to pgx sound as a suitable addition?

Cheers,
Vlad

fastest963:
Previously, a failed connection could be put back in a pool and when the
next query was attempted it would fail immediately trying to prepare the
query or reset the deadline. It wasn't clear if the Query or Exec call
could safely be retried since there was no way to know where it failed.

You can now call LastQuerySent and if it returns false then you're
guaranteed that the last call to Query(Ex)/Exec(Ex) didn't get far enough
to attempt to send the query. The call can be retried with a new
connection.

This is used in the stdlib to return a ErrBadConn if a network error
occurred and the statement was not attempted.

Fixes #427

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -135,6 +135,7 @@ type Conn struct {
 
 	pendingReadyForQueryCount int // number of ReadyForQuery messages expected
 	cancelQueryCompleted      chan struct{}
+	lastStmtSent              bool
 
 	// context support
 	ctxInProgress bool
@@ -1731,6 +1732,7 @@ func (c *Conn) Ping(ctx context.Context) error {
 }
 
 func (c *Conn) ExecEx(ctx context.Context, sql string, options *QueryExOptions, arguments ...interface{}) (CommandTag, error) {
+	c.lastStmtSent = false
 	err := c.waitForPreviousCancelQuery(ctx)
 	if err != nil {
 		return "", err
@@ -1770,6 +1772,7 @@ func (c *Conn) execEx(ctx context.Context, sql string, options *QueryExOptions, 
 	}()
 
 	if (options == nil && c.config.PreferSimpleProtocol) || (options != nil && options.SimpleProtocol) {
+		c.lastStmtSent = true
 		err = c.sanitizeAndSendSimpleQuery(sql, arguments...)
 		if err != nil {
 			return "", err
@@ -1786,6 +1789,7 @@ func (c *Conn) execEx(ctx context.Context, sql string, options *QueryExOptions, 
 
 		buf = appendSync(buf)
 
+		c.lastStmtSent = true
 		n, err := c.conn.Write(buf)
 		if err != nil && fatalWriteErr(n, err) {
 			c.die(err)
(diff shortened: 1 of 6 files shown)


fastest963:
@jackc let me know what you think of this. It's a lot cleaner than both of my last attempts. I ended up not liking the `RetryableConn` because I constantly had to switch between it and a normal Conn.

jackc:
This is definitely better.

I noticed two things as I read the code:

Is the mutex around query sent indicator required? I didn't see anywhere it would be read or written concurrently (I may have missed it of course).

Would it be reasonable to encode this data into an error? It would be better if the connection didn't store that state and it was part of the return value from query functions. The previous attempt at this did something like that internally and altered the return signature. But what if the signature could remain the same? Something like the [net.Error](https://golang.org/pkg/net/#Error) interface exposes for timeout and temporary errors.  A helper function would probably be necessary to handle unwrapping errors (https://godoc.org/github.com/pkg/errors style)

We could end up with something like:

```go
_, err := conn.Query(...)
if err != nil {
  if pgx.QueryNeverSent(err) { // not sure I like this name -- and maybe even it belongs in a separate package
    // whatever retry logic
  }
}   
```

This may require too much change to how pgx handles errors... but it would be interesting.

fastest963:
Yeah, I actually originally did it that way that but then it wouldn't allow `== sql.ErrNoRows` or more importantly `== context.Canceled`. I wasn't sure how many people checked or cast the response to get the error code and how that might break or change. I think it was a poor decision to encourage on `== someError` but alas...

To be clear, the erro is definitely better I was just less comfortable with it and it being backwards compatible when I first tried it.

> Is the mutex around query sent indicator required?

Theoretically you could call LastStmtSent concurrently to it being set by one of the query methods. I aired on the side of caution but I can remove it.

jackc:
Ah, good point on the errors. Would be very difficult to do that in a backwards compatible fashion.

I don't think the mutex is necessary. pgx.Conn is documented as not being concurrency safe.

jackc:
Thanks!

PR #483, opened 2018-11-25
all: use environment variable PGX_TEST_DATABASE for testing

shuLhan:
Instead of instructing developer/tester to edit and copy file manually,
we can read the connection string for testing from defined OS environment
variable PGX_TEST_DATABASE.

This mode of testing has been used before on package "pgtype".

While at it, fix nil pointer dereference on TestPoolNonBlockingConnections.


shuLhan:
Sorry, I think I miss the notification for above comment and I am not sure why there are other commits above my patch.

Closing this patch as abandoned.

Issue #339, opened 2017-10-09
Support nil pointers to value implementing driver.Valuer


kdar:
For example, you can't do something like this:

```
var d *apd.Decimal
fmt.Println(pool.Exec(`INSERT INTO test (value) VALUES ($1)`, d))
```

Relevant discussion with the Go stdlib is here: https://github.com/golang/go/issues/8415
Commit fixing this in Go stdlib is here: https://github.com/golang/go/commit/0ce1d79a6a771f7449ec493b993ed2a720917870

kdar:
I'm also not sure if pgx checks properly if a type implements driver.Valuer with a pointer receiver. When I try, I get an error like this:

```
cannot convert {0xc4201e2120} to Numeric
```

Instead of calling Value() on the pointer receiver, it seems to try to coerce the pointer into Numeric.

jackc:
I used the same fix as was done for the standard lib. It now incurs a reflect operation for each driver.Valuer argument, but there doesn't seem to be any way around it.

kdar:
Do you have any insight into the second comment I made on this issue?

jackc:
I don't think you should need a wrapper. I have a working test in the test suite with a `nil` `*apd.Decimal` so you should be able to do the same that it does.

PR #484, opened 2018-11-25
query_test: define the type instead of importing other package

shuLhan:
The test for issue #339 actually can work for any custom type that
implement driver.Valuer, we did not need to import another package just
for testing it, which minimize the package dependencies.

While at it, remove invalid instruction to "go get" the package
go-version in README.


jackc:
It's true that strictly speaking it's not required... But the dependency is only in test mode and it gives me more confidence in the regression test to use the exact type that was given in the bug report.

redbaron:
+1 for this change.  Luckily dependency is small, but it is not required nevertheless. Otherwise lets close it to keep list tidy

PR #487, opened 2018-11-28
Add helpers to obtain connection pool with config parsed from env

nuqz:
I use these functions in several of my projects. I would not want to copy and paste them into my next project, but I don’t think it’s worthwhile for them to create a separate package, so I suggest including them in the main repository.


jackc:
I see how it would be more convenient more that particular way of establishing a connection pool. But I think it is a bit too specific for general use.

PR #488, opened 2018-11-30
Apply UUID string length check before parsing

maxarchx:
Simple check for UUID is not an empty string and looks correct.
Solves "slice bounds out of range" runtime error in some situations.


jackc:
Thanks

PR #491, opened 2018-12-11
More precise .pgpass handling

mialinx:
* localhost matches unix sockets too
* use OS user as default username
* use username as default database
* ignore comment lines


jackc:
Thanks!

PR #493, opened 2018-12-12
Adds RowsAffected for CopyToWriter and CopyFromReader

nvorobev:
RowsAffected returns the number of rows affected by COPY
https://www.postgresql.org/docs/9.6/sql-copy.html

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -553,10 +553,10 @@ func (p *ConnPool) CopyFrom(tableName Identifier, columnNames []string, rowSrc C
 }
 
 // CopyFromReader acquires a connection, delegates the call to that connection, and releases the connection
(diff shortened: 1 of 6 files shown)


nvorobev:
RowsAffected returns the number of rows affected by COPY

kirugan:
@jackc hi! Can you please review this PR. We use pgx at work, and unfortunately for copy we have to use go-pg library instead. This changes allow us to remove usage of go-pg in our project.

jackc:
I like the change. The problem is it technically is a breaking change. Any current use of CopyToWriter and CopyFromReader that checks the error response would no longer compile.

I try pretty hard to follow semver / avoid breaking changes... But on the other hand it is also unreasonable for a straight forward improvement to have to wait months or years for a major release.

Hmm...

Well, I guess I'll chance breaking semver this time and see how much push back there is.

PR #500, opened 2019-01-15
support binding of []int type to array integer

dgsb:
binding go type `int` to database type `integer` is handled, so it would also make sense to be able to bind `[]int` to array `integer`.


dgsb:
@jackc will you have some time to review this soon ?

jackc:
Thanks!

dgsb:
@jackc any plan to tag a new release soon whic would embed this fix ?

PR #501, opened 2019-01-18
Fix encoding of ErrorResponse

coder543:
Currently, ErrorResponse does not encode correctly because it is not writing the type of each field before the field.


jackc:
Thanks!

Issue #481, opened 2018-11-14
Access to `pgxDriver` in `stdlib` package to use datadog tracing 


platelk:
Hi !

On our project we came with a little issue with the use of `gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql` with `pgx`, this package require to "wrap" a `sql.Driver` like this

```go
import sqltrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql"

// ...

sqltrace.Register(
		"pgutils-trace",
		driver,
		sqltrace.WithServiceName(conf.Database),
	)

// ...
```
But as far as I look in the `stdlib` package of `pgx` I can't see a way of accessing to `pgxDriver` or to instantiate "nicely" a new Driver without trying to instantiate private struct member.

Is there is a way of doing this ? or should I add a new function to retrieve this driver ?

jackc:
There isn't a way to do this now. Changing pgx this var in some way is one solution. But as of Go 1.10 I believe there is a better solution available. Go 1.10 added the [OpenDB](https://golang.org/pkg/database/sql/#OpenDB) function. This allows constructing `*sql.DB` directly without using `sql.Register` and `sql.Open`.

I think the DataDog extension could be implemented in terms of OpenDB and not require the underlying driver object to be exposed.

jackc:
I haven't used the datadog extension, but looking at it a little closer it looks like it just needs a database/sql/driver.Driver. Can you just create a new one?

```
driver := stdlib.Driver{}
```

A don't think it needs any initialization in this case.

ahajimir:
@platelk Were experiencing something similar with our projects. What approach did you end up taking? And how did it work out?

Cheers

rakshans1:
@platelk How to integrate with ddtrace? can you provide a code sample

PR #511, opened 2019-02-05
Close issue #481 : Give access to the registered driver instance

platelk:
Some library use a driver to wrap its behavior and give additional
functionality, as the datadog tracing library
("gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql")

This commit aims to give access to this instance which can't be
correctly initialized to due private fields without default values (the
configuration map inside the driver)

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -126,6 +126,12 @@ var (
 	fakeTxConns map[*pgx.Conn]*sql.Tx
 )
 
+// GetDefaultDriver return the driver initialize in the init function
+// and used when register pgx driver
+func GetDefaultDriver() *Driver {
+	return pgxDriver
+}
+
 type Driver struct {
 	configMutex sync.Mutex
 	configCount int64


stdlib/sql.go
```diff
@@ -126,6 +126,12 @@ var (
 	fakeTxConns map[*pgx.Conn]*sql.Tx
 )
 
+// GetDefaultDriver return the driver initialize in the init function
+// and used when register pgx driver
+func GetDefaultDriver() *Driver {
```
lbcjbb: Return **sql.Driver** rather than ***Driver**

jackc: Hmm. Not sure about this. Are there any specific reasons one way or another aside from general rules of thumb regarding when to return an interface or not?

platelk: I followed the default way recommanded in go : Take interface, return structs, I thinks its easier to access specific plublic method

PR #515, opened 2019-03-07
Allow normal queries on replication connections

fcelda:
This change exposes `pgx.Conn` methods on the `pgx.ReplicationConn`.

The replication connection allows executing most of the SQL queries which are available on non-replication connections. There are some limitations to that (e.g. prepared statements are not available) but enabling `PreferSimpleProtocol` in the connection configuration solved all the issues I encountered.

Being able to execute normal queries on the replication connection is very useful for building full logical replication client. For instance, it is useful to query system catalogs to learn about the schema and also to perform initial snapshot of tables.

Particularly, PostgreSQL server in a role of the logical replication client creates a temporary logical replication slot within a transaction and then gets the data using COPY. This sequence of steps is possible only on a replicated connection and allows the client to get the LSN corresponding to the snapshot.

I believe this change should not break API but let me know if you would prefer a different solution.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -201,18 +201,6 @@ func (rc *ReplicationConn) SendStandbyStatus(k *StandbyStatus) (err error) {
 	return
 }
 
-func (rc *ReplicationConn) Close() error {
-	return rc.Close()
-}
-
-func (rc *ReplicationConn) IsAlive() bool {
-	return rc.IsAlive()
-}
-
-func (rc *ReplicationConn) CauseOfDeath() error {
-	return rc.CauseOfDeath()
-}
-
 func (rc *ReplicationConn) GetConnInfo() *pgtype.ConnInfo {
 	return rc.ConnInfo
 }
--- a/replication.go
+++ b/replication.go
@@ -168,18 +168,18 @@ func ReplicationConnect(config ConnConfig) (r *ReplicationConn, err error) {
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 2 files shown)


jackc:
I agree that the use case you present is valuable and that it doesn't seem to break the API. But I'm not sure how I feel about exposing the entire Conn interface when a substantial chunk of it may not be useful or may not work on a replication connection. Though as you say, logical replication has changed things a bit in the last few years.

Maybe if we added some documentation to ReplicationConn explaining that while all of Conn's interface is exposed there are significant limitations to what can actually be used. I expect anyone building a replication client would need to have a high level of PG expertise, so a simple "be careful" might be enough.

fcelda:
Sure. I can improve the documentation. Please, give me a few days.

fcelda:
@jackc I've added documentation to the ReplicationConn struct. Let me know what you think.

I also noticed that the [documentation](https://www.postgresql.org/docs/11/protocol-replication.html) explicitly states that replication mode requires the simple protocol mode. I originally found that by trial and error. I wonder if it makes sense to enforce PreferSimpleProtocol on the config in ReplicationConnect().

jackc:
Additional docs look fine. I agree that enforcing PreferSimpleProtocol would be a good idea as well.

fcelda:
@jackc Updated. I've also added a unit test to check that PreferSimpleProtocol is set on new replication connections. I think this is ready to be merged.

jackc:
Thanks!

PR #518, opened 2019-03-11
Fix enum handling

### LINKED ISSUE #335: possible bug with enum handling
I read #287, could it be it broke something with enum handling?

I get a panic when trying to insert a custom type (with `string` being the underlying type):

`*pgtype.GenericText is not pgtype.BinaryEncoder: missing method EncodeBinary`

The panic occurs on line `values.go:166` (function `encodePreparedStatementArgument`):

`argBuf, err := value.(pgtype.BinaryEncoder).EncodeBinary(ci, buf)`

Any idea?

bobheadxi:
closes #335 , and implements the fix as described by @lbcjbb

please let me know if there's a better way I'm supposed to implement the test! (9a3e403, which I've removed for the time being in 0a8645d since it doesn't work)

### DIFF
--- a/values.go
+++ b/values.go
@@ -189,7 +189,15 @@ func encodePreparedStatementArgument(ci *pgtype.ConnInfo, buf []byte, oid pgtype
 
 		sp := len(buf)
 		buf = pgio.AppendInt32(buf, -1)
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #521, opened 2019-03-18
Remove unreachable code

lukedirtwalker:
The returns can never be reached because the loop is guaranteed to return.


jackc:
Thanks!

PR #523, opened 2019-03-28
[stdlib] Add support for creating a DB from pgx.Pool

fzerorubigd:
Also, the configuration used in the Conn structure (used to implement the driver.Conn interface) stores a ConnConfig which is used only for determining if the Connection should be used with Simple Protocol or not.

With this, I can use the pgx as default and also use the same pool for converting the connection to sql.DB

### DIFF
--- a/stdlib/opendbpool.go
+++ b/stdlib/opendbpool.go
@@ -1,0 +1,60 @@
+// +build go1.10
+
+package stdlib
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+
+	"github.com/jackc/pgx"
+)
+
+// OptionOpenDB options for configuring the driver when opening a new db pool.
+type OptionOpenDBFromPool func(*poolConnector)
+
+// OptionAfterConnect provide a callback for after connect.
+func OptionPreferSimpleProtocol(preferSimpleProtocol bool) OptionOpenDBFromPool {
+	return func(dc *poolConnector) {
+		dc.preferSimpleProtocol = preferSimpleProtocol
+	}
+}
+
+// OpenDBFromPool create a sql.DB connection from a pgx.ConnPool
+func OpenDBFromPool(pool *pgx.ConnPool, opts ...OptionOpenDBFromPool) *sql.DB {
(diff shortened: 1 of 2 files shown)


jackc:
Can you explain a bit more what you are trying to do? I'm not sure I understand the use-case.

fzerorubigd:
We are trying to migrate to pgx from pq (mostly for batching) but there is
several library (including migration tool we are using and several other
maintained-by-other libraries) needs the sql.DB connection. we can use two
separate connection here, (using the OpenDB) but using the same connection
in the entire application makes more sense.

fzerorubigd:
Also to complete my other comment, the problem is how we can control the number of connections for a single app using both pgx and sql.DB. 
there is no clear way to limit the connections on both at the same time. but in this way, we can control the maximum with the pgx.Pool.

jackc:
Thanks!

seh:
What happened to this _stdlib/opendbpool.go_ file? I'm unable to find it on either the "master" or "v4" branches today.

jackc:
It was merged into `v3` (which was `master` at the time) after `v4` branched off and was never ported to `v4`. The connection pool in `v3` is significantly different than `v4` so I am unsure what would be involved in porting this functionality.

seh:
Thank you for clarifying.

Along the "v4" branch, is there any way to use the connection pool capability in concert with the standard library wrapper? I studied the source for a while last night, but couldn't see anything in there that looked like it would accommodate use of the `pgxpool` package.

jackc:
At the moment there is not. If you need to use the same pool for `database/sql` and `pgx` you would need to use the `database/sql` pool and use `stdlib.AcquireConn` to get a `pgx` connection.

seh:
If I just use the `pgxstd.OpenDB` function, and take the returned `*sql.DB` object and use its various execution and query functions, is there any connection pooling going on under the covers? The documentation is vague, but it suggests that using the `DB` object provides pooling.

jackc:
Yes. It uses the standard `database/sql` pool.

Issue #310, opened 2017-08-28
Some types does not support MarshalJSON


jeromer:
I noticed that types like `timestamptz` does not implement `MarshalJSON`. Is it on purpose ? Why ?

Thanks :)

jackc:
The types that have `MarshalJSON` have been added on a as-needed basis. The main reason other types don't have it is no one has needed it enough to do it.

But for `timestamptz` and a few other types there is an additional issue of the lack of a canonical representation. It could be ISO 8601, milliseconds since Unix epoch, or one of several other "standards".

jeromer:
Hi,

On Tue, Aug 29, 2017, at 09:47 PM, Jack Christensen wrote:
> The types that have MarshalJSON have been added on a as-needed basis.
> The main reason other types don't have it is no one has needed it
> enough to do it.
Ok

> But for timestamptz and a few other types there is an additional issue
> of the lack of a canonical representation. It could be ISO 8601,
> milliseconds since Unix epoch, or one of several other "standards".
Makes sense.

Thanks for replying :)

--
Jérôme Renard

azamat-sharapov:
@jackc how can I implement MarshalJSON for `TimestampArray` type? I've tried following, but it doesn't work:

```
type TimestampArray struct {
  pgtype.TimestampArray
}

func (src *TimestampArray) MarshalJSON() ([]byte, error) {
  log.Println("this never called")
  return []byte(), nil
}
```

jackc:
At first glance it looks fine to me. But whatever the issue is, I don't think it has to do with `pgtype.TimestampArray` -- given that your `MarshalJSON` is never called. Perhaps start with an empty struct and get `MarshalJSON` to be called, and then add `pgtype.TimestampArray` back in.

azamat-sharapov:
@jackc Thanks. It worked after I added `UnmarshalJSON`, because it needs to implement json marshaller interface.

PR #528, opened 2019-04-21
Support for pgtype.Date JSON marshal/unmarshal.

kak-tus:
JSON marshalling for types added on a as-needed basis.
Partly closes https://github.com/jackc/pgx/issues/310.

### DIFF
--- a/pgtype/date.go
+++ b/pgtype/date.go
@@ -3,6 +3,7 @@ package pgtype
 import (
 	"database/sql/driver"
 	"encoding/binary"
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


jackc:
Is there a reason for including the time portion instead of only "yyyy-mm-dd"? My initial thought is that it doesn't make sense to include the time portion.

kak-tus:
Hm, you are right. I will fix this.

kak-tus:
Done.

jackc:
Thanks!

PR #531, opened 2019-04-30
Adding support for pgpass file

emdem:
As per the libpq-pgpass documentation linked below:
https://www.postgresql.org/docs/9.4/libpq-pgpass.html

Support for a file format containing a single line:
hostname:port:database:username:password

Edit: I realize now there is already some functionality to parse pgpass, I shall utilize that but still add the environment variable override for the pgpass location.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -885,6 +885,7 @@ func ParseConnectionString(s string) (ConnConfig, error) {
 // PGDATABASE
 // PGUSER
 // PGPASSWORD
+// PGPASSFILE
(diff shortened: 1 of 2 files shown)


emdem:
There are a few bugs that need to be fixed... I will make another PR when I've sorted them out.

jackc:
AFAIK, the `PGPASSFILE` environment variable should already work.

https://github.com/jackc/pgx/blob/master/pgpass.go#L61

PR #533, opened 2019-05-02
Add func CheckReplicationSlotExists and its Test case

davehowell:
Before creating a replication slot or attempting to drop one, and in some other cases, it is helpful to just check if the replication slot with a given name exists. 

For example, attempting to run the `DropReplicationSlot` method means you have to watch for obscure Postgres error like `SQLSTATE 42704` or similar. 

CheckReplicationSlotExists is a simple query to see if the slot with that name already exists so you can avoid attempting to drop the non-existent slot.

[Valid slot names ](https://www.postgresql.org/docs/9.5/warm-standby.html#STREAMING-REPLICATION-SLOTS-MANIPULATION) " can contain lower-case letters, numbers, and the underscore character ", so the dummy slot name I've used is "0" so it sorts last (desc).

Includes an appropriate test case.

### DIFF
--- a/replication.go
+++ b/replication.go
@@ -462,3 +462,23 @@ func (rc *ReplicationConn) DropReplicationSlot(slotName string) (err error) {
 	_, err = rc.Exec(fmt.Sprintf("DROP_REPLICATION_SLOT %s", slotName))
 	return
 }
+
+//Check the replication slot exists for a given name
+func (rc *ReplicationConn) CheckReplicationSlotExists(slotName string) (slotExists bool, err error) {
+	var rows *Rows
+	rows, err = rc.sendReplicationModeQuery(fmt.Sprintf(`
+	SELECT
+		CASE WHEN slot_name = '0' THEN false else true end
+	FROM (
+		SELECT slot_name from pg_replication_slots
+		WHERE slot_name = '%s'
+		UNION ALL
+		SELECT '0' as slot_name
+		) AS mustreturn
+	ORDER BY slot_name DESC LIMIT 1`, slotName))
+	defer rows.Close()
+	for rows.Next() {
+		rows.Scan(&slotExists)
+	}
+	return
+}
--- a/replication_test.go
+++ b/replication_test.go
@@ -368,3 +368,45 @@ func TestSimpleProtocolEnforcement(t *testing.T) {
 	}
 	rows.Close()
 }
+
+func TestCheckReplicationSlotExists(t *testing.T) {
+	if replicationConnConfig == nil {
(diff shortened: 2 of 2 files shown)


jackc:
I don't understand the `slot_name = '0'` thing.

Could the query be something like:

```sql
SELECT 1 from pg_replication_slots WHERE slot_name = '%s'
```

Then use QueryRow and check for `ErrNoRows`.

davehowell:
The SQL is a bit convoluted in retrospect. The intention was to make the sub-query always return a single row that would then make the outer query always return a boolean. 

The '0' is a dummy row so if the slot name doesn't exist, the '0' row will sort first. If the slot name does exist, it will sort first. The query is sorted DESC and limited to 1 row. 
'0' was a poor choice, a blank string would have been more obvious and a better choice given it is the same string type as the slot_name, is not a valid slot name, without having to differentiate any errors. 

Something like this might have been more obvious:
```go
fmt.Sprintf(`
	SELECT
		CASE WHEN slot_name = '%s' THEN true else false end
	FROM (
		SELECT slot_name from pg_replication_slots
		WHERE slot_name = '%s'
		UNION ALL
		SELECT '' as slot_name
		) AS mustreturn
	ORDER BY slot_name DESC LIMIT 1`, slotName)
```

Your suggestion of using ErrNoRows is clearer, I'll do that.

davehowell:
I tried using `QueryRow` but I am using a ReplicationConn so makes it hard to `Scan` into a `bool`, it gives this error: `Query failed: unknown oid: 16, name: bool` 

Using `sendReplicationModeQuery` returns multiple rows so have to loop the scan. 

I guess I could use a non-replication connection, but I feel this function belongs with other replication things; it would be confusing to have to use a second non-replication connection in client code. 

This is the closest I can get to implementing your suggestion, let me know what you'd prefer:

```
//Check the replication slot exists for a given name
func (rc *ReplicationConn) CheckReplicationSlotExists(slotName string) (slotExists bool, err error) {
	var rows *Rows
	rows, err = rc.sendReplicationModeQuery(fmt.Sprintf(`
		SELECT true FROM pg_replication_slots WHERE slot_name = '%s'`, slotName))
	defer rows.Close()
	for rows.Next() {
		rows.Scan(&slotExists)
		break
	}
	switch err {
	case sql.ErrNoRows:
		return false, nil
	case nil:
		return true, nil
	default:
		return false, err

	}
}
```

jackc:
I believe a recent commit made all connection methods available on replication connections. You need to ensure that it uses the "simple protocol", but then you should be able to use QueryRow directly.

davehowell:
this is redundant now

PR #536, opened 2019-05-20
Hstore can have empty keys

coder543:
Postgres actually supports empty keys in hstores. You can test this with a simple:

```sql
INSERT INTO table (hstorecolumn) VALUES ('""=>3');
INSERT INTO table (hstorecolumn) VALUES ('"a"=>4');
```

One record will have an empty key with a value of 3, one will have a key of `a` with a value of 4.


jackc:
Huh, good catch. Thanks!

PR #540, opened 2019-06-03
added PortalSuspended message

### LINKED ISSUE #539: add PortalSuspended BackendMessage to pgproto3
`PortalSuspended` message is missing from the protocol.

need to create the struct that implements `BackendMessage` and add it to the switch case in `(*Frontend).Receive`

reference: https://www.postgresql.org/docs/10/protocol-message-formats.html

avivklas:
## Changes

- created a struct `PortalSuspended` that implements `BackendMessage`
- added `PortalSuspended` to `Frontend` as a valid backend message in `Receive()`

closes #539


jackc:
Thanks!

PR #542, opened 2019-06-06
Add PortalSuspended message

mjibson:
This wasn't present because pgx doesn't use the MaxRows field of Execute. But
if it did, this message would be sent from Postgres.


mjibson:
Oops I had no idea #540 was already opened. Closing this one.

PR #544, opened 2019-06-07
pgtype: Fix -0 for numeric types

### LINKED ISSUE #543: -0 is not supported by pgtype.Numeric
In `numeric.go` in the type switch, the case for float64 is as follows:
```
	case float64:
		num, exp, err := parseNumericString(strconv.FormatFloat(value, 'f', -1, 64))
		if err != nil {
			return err
		}
		*dst = Numeric{Int: num, Exp: exp, Status: Present}
```

Formatting a float can result in, the valid float value, of `-0`. It seems that the `parseNumericString` function cannot handle this value.

I'm not too well versed on the significance of `-0`, so correct me if I'm wrong, but my first assumption would be to drop the sign and continue as usual.

heedson:
Due to the special case of when the digits string was longer than 1 but
only contained the negative sign and a 0, it was incorrectly stripping
the 0 and attempting to parse "-" as a number.

The solution is to check an extra position along to make sure a trailing
0 is not immediately preceeded by a negetive sign.

Fixes #543

### DIFF
--- a/pgtype/numeric.go
+++ b/pgtype/numeric.go
@@ -321,7 +321,7 @@ func parseNumericString(str string) (n *big.Int, exp int32, err error) {
 	if len(parts) > 1 {
 		exp = int32(-len(parts[1]))
 	} else {
-		for len(digits) > 1 && digits[len(digits)-1] == '0' {
+		for len(digits) > 1 && digits[len(digits)-1] == '0' && digits[len(digits)-2] != '-' {
 			digits = digits[:len(digits)-1]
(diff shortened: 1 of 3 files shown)


heedson:
The `math.Copysign(0, -1)` feels somewhat bad for the tests but as described in this [Go issue](https://github.com/golang/go/issues/2196), it seems like a simple `float64(-0)` doesn't cut it.

jackc:
Thanks!

Issue #299, opened 2017-08-09
Usage with RDS or other HA Solutions


MaerF0x0:
When one uses RDS they connect to a given fqdn(host) and port . In the event of a failover scenario, amazon changes the IP that the fqdn points to to a hot spare pg box. 

I tested the failover and it seems that pgx did not detect anything occurred. Is it possible to have pgx [testOnAcquire](http://godoc.org/github.com/garyburd/redigo/redis#Pool.TestOnBorrow)  or similar? Basically I want some kind of hook where I can have the pool detect it's no longer working and make it connect to the new IP address. 

Thoughts?

MaerF0x0:
What if we test the connections before saying they're available here: https://github.com/jackc/pgx/blob/master/conn_pool.go#L280  ?

jackc:
I'd like to support or at least provide hooks to support this use case. But what would the hook do? If Amazon has changed the IP the FQDN resolves to shouldn't creating a connection normally connect to the new server? Calling `ConnPool.Reset` in this case should cause all existing connections to be invalidated. All uses of the pool from that point should be on new connections to the new server.

Regardless of that, the `TestOnBorrow` function in the redis adapter looks like a good idea. I'm :+1: for that.

jackc:
As far as `Stat` testing all connections, I'm not sure I like the performance implications of that. It changes a predictably fast call into an call with an unknown time cost.

riaan53:
Im currently also looking into failover scenarios. I see that the jdbc driver supports multiple hosts in the connection url: https://jdbc.postgresql.org/documentation/head/connect.html . Can also extend that for load balancing.

bpicolo:
@jackc  libpq supports these scenarios now through `target_session_attrs` in v10:

https://paquier.xyz/postgresql-2/postgres-10-libpq-read-write/

It essentially just does a round-robin check for whether the connection is read-only. Is that something you think could get into the roadmap? Would be great to have built-in HA here.

jackc:
I agree it would be good to have compatibility with libpq on this feature -- a PR would be welcomed.

jackc:
I believe this issue is resolved now that pgx has support for `target_session_attrs`.

gfodor:
I don't think target_session_attrs fixes this with RDS domain name updates. RDS does not expose multiple hostnames, they rely upon the client re-resolving the DNS record when there is a connection failure.

PR #545, opened 2019-06-18
Multihost with Target Session Attrs

furdarius:
This PR supports Multiple Hosts in connection string and Target Session Attrs parameter to define writable properties that are expected from the backend server. Changes are fully backward compatible.

https://paquier.xyz/postgresql-2/postgres-10-libpq-read-write/
https://www.postgresql.org/docs/devel/libpq-connect.html#LIBPQ-CONNSTRING

To avoid backward-incompatible changes I've extend value of `Host` parameter of `ConnConfig`. Now it's possible to define `Host` as `2.2.2.2:1,127.0.0.1,4.2.4.2` and this string will be interpreted as `[2.2.2.2:1, 127.0.0.1:5432, 4.2.4.2:5432]` array. In this case, `Port` parameter controls what port will be used if not specified in `Host` string. 

During connection process we loop over array of addrs and trying to connect one by one. If connection is successfully established then we try to check if connection is acceptable with `TargetSessionAttrs` (using `SHOW transaction_read_only` request).

On any auth error return immediately. All errors are accumulated and concatenated at the return. To keep backwards compatibility, if only one error occurred, then return it unchanged.

Example of config, that can be used to specify multiple hosts to connect to, and `TargetSessionAttrs` to control what type of connection we want.
```go
pgx.ConnConfig{
	Host:     "2.2.2.2:1,127.0.0.1,4.2.4.2",
	User:     "pgx",
	Password: "secret",
	Database: "mydb",
	Dial: (&net.Dialer{
		KeepAlive: 5 * time.Minute,
		Timeout: 100 * time.Millisecond,
	}).Dial,
	TargetSessionAttrs: "read-write"
}
```

**UPD (11 July 2019)**
`ParseConnectionString` supports Multi-Hosts and target_session_attrs now.
Examples:
```
# Example URL
postgres://jack:secret@foo.example.com:5432,bar.example.com:5432/mydb?target_session_attrs=read-write

# Example DSN
user=jack password=secret host=host1,host2,host3 port=5432,5433,5434 dbname=mydb sslmode=verify-ca target_session_attrs=read-write
```

`PGTARGETSESSIONATTRS` env variable supported.

TODO:
- [x] ParseConnectionString must support Multi-Hosts
- [x] PGTARGETSESSIONATTRS env.
- [x] ~Allow password to be different for each host/port pair if a password file is used~ (_This PR will not cover it_)
- [x] Think about case when DNS returns many IPs - how to provide writable check for each of endpoint? (_This PR will not cover it._)

related to #299


conn.go
```diff
@@ -262,33 +357,125 @@ func connect(config ConnConfig, connInfo *pgtype.ConnInfo) (c *Conn, err error)
...
+
+	var errmsg string
+	for _, err := range errs {
+		errmsg += "; " + err.Error()
+	}
+
+	return nil, errors.New(errmsg)
+}
+
+func (c *Conn) writable() error {
```
x4m: m.b. rename to checkWritableIfReuired()?
or just checkSessionAtrs()?

furdarius: `checkWritableIfReuired` - looks little bit java-style :) `checkWritable` maybe?

conn.go
```diff
@@ -89,22 +108,94 @@ type ConnConfig struct {
...
+			}
+
+			addr.Host = hostport[:pos]
+			addr.Port = uint16(p)
+		}
+
+		addrs = append(addrs, &addr)
+	}
+
+	return addrs, nil
```
furdarius: This function returns list of addrs parsed from `cc.Host` string, but what if DNS resolves one host to many IP's? Currently, we delegate to DialFunc resolving and connecting to any of resolved IP's. What if we want to store topology of our postgres instances in DNS, in this case we can't guarantee that master can be correctly resolved.

To support this, we must explicitly resolve hostnames to IP here and returns list of available IP's.

If you have thoughts on this topic - let's discuss.

conn.go
x4m: I believe we do not check writability here but check compliance with target session attributes. I think we could also restructure code a bit to make the future implementation of prefer-standby simpler.

conn.go
x4m: As far as I understand (and my understanding is very limited), DNS semantics implies that you do not have to go to ALL ip addresses, but may contact ANY. Thus, it is perfectly reasonable to leave this part to Dialer.

conn.go
furdarius: https://www.postgresql.org/docs/devel/libpq-connect.html#libpq-connstring

> When multiple hosts are specified, or when a single host name is translated to multiple addresses, all the hosts and addresses will be tried in order, until one succeeds. If none of the hosts can be reached, the connection fails. If a connection is established successfully, but authentication fails, the remaining hosts in the list are not tried.

conn.go
furdarius: https://www.postgresql.org/docs/devel/libpq-connect.html#libpq-connstring

> When multiple hosts are specified, or when a single host name is translated to multiple addresses, all the hosts and addresses will be tried in order, until one succeeds. If none of the hosts can be reached, the connection fails. If a connection is established successfully, but authentication fails, the remaining hosts in the list are not tried.

jackc:
Can you check out the v4-dev branch of pgx and see if it does what you want? It's not quite ready for production use yet, but it does have libpq style multi-host support (through the new underlying https://github.com/jackc/pgconn layer).

I'd still be willing to merge in an implementation for the current pgx v3, but depending on your time frame you may want to look at v4 first.

furdarius:
@jackc Thank you for your comment and for work on pgx driver)

> Can you check out the v4-dev branch of pgx and see if it does what you want?

I read the code of v4 pgx, and already opened [PR](https://github.com/jackc/pgconn/pull/1) with some small fixes and [issue](https://github.com/jackc/pgx/issues/546) about additional OIDs :)

What about Multi-Host? Our approach is very similar, but I have some question about implementation.

As I see, you decided to use [AfterConnectFunc](https://github.com/jackc/pgconn/blob/master/config.go#L241) to check writability. But what if both `settings["target_session_attrs"]` and `config.AfterConnectFunc` set by user? In this case user defined `config.AfterConnectFunc` will never be executed (only `AfterConnectTargetSessionAttrsReadWrite` will be). Isn't it error prone behavior?

And second question, that also related to my implementation of multi-host parsing is "what about case when DNS returns many IPs - how to provide writable check for each of endpoint?". https://github.com/jackc/pgx/pull/545#discussion_r295143354

In pgx v4 and in this PR we have no ability to provide failover scenario when DNS used as service discovery (only one hostname defined in config, but this hostname resolved to many IP's that cover master and all of its replicas).

This problem occurs due to different levels of abstraction: libpq driver resolves hostnames by itself and builds a list of IP's that need to be checked, but pgx driver relies on dialer interface to resolve hostname, so, during connection establishing we don't how many IPs really exists and thus we can't check writability of them. 

We discussed it with @x4m and idea is add Resolver interface to `ConnConfig`, and if resolver is not nil, then use it to fetch list of IPs from hostname and check writability, otherwise, just call DialFunc and no more. This approach will cover all use-cases without breaking behavior expectations. Of course, this complicates understanding of connection configuration.

@jackc what do you think about it?

jackc:
As far as `AfterConnectFunc` in `v4` how I anticipated it being used was `ParseConfig` is called first and assigns `AfterConnectTargetSessionAttrsReadWrite` to `AfterConnectFunc`. Then if the user wants to assign `AfterConnectFunc` the new user function would wrap the old one. I suppose that is a potential foot-gun if it is replaced instead of wrapped.

But I do like the elegance of the libpq behavior being implemented in more general terms of list of configs and an after connect function. This model can implement libpq's strange SSL `prefer` default behavior, `target_session_attrs`, and alternative HA schemes. We could add additional documentation or maybe add a `ValidateConnectionFunc` that is specifically for these HA cases and leave `AfterConnectFunc` to setup the connection.

The multiple IPs per host issue is tricky. I'm not sure if injecting a resolver is the ideal terminology / abstraction. It's not just resolving the host address(es) it is also potentially expanding from 1 to multiple dial attempts. In `v4` this means more `FallbackConfig`s. In `v3` I guess it would be another special case or loop... In both `v3` and `v4` I would envision `target_session_attrs` automatically triggering this alternative DNS resolution mode. As much as possible I would like the same connection string / environment variables to behave the same in libpq and pgx.

One probably unavoidable downside is DialFunc would not have access to the host only the resolved IP. Not sure how often this would matter though -- maybe if someone is doing SSH tunneling or the like. I suppose we could abuse the `ctx` argument to pass host into the dialer (in `v4` anyway)...  😞

furdarius:
I got your point about `AfterConnectFunc` in `v4`, looks reasonable. In this approach, I would prefer to explicitly define `ValidateConnectionFunc`, but as you wish)

> it is also potentially expanding from 1 to multiple dial attempts

`net.Dialer` does multiple connection attempts currently if hostname resolved to many IP's. So, the default behavior will not be changed.

> In v3 I guess it would be another special case or loop... 

[networkAddresses func](https://github.com/jackc/pgx/blob/2837818b67f39d5dfb49a6b37f7d8a23ef263896/conn.go#L144) will just return more net.Addr. So, implementation of loop and writability checking from this PR is enough to cover it. 

> I'm not sure if injecting a resolver is the ideal terminology

I agree with you, but what if user uses [Consul](https://www.consul.io/) or [etcd](https://etcd.io/) as service discovery instead of DNS (_today, this is a fairly common case_). Then if we do not give the ability to customize resolver this case can't be covered.

Options, that we have:
1. Leave PR as is
    * Simple codebase (+)
    * Unchanged behavior (+)
    * DNS as service discovery is not possible (-)
    * Consul/Etcd as service discovery is not possible (-)
    * Is not fully compatible with libpq (-)
2. Add Resolver interface to ConnConfig, but do not resolve hostnames, if not defined
    * Code complication (-)
    * Unchanged behavior (+)
    * DNS as service discovery is possible (+)
    * Consul/Etcd as service discovery is possible (+)
    * Is not fully compatible with libpq (-)
3. Add Resolver interface to ConnConfig, use `net.Resolver` by default, if not defined
    * Code complication (-)
    * Changed behavior (-)
    * DNS as service discovery is possible (+)
    * Consul/Etcd as service discovery is possible (+)
    * Fully compatible with libpq (_equal connstrings - equal behavior_) (+)
4. Do not add Resolver interface to ConnConfig. Always use `net.Resolver` for hostnames resolving. 
    * Not so big code complication (+/-li)
    * Changed behavior (-)
    * DNS as service discovery is possible (+)
    * Consul/Etcd as service discovery is not possible (-)
    * Fully compatible with libpq (_equal connstrings - equal behavior_) (+) ([libpq](https://github.com/postgres/postgres/blob/8255c7a5eeba8f1a38b7a431c04909bde4f5e67d/src/common/ip.c#L57) uses [getaddrinfo](http://man7.org/linux/man-pages/man3/getaddrinfo.3.html) to resolve hostname)

Pick one to go further :D

jackc:
I'd like to minimize changes to v3 in the lead up to v4 so if option 1 meets your needs today then I would prefer that -- the connection system is pretty messy even before this change. If not, option 4 would be my preference (though I suspect that might be harder than it looks -- my guess is we might end up with something like the libpq ` PGHOSTADDR` variable somewhere).

I have no first hand experience needing to integrate service discovery as you describe for options 2 and 3. My initial reaction is that service discovery should be handled independently of the database driver -- but as I said I have no first hand experience there so I could be totally wrong. But if it is necessary, I see that as an independent and additive feature (option 4 becomes option 3 with the addition of an injectable resolver).  So longer term for v4 I see moving to option 4 and eventually 3 if required.

conn.go
furdarius: Continue here https://github.com/jackc/pgx/pull/545#issuecomment-504661449

conn.go
```diff
@@ -262,33 +357,125 @@ func connect(config ConnConfig, connInfo *pgtype.ConnInfo) (c *Conn, err error)
 		}
 	}
 
+	if c.config.TargetSessionAttrs != "" &&
```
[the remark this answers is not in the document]
furdarius: Defined `TargetSessionAttrs` property as custom type (`TargetSessionType`) instead of `string`. Now all validation logick is inside `isValid` method for this type.

I'm not sure, if I can do it better.

conn.go
furdarius: I got you point, but want to leave this function called `checkWritable`, cuz all other options don't helps to read and understand code.

furdarius:
I decided not to add any kind of hosts resolving in this PR, but will start work on it in next PR.
Important TODOs closed, so PR is ready to review.

Just to be sure, that all works as expected, I built PG HA group (master + standby) and tested failover situation with `target_session_attrs=read-write`. Driver successfully reconnects to promoted standby on master failure.

Asciinema rec with failover process:
[asciicast](https://asciinema.org/a/Mtvkgcu1YHmR9AYLXIm73kQZR)

jackc:
LGTM, thanks!

pashagolub:
The upcoming v14 introduces new values for `targetSessionAttrs`. How should we approach implementing this functionality? 
https://www.cybertec-postgresql.com/en/new-target_session_attrs-settings-for-high-availability-and-scaling-in-postgresql-v14/

furdarius:
pgx v4 works well with `targetSessionAttrs`, so there are no problems here.

You can check [https://github.com/jackc/pgconn](pgconn), to find out how it's implemented and should be used:

https://github.com/jackc/pgconn/blob/5daa019e4eb52df3409ebf17c83116b7c0e827e5/pgconn_test.go#L338

pashagolub:
Sorry, I probably wasn't clear enough. There are new attributes in the upcoming v14, so do we need to handle that new values somehow or the code is universal?

Thanks in advance!

furdarius:
Ahh, yeah, I see now.

> Since this functionality is part of the client C library libpq, all clients that link with that library automatically support this connection parameter.

And it's not the case for the pgx (pgconn), so it makes sense to add it.

As I see from the [commit](https://git.postgresql.org/gitweb/?p=postgresql.git;a=blobdiff;f=src/interfaces/libpq/fe-connect.c;h=9812a14662dd3b669d9e82d346fb84880e3e1dcb;hp=db71fea169c96f1d417f843ee2ef9928efe4ceef;hb=ee28cacf619f4d9c23af5a80e1171a5adae97381;hpb=57e6db706e81fd2609fa385677e6ae72471822fe), to maintain the functionality we should check both `SHOW transaction_read_only` and `SELECT pg_catalog.pg_is_in_recovery()` depending on the value of target server.

> "transaction_read_only = on" proves that at least one of default_transaction_read_only and in_hot_standby is on,
> but we don't  actually know which.
> We don't care though for the purpose of identifying a read-only session,
> so satisfy the CONNECTION_CHECK_TARGET code by claiming they are both on.
> On the other hand, if it's a read-write session, they are certainly both off.


> If the server didn't report "default_transaction_read_only" or "in_hot_standby" at startup,
> we must determine its state by sending the `query "SHOW transaction_read_only"`

Anyway, it makes sense to create a brand new issue to the https://github.com/jackc/pgconn with the functionality description, and continue the discussion there.

PR #547, opened 2019-06-24
cast bytea to make []byte suitable for both string and binary string

kingluo:
Chances are that we input `[]byte` for string type, e.g. `text`, so it's better to cast the quoted bytes explicitly.

### DIFF
--- a/internal/sanitize/sanitize.go
+++ b/internal/sanitize/sanitize.go
@@ -87,7 +87,7 @@ func QuoteString(str string) string {
 }
 
 func QuoteBytes(buf []byte) string {
-	return `'\x` + hex.EncodeToString(buf) + "'"
+	return `'\x` + hex.EncodeToString(buf) + "'::bytea"
 }
 
 type sqlLexer struct {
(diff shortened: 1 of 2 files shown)


jackc:
Sorry, I don't understand the intention here. Do you have an example use case that this will improve?

kingluo:
For example, I have a column in `text` format, but sometimes I would insert string in `bytea` format, maybe due to escaped data input, or some binary source input which I don't want to do string cast (as known in golang, the string cast would do copy, for big string, it's a bit expensive.).
If `SanitizeSQL` does not put `::bytea` suffix to the quoted `[]byte` argument, postgres would store the data literally into that column, and it's a wrong expectation.
For example, I input "hello" in `[]byte`: [104 101 108 108 111], then `SanitizeSQL` would quote it as `'\x68656c6c6f'`, it's not equal to "hello" in text anymore. But with `::bytea` suffix, it works (it also works for `bytea` column, so nothing breaks).

jackc:
Okay, seems fair. Thanks.

PR #548, opened 2019-06-24
QueryEx: send simple query if len(args) is 0

kingluo:
Why not send simple query if no arguments provided, just like what `conn.execEx()` already did?

### DIFF
--- a/query.go
+++ b/query.go
@@ -452,6 +452,15 @@ func (c *Conn) QueryEx(ctx context.Context, sql string, options *QueryExOptions,
 		return rows, nil
 	}
 
+	if len(args) == 0 {
+		c.lastStmtSent = true
+		if err := c.sendQuery(sql, args...); err != nil {
+			rows.fatal(err)
+			return rows, err
+		}
+		return rows, nil
+	}
+
 	ps, ok := c.preparedStatements[sql]
 	if !ok {
 		var err error


jackc:
I think this fails if the sql text is a prepared statement name.

It would not always be a performance win either. It's a trade-off. You save a network round trip in exchange for using the text instead of binary format for the result. `execEx` never has a result set so saving the round trip is (almost) always the better choice.

If you specifically want to use the simple protocol this can be done by setting `SimpleProtocol`.

kingluo:
> I think this fails if the sql text is a prepared statement name.
> 

Yes, because pgx uses the same interface for prepared execution and normal sql string query. Why not put the prepared interface on `PreparedStatement`?

> It would not always be a performance win either. It's a trade-off. You save a network round trip in exchange for using the text instead of binary format for the result. `execEx` never has a result set so saving the round trip is (almost) always the better choice.

I have no benchmarks to prove it. But in my experience, it's almost faster than two-round-trips prepare-and-execute. And the lib/pq also does so:
https://github.com/lib/pq/blob/master/conn.go#L825

> If you specifically want to use the simple protocol this can be done by setting `SimpleProtocol`.

`sanitizeAndSendSimpleQuery` always parse the sql string even if there is no arguments, so is it better do a check there and just `sendSimpleQuery` if no arguments?

bfontaine:
Almost 4 years later, does it make sense to keep this PR open?

Issue #420, opened 2018-05-09
Registering composite types


dvic:
I noticed in https://github.com/jackc/pgx/blob/master/conn.go#L415 that composite types, e.g., 
```
	CREATE TYPE rgb AS (
		r integer,
		g integer,
		b integer
	);
```
are not included (these are represented with `typtyp='c'`).

Is there are particular reason for this?

Adding `'c'` to this list seems to make my custom types automatically work, without registering the OID (as long as they implement `pgtype.TextDecoder`).

jackc:
It used to include them. The reason for the change is every table also creates a type. Connecting to a database with 1000's of tables would experience an unreasonable slowdown.

See #140 for the original issue.

dvic:
Ah I see, do you think it's worth it to include a whitelist in the config so that the type names in this list  are included? Or maybe ship a `AfterConnect` helper that does this?

jackc:
I'm not sure. Does the type even need to be registered? As long as the values you are scanning into implement `pgtype.TextDecoder` it should work without it.

johanbrandhorst:
In my experience the composite types does need to be registered for scanning into, at least when used through the `sql.DB` interface. I tries to lookup the `oid` in `ConnInfo` and fails at:
https://github.com/jackc/pgx/blob/aa561b8f3c545aad9e0b5fb1514e440d330dd016/conn.go#L1130

This seems to be happening no matter how well instrumented the scanned-into-type is.

dntj:
If we change the query [here](https://github.com/jackc/pgx/blob/master/conn.go#L400) to:
```sql
select t.oid,
	case when nsp.nspname in ('pg_catalog', 'public') then t.typname
		else nsp.nspname||'.'||t.typname
	end
from pg_type t
left join pg_type base_type on t.typelem=base_type.oid
left join pg_namespace nsp on t.typnamespace=nsp.oid
left join pg_class cls on t.typrelid = cls.oid
where (
		t.typtype in('b', 'p', 'r', 'e', 'c')
		and (base_type.oid is null or base_type.typtype in('b', 'p', 'r'))
		and (cls.oid is null or cls.relkind = 'c')
	)
```

this works for composite types in simple cases I've tested.  Happy to put together a PR if you think this'll work generally.

jackc:
Seems reasonable.

PR #549, opened 2019-06-25
Read OIDs for composite types on connection init.

dntj:
This used to be done, but pulled in tables which slowed down connections on databases with a large number of tables; see https://github.com/jackc/pgx/issues/140.

This change includes composite types but excludes tables by joining against [pg_class](https://www.postgresql.org/docs/11/catalog-pg-class.html) in which `relkind` is `'c'` for the former and `'r'` for the latter.

Fixes https://github.com/jackc/pgx/issues/420.


jackc:
Thanks!

PR #555, opened 2019-07-04
Use zap.Any for handling interface{} -> zap.Field conversion

NWilson:
zap.Any falls back to zap.Reflect, but is better for this case, because
it first checks for the types that zap handles specially.  For example,
time.Duration, or error, which zap.Reflect will just treat as untyped
int64 or struct objects, but zap.Any is able to detect these types and
print them properly.


jackc:
Thanks!

fxrlv:
fix not merged into v4 and master
@jackc

PR #556, opened 2019-07-05
Register composite types as a `pgtype.Record`.

DavidArchibald:
This registers composite types as a `pgtype.Record` instead of leaving it to the whims of `ConnInfo.InitializeDataTypes` which will default to `pgtype.GenericText`, which is not a very usable format. I found in testing that using it as record would work well for a composite type but if a new type (ie a `pgtype.Composite`) would be preferred I would readily implement that as well.

Most of the code has been modelled around how Enum Arrays are registered in `ConnInfo.initConnInfoEnumArray` and PR #420 so it should mesh well with the rest of the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,8 @@ if err != nil {
 
 ## v4 Coming Soon
 
-This is the current stable v3 version. v4 is currently is in prelease status. Consider using [v4](https://github.com/jackc/pgx/tree/v4) for new development or test upgrading existing applications.
+This is the current stable v3 version. v4 is currently is in release candidate status. Consider using
+[v4](https://github.com/jackc/pgx/tree/v4) for new development or test upgrading existing applications.
 
 ## Features
 
--- a/batch.go
+++ b/batch.go
@@ -135,7 +135,7 @@ func (b *Batch) Send(ctx context.Context, txOptions *TxOptions) error {
 
 	_, err = b.conn.conn.Write(buf)
 	if err != nil {
-		b.conn.die(err)
+		b.die(err)
 		return err
 	}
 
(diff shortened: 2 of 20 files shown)


jackc:
Hmmm... Composite types are tricky. `pgtype.Record` **is** the correct type in pgx, but it is somewhat limited. It can only be used to decode values sent via the binary format (it will fail to decode the text format and cannot encode in either format). So I'm not sure if there would be any unfortunate implications about registering it by default.

Beyond that, can you give me an example usage that this makes easier? Scanning into a `pgtype.Record` should work without it.

DavidArchibald:
Well, it's helpful and more understandable when reading a bit into the internals; composites will seem to be a `pgtype.GenericText` instead of a `pgtype.Record` or some more descriptive type. This is largely important for frameworks around the library, but I don't think should be refused simply because you *can* scan things out as a `pgtype.Record`.

DavidArchibald:
I'm now on my development machine and I rediscovered the initial problem I was trying to solve. You actually *can't* scan into a `pgtype.Record`, at least in the intuitive way, perhaps there are other ways to do so? I was misled because when I tried to reproduce on my local computer I'd forgotten.

```
var compositeExample pgtype.Record
row := conn.QueryRow("SELECT compositetype FROM exampletable;")
err := row.Scan(&compositeExample)
// err ~= "can't scan into dest[i]: cannot decode &pgtype.Text{String:"(\"example\", \"composite\")", Status:0x2} into *pgtype.Record
```

This example does work if you scan into a `*pgtype.Text` instead, but then you just get a string that looks like `("example", "composite")` and you'd have to parse it yourself. I don't think that this is a normally desired behavior.

jackc:
Okay, I think I understand what is going on now. `pgtype.Record` can only decode the binary format. Since the OID isn't registered, the value is retrieved in the text format.

I was expecting it to work because it *does* work with records constructed with the row constructor. e.g. `select row(1,2,3)`. But that is because generic record is a registered type.

So I guess my earlier assertion is wrong. I don't think pgtype.Record is the correct Go type for PostgreSQL composite types. Generic records have several distinct disadvantages.

1. They can't written, only read.
2. The fields are unnamed.

I would recommend an application using composite types to implement the appropriate pgtype interfaces on their own struct instead of using a generic record type.

That said, I guess it doesn't really hurt anything to have limited read support built in. So I suppose I'm still okay with merging this change -- assuming we can get the tests passing.

DavidArchibald:
I apologize for very delayed time period, I had the changes uncommitted and forgot to push them; this also lead to the messy merge which I force pushed away to prevent cluttering of the history due to more of my mistakes. I hope this is acceptable now. I would not mind a squash commit if you wish to do so, as it has more merges than commits.

jackc:
LGTM - thanks!

Issue #546, opened 2019-06-22
[v4] Configurable custom OID's fetching


furdarius:
In _v3_ of pgx driver on connection establishing driver sent [request to fetch OIDs](https://github.com/jackc/pgx/blob/5044e8473ad948114b6cb63f6f30f94fc7834667/conn.go#L398) from DB. In new version (_v4_), known OIDs registered via [pgtype.NewConnInfo()](https://github.com/jackc/pgtype/blob/4ed0de4755e042908ec6b12c68c0f900b7078726/pgtype.go#L175).

Is the list of hardcoded OID's complete enough? In postgres sources, there are more enumerated ids: https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat

Do I understand correctly that without additional OID's fetching, domain types and enums are not available?

Maybe it is worth to make configurable fetching of additional OID's?

In my practice, in _v3_ of pgx driver I prefer to limit or completely disable fetching of additional OID's via [CustomConnInfo](https://github.com/jackc/pgx/blob/5044e8473ad948114b6cb63f6f30f94fc7834667/conn.go#L79), cuz this query heavily loads the database due to locks. This approach (_disable OID's fetching_) works if we set restrictions about the usage of domain types and enums. Otherwise, in certain conditions (OID's fetching on each connection establishing, unlimited connection pool, db under heavy load), this will cause the database to become unavailable. Once this happened to one of our database:

(_This graph shows the number of requests for additional OID's fetching._)
![downtime](https://user-images.githubusercontent.com/2843971/59962938-e36ccc00-94f4-11e9-9217-254c449d3b4b.png)

jackc:
You are correct about the current reasoning regarding enum and domain types. You are also correct that hard-coding the standard OIDs is probably a better approach than querying on every connection. In fact, that change is already in place for v4.

Regarding making the change for v3, I'm open to it if it can be done without breaking backwards compatibility (not sure how plausible that is).

Regarding the performance problem you are seeing, I would not have expected it to be so severe. As I recall, the connection pool caches the OID information and only makes the query on the first connection.

furdarius:
> In fact, that change is already in place for v4.

In v4, as I see, there is no fetching of additional OID's. I suggest to make it configurable (disabled by default).

And, by the way, maybe the list of hardcoded OID's should be extended? In postgres sources, there are more OID's: https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat
 
[...]

PR #561, opened 2019-07-14
Reuse pool.connInfo for createConnectionUnlocked method

furdarius:
`ConnPool.createConnection` reuse `p.connInfo` already. So, `createConnectionUnlocked` must be consistent with it.

Otherwise, we inefficiently fetch data from DB on connection establishing. This could lead to performance degradation and downtime.

As I already notes in #546 once we got downtime due to the increased number of OID's fetching requests. But we used `database/sql` connection pool, that has no ability to reuse `connInfo` and connection established with `minimalConnInfo` each time.

(_This graph shows the number of requests for additional OID's fetching._)
downtime

### DIFF
--- a/conn_pool.go
+++ b/conn_pool.go
@@ -319,7 +319,7 @@ func (p *ConnPool) createConnection() (*Conn, error) {
 func (p *ConnPool) createConnectionUnlocked() (*Conn, error) {
 	p.inProgressConnects++
 	p.cond.L.Unlock()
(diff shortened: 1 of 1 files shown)


furdarius:
Arghh, race error. I'll try to find any workaround.

furdarius:
Works perfectly with `DeepCopy`!

jackc:
Thanks!

PR #563, opened 2019-07-24
Add conversions that UUID types from libraries can be handled by pytype.UUID

NWilson:
This is similar in spirit to all the existing pgtype conversion code.

It's just that arrays (not slices) of primitive types aren't currently handled.

In my case, I was using the popular Google UUID library (https://github.com/google/uuid), which wraps `[16]byte`.

### DIFF
--- a/pgtype/convert.go
+++ b/pgtype/convert.go
@@ -149,7 +149,7 @@ func underlyingTimeType(val interface{}) (interface{}, bool) {
 	switch refVal.Kind() {
 	case reflect.Ptr:
 		if refVal.IsNil() {
-			return time.Time{}, false
+			return nil, false
 		}
 		convVal := refVal.Elem().Interface()
 		return convVal, true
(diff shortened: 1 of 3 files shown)


pgtype/convert.go
```diff
@@ -149,7 +149,7 @@ func underlyingTimeType(val interface{}) (interface{}, bool) {
 	switch refVal.Kind() {
 	case reflect.Ptr:
 		if refVal.IsNil() {
-			return time.Time{}, false
+			return nil, false
```
jackc: This change seems unrelated. Is it intentional?

NWilson: Hi, sorry for the late reply.

Yes, it is intentional, it's in a separate commit. (I could have made a separate PR but it's so minor...) The first return value is ignored when the second is false - so there's no point heap-allocating a zero time.Time that will be thrown away by the caller. It's possibly just a mistake/holdover from when the function's return signature was `(time.Time,bool)`?

jackc:
Thanks!

Issue #502, opened 2019-01-20
Is it possible to configure underlying ChunkReader with custom MinBufLen?


furdarius:
As I can see, it's not possible to use `NewFrontend` with custom `ChunkReader` https://github.com/jackc/pgx/blob/f4396b2e23b1e53a3047c86eb025b682344eaa56/conn.go#L331

And beside that, it's even not possible to use custom `Frontend` in `Conn.connect`.

If you find it useful to give the ability to configure `ChunkReader`, I can create patch.

Maybe I'm missing something and it's possible to configure `MinBufLen` for `ChunkReader`?

furdarius:
Okey, got it. I'll try to collect metrics some metrics which this setting may affect.
Prepared a fork for experimentation: #508

jackc:
This is doable through the pgconn connection now.

PR #565, opened 2019-07-29
Set 8KB as default size of ChunkReader buffer

furdarius:
The reason for these changes is size of send buffer inside Postgresql (**8KB**): https://github.com/postgres/postgres/blob/249d64999615802752940e017ee5166e726bc7cd/src/backend/libpq/pqcomm.c#L134

So, if in driver we can't configure it, then the best option would be set at least as big as the Postgres has itself.

Some discussion and benchmarks could be found here https://github.com/jackc/pgconn/pull/10#issue-301561353 

RelatedTo: https://github.com/jackc/pgx/issues/502

### DIFF
--- a/chunkreader/chunkreader.go
+++ b/chunkreader/chunkreader.go
@@ -28,7 +28,11 @@ func NewChunkReader(r io.Reader) *ChunkReader {
 
 func NewChunkReaderEx(r io.Reader, options Options) (*ChunkReader, error) {
 	if options.MinBufLen == 0 {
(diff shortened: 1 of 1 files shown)


jackc:
Seems reasonable.

lozenko:
@jackc can we merge it finally? thanks

jackc:
@lozenko  I'm guessing you mean the PR on chunkreader as this one was merged 6 months ago... Done.

PR #572, opened 2019-08-07
Allow specifying sql.LevelRepeatableRead in BeginTx

mastercactapus:
When testing a migration from lib/pq I found that specifying `sql.LevelRepeatableRead` comes back with an unsupported isolation error.

This PR adds `sql.LevelRepeatableRead` to map to `pgx.RepeatableRead` along side `sql.LevelSnapshot`.

I tried digging to see if it was intentionally left out, but came up empty.

Since Postgres [docs](https://www.postgresql.org/docs/current/transaction-iso.html) reference `RepeatableRead`, and users coming from lib/pq can only use `LevelRepeatableRead` (pq does not allow/recognize `LevelSnapshot`) I think it makes sense for pgx/stdlib to support specifying it in `BeginTx`.


jackc:
LGTM. Thanks!

PR #576, opened 2019-08-11
Correct WaitForNotification example

sigmavirus24:
While working on a project that was using this, I tried using the
example code but instead found that WaitForNotification expects a
Context (which makes sense).

This corrects the docs for folks using that as a jumping off point.


jackc:
Thanks!

Issue #580, opened 2019-08-16
"Unknown oid" error on user-defined array type 


damour:
Database state:
```
CREATE TYPE custom_type AS (some_key int);

CREATE TABLE "table_with_custom_array_type" (array_column custom_type[]);
```

Example:
```
package main

import (
	"database/sql"
	"log"

	_ "github.com/jackc/pgx/stdlib"
)

func main() {
	conn, err := sql.Open("pgx", "...")
	if err != nil {
		log.Fatal(err)
	}

	_, err = conn.Query("SELECT * FROM table_with_custom_array_type")
	if err != nil {
		log.Fatal(err)
	}
}
```

Failed with: `2019/08/16 13:10:28 unknown oid: 106004, name: array_column`

PGX version: 3.5.0

`select oid, typname, typelem, typtype from pg_type t WHERE typname IN ('custom_type', '_custom_type')`
```
oid	  typname      typelem  typtype
106005	custom_type	0      	c
106004	_custom_type	106005	b
```

PR #581, opened 2019-08-16
Load user-defined array type oids.

damour:
Fix https://github.com/jackc/pgx/issues/580

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -611,12 +611,14 @@ func initPostgresql(c *Conn) (*pgtype.ConnInfo, error) {
 	end
 from pg_type t
 left join pg_type base_type on t.typelem=base_type.oid
+left join pg_class base_cls ON base_type.typrelid = base_cls.oid
 left join pg_namespace nsp on t.typnamespace=nsp.oid
 left join pg_class cls on t.typrelid=cls.oid
 where (
 	  t.typtype in('b', 'p', 'r', 'e', 'c')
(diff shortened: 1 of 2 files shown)


jackc:
Is it possible to do what you want without loading array types for all tables?

We previously had a problem where it was extremely slow to connect to databases with large numbers of tables: #140. This change would return to the behavior before that was fixed.

damour:
Sorry for late response I was on vacation.
Yes, I think it's possible something like this:

```
select t.oid,
	case when nsp.nspname in ('pg_catalog', 'public') then t.typname
		else nsp.nspname||'.'||t.typname
	end
FROM
  pg_type t
LEFT JOIN pg_type base_type
 ON t.typelem = base_type.oid
LEFT JOIN pg_class base_cls
 ON base_type.typrelid = base_cls.oid
LEFT JOIN pg_namespace nsp
 ON t.typnamespace = nsp.oid
LEFT JOIN pg_class cls
 ON t.typrelid = cls.oid
WHERE (
  t.typtype IN ( 'b','p','r','e','c' ) and (
      base_type.oid is null or base_type.typtype IN ( 'b','p','r','c' )
  )
  and (cls.oid is null or cls.relkind = 'c') 
  and (base_cls.oid is null or base_cls.relkind = 'c') 
)
```
If this is appropriate solution I can update MR.

jackc:
I think this would work.

damour:
Done, PTAL.

jackc:
Thanks. LGTM.

PR #586, opened 2019-08-25
wrap errors support both go1.13 and go1.12

tsingson:


jackc:
I'm not sure I understand what this is trying to do. I thought the `golang.org/x/xerrors` already provided a backwards compatible version of the Go 1.13 error system.

I'm testing on Travis against Go 1.12 and tip and it seems to be working...

https://travis-ci.org/jackc/pgx

tsingson:
golang.org/x/xerrors working fine in go1.12 but  xerrors.Errorf no work in go1.13 rc1

jackc:
Whatever the issue was it appears to not affect the final release of 1.13. It is working now with the existing code.

PR #589, opened 2019-08-28
Include ParameterOIDs when preparing statements on new pool connections

vcabbage:
ParameterOIDs passed to ConnPool.PrepareEx are used to prepare the statement on existing connections in the pool. If additional connections are later created ParameterOIDs are omitted, potentially causing query failures.


vcabbage:
Thank you for the quick merge and for the project!

PR #598, opened 2019-09-12
Fixes #376

### LINKED ISSUE #376: ParseDSN doesn't support single quotes, double quotes not handled correctly
Hi,

thanks for this driver. I've been trying to use it with our DSN connection strings, but there seems to be some issues with ParseDSN.

According to https://www.postgresql.org/docs/10/static/libpq-connect.html#LIBPQ-CONNSTRING values containing spaces should be quoted with single quotes. That's not supported by ParseDSN in this package. Instead it supports double quotes, but unfortunately the double quotes are included as part of the value, which then becomes invalid, at least in some cases.

**Program to reproduce**

```go
package main

import (
        "fmt"

        "github.com/jackc/pgx"
)

func main() {
        c, _ := pgx.ParseDSN(`host='foo bar' options="-c search_path=fooschema,public"`)
        fmt.Printf("%s\n", c.Host)
        fmt.Printf("%s\n", c.RuntimeParams["options"])
}
```

**Expected output**

```
foo bar
-c search_path=fooschema,public
```

or I suppose " isn't needed to be supported... it's not mentioned in the link above.

**Actual output**

```
'foo
"-c search_path=fooschema,public"
```

jbarone:
Replaced regex DSN parser with a simple parser based on how libpq handles parsing the DSN. This works with `key=value` pairs as well as `key='value'` pairs. It doesn't work with Double Quotes, as that is not listed in the documentation for PostgreSQL. If required, it should be straight forward for it to be added.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,7 +17,6 @@ import (
 	"os/user"
 	"path/filepath"
 	"reflect"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

jbarone:
Glad I could help. Hope to dig in and contribute more.

PR #599, opened 2019-09-13
pgtype: Add ext type for gofrs uuid implementation

heedson:
Add ext type for https://github.com/gofrs/uuid uuid type.

Change test and README from `github.com/satori/go.uuid` to `github.com/gofrs/uuid`. The reason behind this change is due to this issue:
https://github.com/satori/go.uuid/issues/73
This was taken on board and fixed in the community project of `gofrs`. The `gofrs` implementation has the same interface as the original.

Would be happy to separate out the ext addition and the satori removal from the main tests but I thought both would bring benefits, albeit on the minor side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,11 @@ skip tests for connection types that are not configured.
 To setup the normal test environment, first install these dependencies:
 
     go get github.com/cockroachdb/apd
+    go get github.com/gofrs/uuid
     go get github.com/hashicorp/go-version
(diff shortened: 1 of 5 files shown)


heedson:
I'm unsure if the removal of satori from the README "normal test environment" is appropriate. If it's wrong, I can adjust it.

jackc:
Thanks!

PR #608, opened 2019-09-18
Punctuation and word fixes.

DavidArchibald:
Adds periods after some sentence fragments for consistency with those sentences ending with periods. Secondly, fixes two minor wording mistakes(is instead of as, and the omission of or).

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,25 +58,25 @@ encode and decode values for PostgreSQL.
 
 pgx supports many additional features beyond what is available through database/sql.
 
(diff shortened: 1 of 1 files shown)


jackc:
The wording fixes and punctuation on sentences looks good, but I'm pretty sure that non-sentence bullet points should _not_  have periods at the end.

DavidArchibald:
Sure, will fix, I just added them because one of the bullet points did.

jackc:
Thanks!

PR #615, opened 2019-09-26
conninfo race bugfix

kingluo:
In v3, the conninfo is initialized to static variable `minimalConnInfo`, and get refreshed as new clone when receiving `ReadyForQuery`.

But in replication connection case, it would retain  `minimalConnInfo`, so that if I do `RegisterDataType()` for different replication connections from different goroutines, it would be a data race. 

I add a `RWMutex` to protect it.

### DIFF
--- a/pgtype/pgtype.go
+++ b/pgtype/pgtype.go
@@ -2,6 +2,7 @@ package pgtype
 
 import (
 	"reflect"
+	"sync"
 
 	"github.com/pkg/errors"
 )
@@ -142,6 +143,7 @@ type DataType struct {
 }
 
 type ConnInfo struct {
+	sync.RWMutex
 	oidToDataType         map[OID]*DataType
 	nameToDataType        map[string]*DataType
 	reflectTypeToDataType map[reflect.Type]*DataType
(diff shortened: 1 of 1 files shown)


jackc:
`ConnInfo` isn't designed to be concurrency safe. It looks to me like the bug is more that `minimalConnInfo` isn't getting duplicated.

There already is a `DeepCopy` method on `ConnInfo`. Would it make more sense to use that on `minimalConnInfo` when a replication connection is established?

kingluo:
`DeepCopy()` could not help, it just returns a new conninfo, but not bound on the connection. I need to register new type on the replication connection before doing query on it, otherwise it would reports error.

For normal connections, concurrent unsafe is ok, but for replication connections, they all point to the same static `minimalConnInfo`, then no safe way to register types after connection established, the golang runtime reports concurrent map access error.

jackc:
This is what I mean: 23388fecf653ddfce2a4149acf71313b87d64e3d

Does that resolve the problem of multiple connections sharing `minimalConnInfo`?

kingluo:
> This is what I mean: [23388fe](https://github.com/jackc/pgx/commit/23388fecf653ddfce2a4149acf71313b87d64e3d)
> 
> Does that resolve the problem of multiple connections sharing `minimalConnInfo`?

Yes. Please merge it into v3 branch, thanks.

PR #623, opened 2019-10-03
Fix panic in certain circumstances

jaltavilla:
The panic occurs when three things are true:
1. The passed in arg is a pointer to nil, but not a nil interface.
2. The passed in arg is of a type that implements BinaryEncoder and/or TextEncoder.
3. The method that implements the interface takes a non-pointer receiver.

When these three things are true, the initial nil check does not see arg as nil. It then converts arg to an Encoder interface. The call to encode then panics because the runtime needs to convert arg to a non-pointer, but arg is pointing to nil.

Callers could try to work around this by making their implementation take a pointer receiver. This prevents the panic. However, in that case the type cannot be passed by value to Exec, Query or QueryRow. (This is because the method set of a non-pointer does not include the encode implementation, since it takes a pointer).

Arguably, checking if the interface is nil or a typed nil first is logically correct. In both cases NULL should be sent to the database, and nothing else needs to be done. I suspect there might be other places (like `encodePreparedStatementArgument`) that could panic similarly.

### DIFF
--- a/extended_query_builder.go
+++ b/extended_query_builder.go
@@ -67,6 +67,13 @@ func (eqb *extendedQueryBuilder) encodeExtendedParamValue(ci *pgtype.ConnInfo, o
 		return nil, nil
 	}
 
+	refVal := reflect.ValueOf(arg)
+	argIsPtr := refVal.Kind() == reflect.Ptr
+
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

PR #626, opened 2019-10-13
Readme: update pkg paths for pgtype & pgmock

### LINKED ISSUE #607: Missing pgmock in version 4.0.0
In previous versions, there has always been an existing folder github.com/jackc/pgx/pgmock, but this seems to be removed in the latest release four days ago. However, it still exists in the documentation.

Is it possible to provide some guidance -- is this logic now removed or has it been migrated somewhere?

This was unfortunately causing a breaking change for our team so any help would be greatly appreciated!

tbutts:
With the big v4 changes, some libraries that were part of pgx were migrated to standalone packages. The readme has been updated to reflect these new import paths, which will help consumers of the old pgtype/pgmock paths fix their own import statements.

-----

Thanks for the huge amount of work that has gone into pgx, it has been one of my go-to staples for years and I'm really excited to start working with v4! I wish I could help out more than just updating the readme, but perhaps one day I'll find some piece I can lend a hand with.

Resolves #607

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,9 +108,9 @@ pgxpool is a connection pool for pgx. pgx is entirely decoupled from its default
 
 database/sql compatibility layer for pgx. pgx can be used as a normal database/sql driver, but at any time the native interface may be acquired for more performance or PostgreSQL specific functionality.
 
-## github.com/jackc/pgx/pgtype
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

PR #640, opened 2019-11-22
Clean up two test cases

apantel:
In TestListenNotifyWhileBusyIsSafe, the existing behavior is a race condition that may panic if the database does not support listen/notify.

In TestConnQueryErrorWhileReturningRows, if the conversion to PgError fails, then the PgError is nil. Also if the Scan fails, then the output was confusion.

This is a first pass for compatibility with CockroachDB.

---
This change is [Reviewable](https://reviewable.io/reviews/jackc/pgx/640)


jackc:
Thanks!

PR #646, opened 2019-12-07
Add back GetDefaultDriver

### LINKED ISSUE #645: v4: Add back `stdlib.GetDefaultDriver`
`v3` exposed the convenience function [stdlib.GetDefaultDriver](https://github.com/jackc/pgx/blob/23388fecf653ddfce2a4149acf71313b87d64e3d/stdlib/sql.go#L131), which was useful when wrapping the driver to add metrics and tracing (see https://github.com/opencensus-integrations/ocsql and https://github.com/luna-duclos/instrumentedsql). `v4` removed this function, but I would argue that it is still useful to users, and would like to have it back.

There is a workaround, which is to create an `sql.DB` with `stdlib.OpenDB` or `sql.Open` and calling `sql.Driver()`, before closing the `DB` again. This is, of course, quite grotesque.

I'd be happy to implement this if this proposal is accepted.

johanbrandhorst:
This method is useful when it is necessary
to wrap the pgx stdlib driver, for example to
add tracing and metrics.

Fixes #645

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -114,6 +114,12 @@ var (
 	fakeTxConns map[*pgx.Conn]*sql.Tx
 )
 
+// GetDefaultDriver returns the driver initialized in the init function
+// and used when the pgx driver is registered.
+func GetDefaultDriver() driver.Driver {
+	return pgxDriver
+}
+
 type Driver struct {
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

PR #657, opened 2020-01-03
Add arguments count check to execParamsAndPreparedPrefix

### LINKED ISSUE #656: Panic in Exec method if pass more arguments than expected
I catch a panic, when trying to make a query using `Exec` method: 
`panic: runtime error: index out of range [1] with length 1`

pgx version is 4.1.2 and it also reproduces in the master branch.

The code that reproduces the error:
```golang
poolConn, _ := pgxpool.ParseConfig("host=127.0.0.1 user=pgx")
poolConn.MaxConns = 10
pool, _ := pgxpool.ConnectConfig(context.Background(), poolConn)

var (
	test1 int
	test2 string
)

_, err = pool.Exec(context.Background(), "SELECT id FROM pdata WHERE age=$1 LIMIT 1", test1, test2)
if err != nil {
	log.Println(err)
}
```

I did a little digging and it looks like a length check (like in Query method) was forgotten:
```golang
if len(sd.ParamOIDs) != len(args) {
	rows.fatal(errors.Errorf("expected %d arguments, got %d", len(sd.ParamOIDs), len(args)))
	return rows, rows.err
}
```

I added this check to the `execPrepared` method to fix the panic. I chose `execPrepared` because there are two branches with statement description in the upper `exec` method.

I can add some tests and send a PR for review if that's ok

malstoun:
Fix #656

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -501,6 +501,10 @@ func (c *Conn) execSimpleProtocol(ctx context.Context, sql string, arguments []i
 }
 
 func (c *Conn) execParamsAndPreparedPrefix(sd *pgconn.StatementDescription, arguments []interface{}) error {
+	if len(sd.ParamOIDs) != len(arguments) {
+		return errors.Errorf("expected %d arguments, got %d", len(sd.ParamOIDs), len(arguments))
+	}
+
 	c.eqb.Reset()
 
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #665, opened 2020-01-23
Fixes #664 -- implement DriverContext for stdlib

### LINKED ISSUE #664: stdlib: Support DriverContext interfaces
https://golang.org/pkg/database/sql/driver/#DriverContext

Right now there's no way to pass a `ctx` down through the `stdlib` interface. If `DriverContext` were supported, this would be possible!

alex:


alex:
I wasn't quite sure how to write a test case for this, since it's an optional interface, and what you really want to test is that the stdlib uses it.

jackc:
Thanks, LGTM!

alex:
Thanks for the quick merge!

PR #667, opened 2020-01-24
Fix Rows.Values() fallback to generic decoder

### LINKED ISSUE #666: panic: *pgtype.Record is not pgtype.TextDecoder
```
package main

import (
	"context"

	"github.com/jackc/pgx/v4"
)

func main() {
	ctx := context.Background()
	conf, _ := pgx.ParseConfig("user=postgres host=localhost port=5432 client_encoding=UTF8")
	conf.PreferSimpleProtocol = true
	conn, err := pgx.ConnectConfig(ctx, conf)
	if err != nil {
		panic(err)
	}
	rows, _ := conn.Query(ctx, query)
	rows.Next()
	rows.Values()
}

const query = `SELECT (ARRAY[1::OID], NULL)`
```

```
panic: interface conversion: *pgtype.Record is not pgtype.TextDecoder: missing method DecodeText

goroutine 1 [running]:
github.com/jackc/pgx/v4.(*connRows).Values(0xc00011d138, 0x80e301, 0xc000016120, 0x7a0035, 0x1c, 0x0)
	/home/mjibson/go/pkg/mod/github.com/jackc/pgx/v4@v4.2.1/rows.go:217 +0x60e
main.main()
	/home/mjibson/go/src/github.com/cockroachdb/t/t.go:19 +0x101
```

This isn't the [first](https://github.com/jackc/pgtype/issues/7) error of this type. It keeps happening because the assertion at https://github.com/jackc/pgx/blob/69048c281c7cc134b9dd4ed3883e1c639c211d75/rows.go#L217 doesn't have an `, ok` clause. It should gain that, something like:

```
decoder, ok := value.(pgtype.TextDecoder)
if !ok || decoder == nil {
...
```

mjibson:
fixes #666

### DIFF
--- a/query.go
+++ b/query.go
@@ -312,8 +312,8 @@ func (rows *Rows) Values() ([]interface{}, error) {
 
 			switch fd.FormatCode {
 			case TextFormatCode:
-				decoder := value.(pgtype.TextDecoder)
-				if decoder == nil {
+				decoder, ok := value.(pgtype.TextDecoder)
+				if !ok {
 					decoder = &pgtype.GenericText{}
 				}
 				err := decoder.DecodeText(rows.conn.ConnInfo, buf)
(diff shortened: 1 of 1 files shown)


mjibson:
A fix for the v3 branch because we haven't upgraded to the v4 API yet and I don't quite have time to do that right now.

jackc:
LGTM. Thanks!

PR #674, opened 2020-02-03
Implements MinConns, the minimum size of the connection pool

ellulpatrick:
The health check will increase the number of connections to this amount if it had dropped below.


jackc:
Thanks!

PR #678, opened 2020-02-10
add custom queries to be used with CopyFrom

nl5887:
This pr will enable the use of CopyFrom with custom queries. This enables to define behaviour on conflicts.

```go
copyCount, err := conn.CopyFrom(context.Background(), pgx.Identifier{"foo"}, []string{"a", "b"}, pgx.CopyFromRows(append(inputRows, duplicateRows...)), pgx.WithQueryTemplate(`CREATE TEMP TABLE tmp_table
		ON COMMIT DROP
		AS
		SELECT *
		FROM {{ .TableName }}
		WITH NO DATA;
		COPY tmp_table ({{ .ColumnNames }}) FROM stdin binary;
		INSERT INTO {{ .TableName }}
			SELECT * FROM tmp_table
		ON CONFLICT DO NOTHING;`))
```

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -5,6 +5,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"text/template"
+
+	"strings"
 
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgio"
@@ -56,6 +59,7 @@ type copyFrom struct {
 	columnNames   []string
 	rowSrc        CopyFromSource
 	readerErrChan chan error
+	queryTemplate string
 }
 
 func (ct *copyFrom) run(ctx context.Context) (int64, error) {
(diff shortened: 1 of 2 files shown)


jackc:
I'm confused about how this works. The server receives the SQL statements after the copy statement before the the data is copied. Also, if a statement after the copy fails what happens. Evidently, according to the tests, it works... but I find it very difficult to reason about...

Does this enable anything that cannot be done already with separate statements? e.g.

```go
tx, _ := conn.Begin()
tx.Exec(ctx, `create temporary table...`)
tx.CopyFrom(ctx, ...)
tx.Exec(ctx, `insert ... on conflict do nothing`)
tx.Commit(ctx)
```

Also, CI found a data race, but I ran the tests 100's of times locally with the race detector and couldn't reproduce. I *think* I see a preexisting issue but it is really tricky to be sure without the ability to reproduce it.

jackc:
You can disregard the data race -- it was a preexisting bug. I was able to build a reproducible test case and fix it.

bfontaine:
Anything new here?

PR #681, opened 2020-02-19
Add query logging for individual batch items

### LINKED ISSUE #680: How to log batched SQL queries?
Hi there!

I'm working with logging at the moment, and noticing that in the v4 `Logger` interface, I see SQL text in the `data` map under the "sql" key.  For batched queries, unfortunately, it seems that this field is always the literal value `"batch query"`.

Is there a way, in pgx, to log the individual statements that make up a batch?

My log configuration code, in essence:

```go
func (la *LogAdapter) Log(ctx context.Context, level pgx.LogLevel, msg string, data map[string]interface{}) {
  if sql, ok := data["sql"]; ok {
    la.logger.Infof("Executing SQL: %s", sql)
  }
}

config, err := pgxpool.ParseConfig(...)
// error handling omitted
config.ConnConfig.Logger = &LogAdapter{...}
```

benjamin-bader:
Following the suggestion in #680, this PR adds the source `*Batch` to the `batchResult` struct, along with an index.  As each query's results are fetched, via `Exec` or `Query`, we'll use those new fields to find the original source query and log it.

A common use of a `BatchResult` is to simply `Close()` it, which will cycle through all remaining results.  We'll log each "unhandled" query in `Close()` to make sure they're recorded.  This is decidedly suboptimal, since we won't be able to associate an error with the specific query that caused it - at least, not in these log statements.  Would love to hear a better approach here.

Fixes #680

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -47,6 +47,8 @@ type batchResults struct {
 	conn *Conn
 	mrr  *pgconn.MultiResultReader
 	err  error
+	b    *Batch
+	ix   int
 }
 
 // Exec reads the results from the next query in the batch as if the query has been sent with Exec.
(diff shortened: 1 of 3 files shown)


jackc:
Thanks!

PR #682, opened 2020-02-21
Add Len() method to *pgx.Batch

WGH-:
This makes the API slightly easier to use when number of calls to
Queue() cannot be trivially computed.

For example, if the program contains the loop like the following,
a separate variable counting the iterations is needed:

    numHeaders := 0
    for _, header := range prepareHeadersForInsert(*res.Headers) {
        headerBatch.Queue("INSERT ...", ...)
        numHeaders++
    }

    headerBatchResult := tx.SendBatch(ctx, headerBatch)

    for i := 0; i < numHeaders; i++ {
        _, err := headerBatchResult.Exec()
        // ...
    }

With method Len(), this extra variable can be eliminated.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -26,6 +26,11 @@ func (b *Batch) Queue(query string, arguments ...interface{}) {
 	})
 }
 
+// Len returns number of queries that have been queued so far.
+func (b *Batch) Len() int {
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

Issue #416, opened 2018-05-07
No error when user does not have rights to tables


NathanZarro:
When the username being used for connection has not been granted the appropriate privileges to the tables, no data is returned, but no error message.

jackc:
I am unable to duplicate this.

```go
package main

import (
	"log"

	"github.com/jackc/pgx"
)

func main() {
	config, err := pgx.ParseEnvLibpq()
	if err != nil {
		log.Fatal(err)
	}

	conn, err := pgx.Connect(config)
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close()

	var s string
	err = conn.QueryRow("select id from customers").Scan(&s)
	if err != nil {
		log.Fatal(err)
	}
}
```

Output:

```
jack@happy:~/dev/go/src/github.com/jackc/pgx-416$ PGDATABASE=xxxx PGHOST=127.0.0.1 PGUSER=xxxx PGPASSWORD=xxxx go run main.go 
2018/05/10 17:56:09 ERROR: permission denied for relation customers (SQLSTATE 42501)
exit status 1
```

vkonstantin:
I have the same bug when a using conn.Query(...), when I am using QueryRow or Exec - it works fine.

jackc:
conn.Query sends the query but doesn't immediately read the results like QueryRow and Exec. The error is returned as part of the results. If you try to use the results of conn.Query() you will get an error. 

```go
package main

import (
	"log"

	"github.com/jackc/pgx"
)

func main() {
	config, err := pgx.ParseEnvLibpq()
	if err != nil {
		log.Fatal(err)
	}

	conn, err := pgx.Connect(config)
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close()

	rows, err := conn.Query("select id from users")
	if err != nil {
		log.Fatalf("Query err: %v", err)
	}

	for rows.Next() {
		var s string
		err = rows.Scan(&s)
		if err != nil {
			log.Fatalf("Scan err: %v", err)
		}
	}

	if rows.Err() != nil {
		log.Fatalf("rows.Err(): %v", rows.Err())
	}
}
```

Output:

```
jack@happy:~/dev/go/src/github.com/jackc/pgx-416$ PGDATABASE=xxxx PGHOST=127.0.0.1 PGUSER=xxxx PGPASSWORD=xxxx go run main.go
2018/10/19 10:29:47 rows.Err(): ERROR: permission denied for relation users (SQLSTATE 42501)
[...]

PR #692, opened 2020-03-17
bail early if preloading rows.Next() results in rows.Err()

basvanbeek:
This changes behavior a bit to be more inline with other stdlib database/sql drivers.

When a Query results in rows.Err() (eg. permission failures) the query succeeds but ends with no rows. One must check with rows.Err() after a results.Next() loop.

Most people don't expect this and for instance `lib/pq` returns permission errors straight from the query.

With this minor adjustment Query will behave similar to `lib/pq` and more importantly people will get the error early (right from the Query return params). Since the package already preloads the first row we can do this easily.

This solves the misunderstanding as identified in https://github.com/jackc/pgx/issues/416


jackc:
Thanks!

PR #693, opened 2020-03-19
remove dynamic MaxConns

kotyara1005:
Hello!
I would like to push changes here. 
I find this behavior really unobvious.
There is nothing about it in docs and it could be dangerous when you run your app in a cluster.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -2,7 +2,6 @@ package pgxpool
 
 import (
 	"context"
-	"runtime"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jackc:
I don't know...

For a single application server you shouldn't have to configure the pool size. The current logic provides a good default in this case.

The cluster case could be a problem -- but there is no good default for that -- if I were running a cluster I would expect to have to manually set the pool size of each app instance.

redbaron:
Single CPU instance is absolutely capable utilising more than 1 PG connection. Default of 4 does no harm IMHO , if anything it should probably be number of GOMAXPROCS * N where N is low single digit.

Issue #703, opened 2020-03-30
Make context available in stdlib.wrapTx.Commit() and Rollback()


georgysavva:
Since `pgx` implements timeouts via context, absence of the context in `stdlib.wrapTx.Commit()` and `Rollback()` leaves us with no option to timeout those calls. And they can block indefinitely. (it's all only applicable if you use pgx via `sql` library).  

Standart `sql` library doesn't pass context to `driver.Tx.Commit()` and `Rollback()` methods. And you are using `context.Background()` in transaction wrapper: [here](https://github.com/jackc/pgx/blob/0446492fe9dc68afff5926b15ec74412ac9c390b/stdlib/sql.go#L544-L546)
But you have access to the context when `sql` library calls `driver.Conn.BeginTx(ctx, ...)` so you can include it in the `wrapTx` structure when you build it: [here](https://github.com/jackc/pgx/blob/0446492fe9dc68afff5926b15ec74412ac9c390b/stdlib/sql.go#L269)
I think it's okay to put context inside a transaction because it has a request bounded live cycle.
Also, standart `sql` library includes context in its own `sql.Tx` transaction implementation: [here](https://github.com/golang/go/blob/3ee782b11da3fb0313603ad0e3be8ab6755802a9/src/database/sql/sql.go#L2028)

Am I missing something or it would be a good thing to improve?

jackc:
Looks good to me.

jackc:
> Looks good to me.

It just occurred to me that could be read two ways. What I mean is 👍 on the suggested change.

georgysavva:
Neat! But I still managed to read two ways)) Does it mean that you prefer me to submit a PR or you will do it yourself?

jackc:
A PR would be great.

georgysavva:
Okay 👍

PR #706, opened 2020-04-02
add context to `stdlib.wrapTx{}`

georgysavva:
Include context into `stdlib.wrapTx{}` in order to propagate it to the underlying `pgx.Tx.Commit()` and `Rollback()` methods. This solves https://github.com/jackc/pgx/issues/703 issue.


jackc:
Thanks!

PR #709, opened 2020-04-03
Add ability to skip network operations on pool initialization.

georgysavva:
I have a proposal:
Add ability to skip network operation on pool initialization, This will allow to decouple your service from the database on startup. This will improve resilience of the service, because most likely you don't actually require the database on startup, and some transient network error can fail the whole application to start.  
Few examples:

1. `database/sql` with `pgx` as a driver, doesn't do any I/O operation on `sql.Open()` (I tested it). 
2. Official `mongoDB` driver also doesn't do any I/O operation on `Client.Connect()`, here are the [docs](https://godoc.org/go.mongodb.org/mongo-driver/mongo#Connect).

I think it will be a good feature for microservices environment. It should be opt-in, to not break expectations of existing users, but it could become the default behaviour in the next major version. What do you think?


jackc:
Thanks!

PR #715, opened 2020-04-10
Document row and composite types handling

redbaron:
It took me a while to figure it out, maybe useful for others


jackc:
Thanks!

PR #716, opened 2020-04-10
Clarify row type handling example

redbaron:


redbaron:
I just realised, that Record.DecodeBinary allocates fields as it sees them in the row. This makes error handling to verbose when it is used as part of another DecodeBinary.

What do you think if Record.DecodeBinary will reuse `Record.Fields` slice if they are not nil? That slice is then used to set expectations of which types can be seen in which positions,as well as number of fields in a row. If there is a mismatch it will trigger a nice and convenient decoding error

jackc:
> What do you think if Record.DecodeBinary will reuse Record.Fields slice if they are not nil? That slice is then used to set expectations of which types can be seen in which positions,as well as number of fields in a row. If there is a mismatch it will trigger a nice and convenient decoding error

While that may work if using a `pgtype.Record` internally as part of another type that would
probably cause problems with the `pgtype.Record` that is reused in the `pgtype.ConnInfo`.

But it might be interesting as a new type that is similar to `pgtype.Record`.

Issue #732, opened 2020-04-29
stdlib: improve context timeouts


georgysavva:
I was reviewing the `database/sql` package and how it integrates with third-party drivers. And I found 3 flows, I opened an issue https://github.com/golang/go/issues/38185, and they confirmed all of them:
- They call `Connector.Connect()` from the background goroutine and the context that this function receive is a background one and it can't have any timeout set by the user and driver might block, since most likely it will do write/read operations aside from the dialling.
- They call `Conn.Close()` without a context and driver might block on this call.
- They call `Stmt.Close()` without a context and driver might block on this call.

As resolution on the issue, they said that it's responsibility of the driver to set reasonable timeouts on all those operations and they will update the docs to reflect this: [PR](https://go-review.googlesource.com/c/go/+/230438/).

I think we should make some changes in the `stdlib` package to implement those expectations.
For `Conn.Close()` and `Stmt.Close` it's quite simple, just add some default timeout to the background context (e.g. 5s-10s):
https://github.com/jackc/pgx/blob/63428b802dc842e375a64adc9d9d85700984c1b7/stdlib/sql.go#L227-L229
https://github.com/jackc/pgx/blob/63428b802dc842e375a64adc9d9d85700984c1b7/stdlib/sql.go#L327-L329

For `Connector.Connect()` it's a bit trickier. I would use the `connect_timeout` setting. Right now it restricts only tcp dialling part, but `pgx` does some internal write/read operations right after establishing the tcp connection and the connection isn't ready until all those operations succeed and they might block. I think they should be considered as a part of the connection process and should be covered by the `connection_timeout`. This way, if user sets the `connect_timeout` setting, `pgx` will set it into context on the `pgx.Connect()` call and ensure that the whole call won't block for more than `connection_timeout`, including all write/read operations inside the `.Connect()` call. 
Alternatively we could do it on the `stdlib` driver level and don't bring this logic to the core `pgx` library.
https://github.com/jackc/pgx/blob/63428b802dc842e375a64adc9d9d85700984c1b7/stdlib/sql.go#L162-L184
What do you think about those changes? If you agree I can submit a PR.

jackc:
Sounds good to me.

PR #734, opened 2020-05-02
Improve Stdlib context timeouts

georgysavva:
As discussed in https://github.com/jackc/pgx/issues/732, adding a default context timeout for stdlib Conn.Close() and Stmt.Close(). 
I also ran `go mod tidy` and included the diff.


jackc:
Thanks!

Issue #742, opened 2020-05-10
Save connString into pgx.Conn, pgxpool.Pool.


georgysavva:
I think it can be helpful to store the original connection string into `pgx.Conn` and `pgxpool.Pool`. It makes sense, since the only valid way to get those objects is to parse a connection string.
What do you think? I Could submit a PR if you agree with that.

jackc:
Seems reasonable.

PR #751, opened 2020-05-16
Add original connection string to Pool and Conn struct

georgysavva:
Now we can access connection string that was used to initialize pool or connection as it was requested in https://github.com/jackc/pgx/issues/742

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -30,7 +30,7 @@ type ConnConfig struct {
 	LogLevel LogLevel
 
 	// Original connection string that was parsed into config.
-	ConnStr string
+	ConnString string
 
 	// BuildStatementCache creates the stmtcache.Cache implementation for connections created with this config. Set
 	// to nil to disable automatic prepared statements.
@@ -160,7 +160,7 @@ func ParseConfig(connString string) (*ConnConfig, error) {
 		createdByParseConfig: true,
 		LogLevel:             LogLevelInfo,
 		BuildStatementCache:  buildStatementCache,
-		ConnStr:              connString,
+		ConnString:           connString,
 	}
 
 	return connConfig, nil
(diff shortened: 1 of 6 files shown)


jackc:
Looks good, but I have a couple suggestions.

The `ConnStr` field on the `Config` struct should never be set by application code. Perhaps it should be private and a method should be provided to access it.

Would it make more sense to expose the config object from `Conn` and `Pool` rather than the conn string? The string could still be accessed through the config. Otherwise an incomplete picture would be exposed if `ParseConfig()` was used and the config was modified from there before using `ConnectConfig()`.

georgysavva:
Okay, sounds good.

jackc:
I merged it with the subtraction of `Conn.ConnString()` and `Pool.ConnString()` I think it is important to require access through the `Config` struct.

A `Config` object may be created via `ParseConfig` and then further
modified. Requiring access to the original `ConnString` via the Config
helps indicate that the `Config` is the source of truth as to how the
connection was actually established. That is, `ConnString` should be considered purely informational. If it is desired to create another connection with the same settings `ConnectConfig()` with the `Config` struct should be used, not `Connect()` with the `ConnString()`.

georgysavva:
My original thought was that user can initalize a pool or connection via just Connect() and such thing as Config stays hidden from him and it would be more intuitive to access ConnString() directly from the Poll or Conn objects. But your concern is more important and I agree, that overall it's better to access ConnString() only from Config. Thanks!

georgysavva:
I think I made a mistake. Right now `pgx.Conn` doesn't store the original config, it stores a copy that is modified in the `pgx.connect()` function:
https://github.com/jackc/pgx/blob/b19811bbebf7e1211acaa4f804ec12b6cd859285/conn.go#L178-L195
If user has a `pgx.Conn` (conn1) and later he wants to establish a new `pgx.Conn` (conn2) using config from the conn1.
`conn1.Config()` will return config with `.OnNotification` set and it won't be reset for the conn2 in `pgx.connect()` function. This can lead to unexpected behaviour. `pgx.Conn` should store the original config before copying and patching it.
BTW this is not a problem for the `Pool` because it saves config before `pgx.connect()` will copy and patch it.

georgysavva:
Apart from that, It would be better if `Conn` and `Pool` had something like `.ConfigCopy()` method that returns a smart copy of the config. 
Let's Imagine the situation that you have a pool and you want to establish a new one with the same config, but to a different logical database. You call `pool.Config()` and since it's a pointer to the original config you don't want to modify it directly and try to make a copy of it:
```golang
newConfig := &pgxpool.Config{}
*newConfig = *pool.Config()
newConfig.ConnConfig.Database = "another_db"
```
But this won't copy the nested `ConnConfig` struct since it's also a pointer and you will still end up modifying the original config. 
Deep copy libraries that do it via reflection are not an option too, since they don't copy private fields and it means that `.createdByParseConfig` will be set to `false` for the new config.
I think the possibility to get a copy of the config from `Conn` and `Pool` should be builtin in the library.

PR #757, opened 2020-05-25
Permit SendBatch with Simple Protocol

### LINKED ISSUE #756: Receiving multiple result sets via the simple protocol
Thanks for your great work on pgx. We're trying to switch to using it more in some client code for https://github.com/cockroachdb/cockroach, as we like the way its Postgres-protocol abstractions are built.

There's one issue I'm running into, which is that it doesn't seem currently possible to run a query via the Simple protocol that returns multiple result sets. Am I missing something? If I look at the code in `Conn.Query`, there's a special case for the simple protocol that's run, but it doesn't seem to pay attention if there were multiple result sets returned by the server.

The `connRows` that's returned from `Conn.Query` in fact has a handle on the `multiResultReader` that could be queried to get the remaining result sets from the server, but as far as I can tell the `Query` abstraction isn't designed for multiple result sets and so that `multiResultReader` isn't really used in this context.

On the other hand, if we use the `ExecBatch` method on `Conn`, there's only the option of running via the extended protocol.

The difficulty is that the Simple protocol actually does let you send a batch in a single query, and it returns multiple result sets - and some things that we've made depend on this functionality. Ideally, the `ExecBatch` interface could allow you to pass in a single, compound query, and respect the `PreferSimpleProtocol` setting if you've asked for it. The reason that this is hard to work around is that in order to use the extended protocol in this case, we'd have to parse the query and split it up into component semicolon-separated queries on the client, which we'd rather not do.

----

Sorry for the long winded intro. I'd like to propose concretely that `ExecBatch` be able to send queries via the simple protocol. Is that something that you're amenable to? I'm happy to contribute a patch.

Cheers,
Jordan

jordanlewis:
Closes #756.

This commit adds support for sending batches of queries via the Simple
protocol with SendBatch. The result appears identically to how it would
if it were created with the extended protocol.

This could probably use some more testing. Pushing up to see what you think, @jackc.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -708,6 +708,29 @@ func (c *Conn) QueryRow(ctx context.Context, sql string, args ...interface{}) Ro
 // explicit transaction control statements are executed. The returned BatchResults must be closed before the connection
 // is used again.
 func (c *Conn) SendBatch(ctx context.Context, b *Batch) BatchResults {
+	simpleProtocol := c.config.PreferSimpleProtocol
+	var sb strings.Builder
+	if simpleProtocol {
+		for i, bi := range b.items {
+			if i > 0 {
+				sb.WriteByte(';')
+			}
+			sql, err := c.sanitizeForSimpleQuery(bi.query, bi.arguments...)
+			if err != nil {
+				return &batchResults{ctx: ctx, conn: c, err: err}
+			}
+			sb.WriteString(sql)
+		}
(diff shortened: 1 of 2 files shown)


jackc:
I tweaked the test to remove a race condition between the context cancellation and the closing of the connection. This was what caused CI to fail. I also moved the test next to the other batch tests in batch_test.go. But other than that it looks good to me.

jordanlewis:
Thanks!

PR #776, opened 2020-06-18
Fix 490: Add jsonb array support for pgx v3

### LINKED ISSUE #490: how do I write JSONB[] using the stdlib adapter?
I'm using the stdlib adapter.

How am I supposed to insert JSONB[]?
I tried serializing it myself, but while I was able to insert that output from the postgres instance, I got an error while inserting that same value from golang/pgx (`Array value must start with \"{\" or dimension information.`). I imagine this might be a double quoting issue because the value I logged was the value I passed to the sql driver, not necessarily what went over the wire after the sql driver took over.
My next thought was to try to pass a slice of map[string]interface, because if the library is doing the marshalling, then map[string]interface is probably what it wants for JSON. That didn't work because `converting argument $2 type: unsupported type []interface {}, a slice of interface`.

For postgres beginners like myself there are not a lot of hints in this repo so I'm just looking for guidance here - it feels like I'm missing something obvious.

A note about my use case: This is for replicating data, and as such the shape of the data is dynamic, so it is not desirable for me to define json-tagged structs like most of the examples use. Using the standard driver was ideal because the same program also replicates data to mysql, and so a minor amount of code reuse is possible if I can continue using the standard driver for the postgres implementation.

laskoviymishka:
Fix #490

### DIFF
--- a/pgtype/jsonb_array.go
+++ b/pgtype/jsonb_array.go
@@ -1,0 +1,300 @@
+package pgtype
+
+import (
+	"database/sql/driver"
+	"encoding/binary"
+
+	"github.com/jackc/pgx/pgio"
+	"github.com/pkg/errors"
+)
+
+type JSONBArray struct {
+	Elements   []JSONB
+	Dimensions []ArrayDimension
+	Status     Status
+}
+
+func (dst *JSONBArray) Set(src interface{}) error {
+	// untyped nil and typed nil interfaces are different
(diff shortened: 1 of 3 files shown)


jackc:
Can this be generated with `typed_range.go.erb`?

laskoviymishka:
@jackc could you plz review, also fix pg element type from text to JSONB.

PR #784, opened 2020-06-28
Cache column names in stdlib Rows

zikaeroh:
A recent PR to pgconn fixed a bug where the field descriptions were cleared when they shouldn't be. That had a knockon effect of making `Columns` do more work here (since its data wasn't being deleted), but caused a performance impact. See jackc/pgconn#45.

This PR caches the column names, so long as the field descriptions do not change. I'm uncertain if they ever will in the current code, given this driver does not implement `driver.RowsNextResultSet`, but I've attempted to handle it anyway.

Note that the way I'm checking that the field descriptions do not change relies on the fact that `pgconn` will never modify its field description slices and instead make new ones. If you do not believe this to be a good thing to rely on, then I can do a simple equality check against a single string slice, as code that looks like this does not allocate either:

```go
func BytesEqualString(b []byte, s string) bool {
    return string(b) == s  // Compiler proves that b is not modified, so does not allocate.
}
```

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -412,15 +412,20 @@ type Rows struct {
 	valueFuncs   []rowValueFunc
 	skipNext     bool
 	skipNextMore bool
+
+	columnNames []string
 }
 
 func (r *Rows) Columns() []string {
-	fieldDescriptions := r.rows.FieldDescriptions()
-	names := make([]string, 0, len(fieldDescriptions))
-	for _, fd := range fieldDescriptions {
-		names = append(names, string(fd.Name))
+	if r.columnNames == nil {
+		fields := r.rows.FieldDescriptions()
+		r.columnNames = make([]string, len(fields))
+		for i, fd := range fields {
+			r.columnNames[i] = string(fd.Name)
+		}
 	}
-	return names
+
+	return r.columnNames
 }
 
 // ColumnTypeDatabaseTypeName returns the database system type name. If the name is unknown the OID is returned.


zikaeroh:
Perhaps a better fix would be in `pgconn` to expose the names directly, as `pgconn` will know exactly when `FieldDescriptions()` has changed. Up to you; that's an API change. I just know that from reading the Go `database/sql` package, it assumes that it can call `Columns()` whenever it wants, often doing things like `len(rows.Columns())` or `rows.Columns()[i]`, hence why this path is so hot.

zikaeroh:
It turns out that when I was testing this, I had also reverted jackc/pgconn#36 so wasn't seeing the true numbers, but this PR does at least recover some of the lost performance.

zikaeroh:
Sorry send yet another message, I looked into how to regain even more of the lost performance, and found most of the allocations come from the `[]byte` to string conversion.

Doing:

```go
import "github.com/josharian/intern"

// ...

	names := make([]string, len(fields))
	for i, fd := range fields {
		names[i] = intern.Bytes(fd.Name)
	}
```

Helps tremendously, but allowing `pgconn` to explicitly create its strings and allocate them lazily is definitely a more proper solution.

jackc:
This is an interesting performance case. I rarely need to examine column names, so it never occurred to me that would be a performance issue.

As you mentioned, this driver does not implement `driver.RowsNextResultSet` however it still may not be safe to rely on `FieldDescriptions` not changing. `FieldDescriptions` is shared through chain of flyweight objects ultimately leading to one `pgproto3.RowDescription` that is shared for an entire connection.

Beyond that, I suggest a simpler cache approach of only a `[]string`. `nil` meaning it hasn't been cached yet. If `driver.RowsNextResultSet` is ever implemented in the future then `NextResultSet()` could nillify that cached field.

zikaeroh:
I've simplified this down to just a plain slice as requested.

The interning thing might still help, but I can send another PR about that if I think it's worth it (or if I find a different method to avoid allocations).

jackc:
Thanks!

PR #816, opened 2020-08-31
Fix SimpleProtocol with prepared statement

jinzhu:
Hello,

When I am testing the SimpleProtocol with prepared statement, the `pgx` driver returns error `unused argument: 0` which from https://github.com/jackc/pgx/blob/43ce317556c724144d1a44bc8d176c2882ad6ebf/internal/sanitize/sanitize.go#L236

The SQL here is something like `pgx_0` because it is a prepared statement, so it can't pass the `Sanitizer` 

The issue should be fixed after this change, thank you.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -473,12 +473,12 @@ optionLoop:
 		}
 	}
 
-	if simpleProtocol {
-		return c.execSimpleProtocol(ctx, sql, arguments)
-	}
-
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

frederikhors:
Is this released in `v4.9.0`?

jackc:
> Is this released in v4.9.0?

Yes.

PR #842, opened 2020-10-05
use simpler strings funcs

cristaloleg:
`strings.ReplaceAll` was added in Go 1.12 (or 1.11 ?) so it's safe to migrate, `pgx` supports only 2 latest releases, 1.14 and 1.15 currently.

Found by `gocritic`
```
conn.go:95:8: wrapperFunc: use strings.ReplaceAll method in `strings.Replace(ident[i], string([]byte{0}), "", -1)` (gocritic)
conn.go:96:20: wrapperFunc: use strings.ReplaceAll method in `strings.Replace(s, `"`, `""`, -1)` (gocritic)
conn.go:393:15: wrapperFunc: use strings.ReplaceAll method in `strings.Replace(s, `"`, `""`, -1)` (gocritic)
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -92,8 +92,8 @@ type Identifier []string
 func (ident Identifier) Sanitize() string {
 	parts := make([]string, len(ident))
 	for i := range ident {
-		s := strings.Replace(ident[i], string([]byte{0}), "", -1)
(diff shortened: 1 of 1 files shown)


jackc:
👍

PR #843, opened 2020-10-05
remove unused definitions on messages.go

cristaloleg:
Looks like it was added and not used for a long time, safe to delete.

```
messages.go:11:2: `copyData` is unused (deadcode)
messages.go:12:2: `copyFail` is unused (deadcode)
messages.go:13:2: `copyDone` is unused (deadcode)
messages.go:33:6: `appendQuery` is unused (deadcode)
```

By the way, those consts aren't used too https://github.com/jackc/pgx/blob/master/conn.go#L18 and one func https://github.com/jackc/pgx/blob/master/conn.go#L401 Have decided to ask first and then remove, if you're ok with it.

```
conn.go:19:2: `connStatusUninitialized` is unused (deadcode)
conn.go:20:2: `connStatusClosed` is unused (deadcode)
conn.go:21:2: `connStatusIdle` is unused (deadcode)
conn.go:22:2: `connStatusBusy` is unused (deadcode)
conn.go:384:6: `connInfoFromRows` is unused (deadcode)

conn.go:313:16: func `(*Conn).processContextFreeMsg` is unused (unused)
conn.go:322:16: func `(*Conn).rxErrorResponse` is unused (unused)
```

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -3,14 +3,7 @@ package pgx
 import (
 	"database/sql/driver"
 
-	"github.com/jackc/pgio"
 	"github.com/jackc/pgtype"
-)
-
-const (
-	copyData = 'd'
-	copyFail = 'f'
-	copyDone = 'c'
 )
 
 func convertDriverValuers(args []interface{}) ([]interface{}, error) {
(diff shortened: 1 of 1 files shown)


jackc:
Good catch. I think all this is from before the lower level parts of pgx were extracted to pgconn.

cristaloleg:
Thanks, but what about 2nd part of the PR message? Should I remove it too?

jackc:
Yes, I think those code be removed as well.

PR #844, opened 2020-10-06
Add/Copy docstring from puddle

drPytho:
Add docstring for stat, transferred from Puddle with some naming updates where suitable.


jackc:
Thanks!

Issue #845, opened 2020-10-07
Memory Leak on ChunkReader.newBuf


EricByers:
We have been seeing a memory leak that shows up over time, that we can't seem to get to the bottom of, specifically when in a transaction, and we do a SELECT during that query.  This process is being run in parallel ~100 at a time, and will develop a significant leak after 5-10 minutes of running, that never gets cleared up.

The code that primarily shows the memory being retained (when reallocating the chunk reader buffer location).  https://github.com/jackc/chunkreader/blob/3c4a99247c6656284dd2be8d74438ca9bbb4f342/chunkreader.go#L72-L76

pgx portion of the trace:
<img width="590" alt="discover-api_inuse_space" src="https://user-images.githubusercontent.com/403278/95395369-0ecf2200-08c4-11eb-8b19-10fd72e04091.png">

Any insights that could help us narrow down what may be retaining the reference?

jackc:
pgx never reuses the buffers allocated by chunk reader. This allows it to transfer ownership of memory instead of incurring the cost of an allocation and a copy. e.g. When scanning into a `[]byte` no new memory is allocated and no copy occurs. The `[]byte` is simply updated to point at the relevant part of what was read directly from the network connection.

However, this does have one potential downside -- which I'm guessing is the problem you are experiencing. Any scanned `[]byte` will prevent the entire buffer (8KiB by default) from being garbage collected. This is rarely a problem as query results are usually short lived. I've previously considered this a theoretical issue, this is the first I've heard of someone experiencing it.

So that's what I would check for first -- long lived `[]byte` that were scanned.

tugberkugurlu:
@jackc 👋  after recently trying to upgrade from 3.6.2 to 4.6.0, I observed our containers crashing due to running out of memory. After investigating this further by running a data access function with a benchmark as below: 

```
go test -bench=. -benchtime=60s -run=^$ . -memprofile memprofile.out
```

I am observing exactly the same issue as @EricByers's is experiencing. Please see the following pgx portion of the memory profile, which is pointing to [the same place](https://github.com/jackc/chunkreader/blob/3c4a99247c6656284dd2be8d74438ca9bbb4f342/chunkreader.go#L72-L76) as @EricByers has indicated.

![image](https://user-images.githubusercontent.com/328122/95686405-83bc9780-0bf5-11eb-8ae7-f5670aac3f93.png)

> So that's what I would check for first -- long lived []byte that were scanned.

[...]

PR #849, opened 2020-10-14
Add MaxConnLifetime config setting to pool in v3

tugberkugurlu:
In our usage, we want to enforce our connections to be recycled so that the application can pick up the newly added replicas in ur aurora cluster.

In light of the issue we are experiencing with https://github.com/jackc/pgx/issues/845,
and the fact that we want to have the ability to recycle our connections,
this patch proposes to add the `MaxConnLifetime` config value and its behaviour to v3 releases.

The work done in this PR is trying to mimic what has done for v4 track, but w/o the idle connection check since that has turned out be an harder port than the simple implementation of just checking the conn lifetime after its usage during `Release` phase: https://github.com/jackc/pgx/commit/c604afba82679df5ca91fbd6da922fdd205d4310

The the other difference is that this one doesn't have a default maxConnLifetime value defined,
which means that the current v3 usage out there shouldn't be impacted by this change.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -224,6 +224,7 @@ func (cc *ConnConfig) networkAddresses() ([]net.Addr, error) {
 // goroutines.
 type Conn struct {
 	conn               net.Conn  // the underlying TCP or unix domain socket connection
+	creationTime       time.Time // the time indicating when the connection is established
 	lastActivityTime   time.Time // the last time the connection was used
 	wbuf               []byte
 	pid                uint32            // backend pid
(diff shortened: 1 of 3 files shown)


tugberkugurlu:
@jackc 👋  is there a way to kick off the CI server to run the tests for `tugberkugurlu:tu/add-maxconnlifetime` branch?

jackc:
I restarted the CI but it timed out. I ran the tests locally and something seems to hang there as well.

I don't see the reason why but it appears that it is hanging in `TestPoolWithAcquireTimeoutSet`.

rubiagatra:
Do you know if the PR is still relevant? @tugberkugurlu since it has https://pkg.go.dev/github.com/jackc/pgx/v5/pgxpool#Config 

Can we close this PR? @jackc

PR #860, opened 2020-11-02
[READY FOR REVIEW] [DO NOT MERGE] fix prepared statement cache invalidation

### LINKED ISSUE #841: connection local statement cache does not correctly invalidate
# Description of Issue

`pgx` appears to maintain a [connection-local cache](https://github.com/jackc/pgx/blob/2a8ae6f269e17c44d25cd0f713eaa6356547206e/conn.go#L653) of prepared statements that are automatically used for queries performed by `sql.DB.QueryContext`.[1] This cache does not appear to handle the case when the underlying database schema has changed, invalidating the prepared statement.

At first blush, the database schema changing out from under a prepared statement seems pretty weird, so I'll explain how we've seen it being an issue in practice. When doing a deploy of an existing service, there is often a database migration along with a code change that accounts for the new schema. If the migration runs first, the old application servers will still be running with the old code and caches hanging around. Before the servers can be gracefully restarted with new code, the old code will begin experiencing issues because of the cached prepared statements.

# Minimum Replication

The following program demonstrates the issue:

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"log"

	_ "github.com/jackc/pgx/v4/stdlib"
)

// run `createdb pgx_mucking`
func main() {
	ctx := context.Background()

	db, err := sql.Open("pgx", "postgres://localhost/pgx_mucking?sslmode=disable")
	if err != nil {
		os.Exit(1)
	}
	defer db.Close()

	// Grab a single connection on which to perform the SELECT statements so we can be
	// sure to reproduce the bug in this toy example. The bug will be intermittent if we
	// instead just call the QueryContext method on the `*sql.DB`.
	conn, err := db.Conn(ctx)
	if err != nil {
		log.Fatal(err)
	}

	_, err = db.ExecContext(ctx, `
        DROP TABLE IF EXISTS drop_cols;
        CREATE TABLE drop_cols (
            id SERIAL PRIMARY KEY NOT NULL,
            f1 int NOT NULL,
            f2 int NOT NULL
        );
    `)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("connected to db")

	// insert some data (probably not needed for replication)
	_, err = db.ExecContext(ctx, "INSERT INTO drop_cols (f1, f2) VALUES (1, 2)");
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("inserted values")

	getSQL := "SELECT * FROM drop_cols WHERE id = ANY($1)"

	// populate pgx's internal prepared statement cache
	_, err = conn.QueryContext(ctx, getSQL, 1)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("populated stmtcache")

	// change the schema (in a more realistic example, this would mean running a
	// migration while the application server is still running).
	_, err = db.ExecContext(ctx, "ALTER TABLE drop_cols DROP COLUMN f1")
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("altered schema")

	// run the query again, causing us to use a bad prepared statement
	_, err = conn.QueryContext(ctx, getSQL, 1)
	if err != nil {
		log.Fatalf("dies with: %s", err.Error())
	}
	fmt.Println("never prints")
}
```

The output of that program is as follows:

```
connected to db
inserted values
populated stmtcache
altered schema
2020/10/05 10:55:02 dies with: ERROR: cached plan must not change result type (SQLSTATE 0A000)
```

# Potential Fixes

I can see two possible fixes:
1. When Query is called via the `database/sql` interfaces, never use the cache. This would follow the principle of least surprise in my view, but I can definitely see now wanting to give up any perf wins that caching gives.
2. If there is an error when calling Next to get the first row, attempt to recover by invalidating the cache for that statement and retrying the query. This may be a little annoying because the error shows up in [the first Next call](https://github.com/jackc/pgx/blob/2a8ae6f269e17c44d25cd0f713eaa6356547206e/stdlib/sql.go#L360-L364) rather than the call to actually query the db.

I'm happy to try to put together a patch for either one of these approaches. Just let me know which one you prefer.

[1]: This is somewhat surprising because the `sql.DB` interface contains dedicated interfaces for creating prepared statements, so I would not expect the normal Query interfaces to create prepared statements. That said, if there are big perf wins, I can't see how to realize those for people who want to use `sql.DB` directly rather than layering some sort of framework on top of it, so the choice `pgx` makes seems reasonable in that context.

ethanpailes:
This patch fixes jackc/pgx#841. To fix it we notice
when a statement failed because its cached plan is now
invalid and flush it from our cache. We don't try to get
fancy with retries or anything like that, just return the
error and allow the application to handle it.

The flushing of statements is deferred to the end of transactions
because all statements executed after an error in a transaction
will just result in errors. If we are not in a transaction, we
flush eagerly.

I had to make [some][1] [changes][2] to to the jackc/pgconn package as well
as this package, so I've retargeted this this repo on
ethanpailes/pgconn for the time being. This PR is not really in a
mergable state until those PRs merge and I rip out 2f31e98, but I
would say it is ready for review (just review the most recent commit).

One question I have is whether anything else needs to be done around
batch SQL, so guidance there would be appreciated.

Fixes #841

[1]: https://github.com/jackc/pgconn/pull/54
[2]: https://github.com/jackc/pgconn/pull/55


ethanpailes:
Closing in favor of https://github.com/jackc/pgx/pull/865

PR #870, opened 2020-11-12
Expose ResultReader.Err() to allow row iteration to check for errors

eh-steve:
This PR and the accompanying https://github.com/jackc/pgconn/pull/58 address a problem where `ResultReader` errors due to statement cache errors would be swallowed during row iteration (since `r.NextRow()` would return false immediately, since its command is concluded), making a query apparently execute with no errors, but providing no rows.

This might be closed in favour of https://github.com/jackc/pgx/pull/865, which addresses the same problem differently


ethanpailes:
@eh-steve I think this isn't needed anymore.

jackc:
I think the underlying issues have been resolved. Feel free to reopen if that's not the case.

PR #880, opened 2020-11-24
Enable and disable simple protocol with context

covrom:
When using stdlib, we cannot use parameters for queries in arguments, because the database / sql checks the number of arguments and placeholders.
This MR adds a context option to enable or disable simple protocol in queries.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -39,6 +39,20 @@ type ConnConfig struct {
 	createdByParseConfig bool // Used to enforce created by ParseConfig rule.
 }
 
+type preferSimpleProtocol struct{}
+
+// WithSimpleProtocol creates a context that disables the implicit use of a prepared statement
+// for all requests that use that context
+func WithSimpleProtocol(ctx context.Context) context.Context {
+	return context.WithValue(ctx, preferSimpleProtocol{}, true)
+}
+
+// WithoutSimpleProtocol creates a context that disables the use of simple protocol
+// for all requests using this context
+func WithoutSimpleProtocol(ctx context.Context) context.Context {
+	return context.WithValue(ctx, preferSimpleProtocol{}, false)
+}
+
 // Copy returns a deep copy of the config that is safe to use and modify.
 // The only exception is the tls.Config:
 // according to the tls.Config docs it must not be modified after creation.
@@ -410,6 +424,9 @@ func (c *Conn) Exec(ctx context.Context, sql string, arguments ...interface{}) (
 
 func (c *Conn) exec(ctx context.Context, sql string, arguments ...interface{}) (commandTag pgconn.CommandTag, err error) {
 	simpleProtocol := c.config.PreferSimpleProtocol
+	if spctx, ok := ctx.Value(preferSimpleProtocol{}).(bool); ok {
+		simpleProtocol = spctx
+	}
 
 optionLoop:
 	for len(arguments) > 0 {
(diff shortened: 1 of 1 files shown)


jackc:
> When using stdlib, we cannot use parameters for queries in arguments, because the database / sql checks the number of arguments and placeholders.

This works for me.

```go
package main

import (
	"database/sql"
	"fmt"
	"log"

	"github.com/jackc/pgx/v4"
	_ "github.com/jackc/pgx/v4/stdlib"
)

func main() {
	db, err := sql.Open("pgx", "")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	var s string
	err = db.QueryRow("select 'Hello, world'", pgx.QuerySimpleProtocol(true)).Scan(&s)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(s)
}
```

Output:

```
$ go run main.go
Hello, world
```

covrom:
try to use $N arguments with this

jackc:
> Try to use $N or ? arguments with this, and Query for sql.Rows

Still works.

```go
package main

import (
	"database/sql"
	"fmt"
	"log"

	"github.com/jackc/pgx/v4"
	_ "github.com/jackc/pgx/v4/stdlib"
)

func main() {
	db, err := sql.Open("pgx", "")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	rows, err := db.Query("select $1", pgx.QuerySimpleProtocol(true), "Hello, world")
	if err != nil {
		log.Fatal(err)
	}

	for rows.Next() {
		var s string
		err = rows.Scan(&s)
		if err != nil {
			log.Fatal(err)
		}

		fmt.Println(s)
	}

	if err := rows.Err(); err != nil {
		log.Fatal(err)
	}
}
```

covrom:
OK, now try to prepare stmt and query it.
After this see at https://github.com/golang/go/blob/master/src/database/sql/convert.go#L119.
For existing statements database/sql checks number of placeholders.

jackc:
I don't understand trying to execute a prepared statement over the simple protocol. It's not possible. Prepared statements by definition must use the extended protocol.

covrom:
You right. But, how to skip driver prepared cache and run as extended query without using prepared oid in your driver for some queries?

jackc:
Maybe set the prepared statement cache to describe mode. That saves the extra round trip without actually using a true prepared statement.

covrom:
describe mode disables all prepared queries totally, its not good. we need per-query option for describe mode.

jackc:
> describe mode disables all prepared queries totally

Not entirely. It no longer automatically prepares and caches statements. But you can still use explicit prepared statements.

This is a very unusual use case. What exactly is it that won't work with the simple protocol, and won't work with prepared statements in the extended protocol, but does work with the unnamed prepared statement with the extended protocol -- which only differs from normal prepared statements in that the statement is unnamed and ephemeral.

covrom:
for example, zombodb crashed on some prepared queries

covrom:
> But you can still use explicit prepared statements

When 99% queries is prepared and 1% is unprepared, this way is long. Option in context is goot choice for enable "describe" for some queries.

jackc:
> for example, zombodb with elastic queries crashed on some prepared queries, and works good with extended protocol.

It sounds like that is the root problem that should be fixed. 

> > But you can still use explicit prepared statements

> When 99% queries is prepared and 1% is unprepared, this way is long.

But you're original post apparently had to do with prepared statements... Anyway, use the normal statement cache for the 99% and the simple protocol for the 1%. Simple protocol works fine per query.

Issue #882, opened 2020-11-28
Proposal: Convenience function `CopyFromSlice`


egonelbre:
Currently `CopyFromRows` is rather cumbersome to use and similarly implementing a custom type is often annoying. Usually the values people want to insert are in a slice, so a helper for such situation would be neat.

It could work similarly how some `sort` API-s work. e.g.

```
_, err := conn.CopyFrom(ctx,
	pgx.Identifier{"users"},
	[]string{"name", "age"},
	pgx.CopyFromSlice(len(users), func(i int) []interface{} {
		u := &users[i]
		return []interface{}{u.Name, u.Age}
	}),
)
```

Alternatively, it could be a more efficient version that avoids allocating `[]interface{}`:

```
_, err := conn.CopyFrom(ctx,
	pgx.Identifier{"users"},
	[]string{"name", "age"},
	pgx.CopyFromSlice(len(users), func(e *pgx.RowEncoder) error {
		e.String(u.Name)
		e.Int32(u.Age)
		e.Any(u.Misc)
		return nil
	}),
)
```

jackc:
I think using `CopyFromSource` interface is similar to using the `sort.Sort` interface. But I suppose that your proposal would be analogous to `sort.Slice`.

I'm open to this idea. I prefer the interface of the first one better. I doubt that one allocation per row will make a big difference. In the rare case it does matter there is always still the option of implementing `CopyFromSource`.

egonelbre:
Yes, I meant something like `sort.Slice`. I'll make a PR.

PR #886, opened 2020-12-02
add pgx.CopyFromSlice

egonelbre:
Using CopyFromRows can often be inconvenient to use, because you would
need to convert a typed array to an [][]interface{}. Similarly,
implementing a custom CopyFromSource is too verbose for one-off things.

Add CopyFromSlice that allows to more easily convert a slice to a
CopyFromSource. Example:

    copyCount, err := conn.CopyFrom(
        context.Background(),
        pgx.Identifier{"people"},
        []string{"first_name", "last_name", "age"},
        pgx.CopyFromSlice(len(rows), func(i int) ([]interface{}, error) {
            return []interface{user.FirstName, user.LastName, user.Age}, nil
        }),
    )

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -35,6 +35,31 @@ func (ctr *copyFromRows) Err() error {
 	return nil
 }
 
+// CopyFromSlice returns a CopyFromSource interface over a dynamic func
+// making it usable by *Conn.CopyFrom.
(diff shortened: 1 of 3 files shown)


egonelbre:
Fixes #882

jackc:
Merged. I made one change. The `Err()` method of the `CopyFromSource` interface needs to return any errors that occur. So `Values()` now stores any error that occurs and `Err()` returns it.

PR #896, opened 2020-12-20
Updated CopyFromSlice example in doc.go

musbah:
The example had a syntax error and didn't explain what user was to begin with. Fixed it.


jackc:
Thanks!

PR #901, opened 2021-01-02
Add BeforeConnect callback to pgxpool.Config.

rafroehlich2:
This implements #676 allowing for connection settings to be updated without needing to create a new pool. 

This will only trigger a cloning of the pgx.ConnConfig if the BeforeConnect callback is not nil.


jackc:
Thanks!

PR #919, opened 2021-01-25
Add Ping method to pgxpool.Conn

### LINKED ISSUE #918: Expose Ping method in pgxpool.Conn
Was trying to build a health check for the database connection in an application using `pgxpool`. Noticed there wasn't a `Ping` method exposed.

Having a look through the codebase, seems the addition could be made fairly easily by doing:

```go
func (c *Conn) Ping(ctx context.Context) error {
	return c.Conn().Ping(ctx)
}
```

I'd be happy to raise a PR to implement this if that's deemed the appropriate solution.

davidsbond:
Adds the Ping method to pgxpool.Conn, returning the result of calling Ping on the underlying pgx.Conn.

Closes #918


jackc:
Thanks!

PR #921, opened 2021-01-27
added a kitlog level log adapter

akhenakh:
Example https://gist.github.com/akhenakh/4c16673954cd6a25b8611d4015f3396f


jackc:
Thanks!

Juneezee:
@jackc It is likely that this merge request had added too many transitive dependency into this module. See #977

Issue #931, opened 2021-02-07
Public ScanArgError would help other libraries


pausan:
scanArgError in rows.go is private, and thus, when a column type does not match, it formats and returns an error saying something like:

    "can't scan into dest[3]: whatever reason"

There are other libraries building on top of the amazing pgx, like scany or pgxscan, which introduces some handy features on top of this library. Since these libraries allow to map structs to returned columns, it would be really helpful to have ScanArgError public, so it will save some parsing in order to identify the wrong column and show the column name in the returned error instead. Or something in the line of: 

    "can't scan into dest[3] mapped to field "FieldName": whatever reason"

If this error struct is private, such thing cannot be done in a sort of transparent way. The same error cannot be returned either because the error itself is private.

Thanks!

PR #932, opened 2021-02-07
Make ScanArgError public to allow identification of offending column

pausan:
I'm working on fixing some things on https://github.com/randallmlough/pgxscan library, which builds on top of the amazing pgx. I found the limitation of not being able to identify column name when this error arises, and we can improve error reporting a lot with this feature (I mean, it can be done without this, but requires doing some extra parsing of a string, which hurts performance and is a pity taking into account that the information is already there).

I've created this ticket to report the problem, but since the change is straightforward I made it myself.
https://github.com/jackc/pgx/issues/931

Thanks a lot for considering it.


jackc:
Thanks!

PR #940, opened 2021-02-13
Make ScanArgError fields public

pausan:
Hey! Thanks for merging the previous pull request (https://github.com/jackc/pgx/pull/932). 

Silly me. Totally my fault. I added ScanArgError as public, but not its fields, so now the error can be detected but it is still required to parse the error string to get the column index. My fault. Can you merge again? Thanks a lot for considering this!

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -221,7 +221,7 @@ func (rows *connRows) Scan(dest ...interface{}) error {
 
 		err := rows.scanPlans[i].Scan(ci, fieldDescriptions[i].DataTypeOID, fieldDescriptions[i].Format, values[i], dst)
 		if err != nil {
-			err = ScanArgError{col: i, err: err}
+			err = ScanArgError{ColumnIndex: i, Err: err}
 			rows.fatal(err)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


jackc:
Would it be better to implement `Unwrap` instead of making `err` public?

And if `Col` is going to be public maybe it deserves a better name? e.g. something like `ColumnNumber`, `ColumnIndex`, `ColIdx`, etc. I'd probably go for `ColumnIndex` but I haven't thought much about it.

pausan:
I totally agree with the name, I actually thought about changing name but I was hesitant to change it. ColumnIndex sounds perfect.

Regarding Unwrap I will have to learn a little bit more golang to make sure I do it right, since I'm kind of new to the language, but I'll have a look and do a commit with your suggestions. Thanks!

pausan:
I edited the commit to update the field name and provide a way to Unwrap the error.

After playing a little bit with Unwrapping and reading about the changes made in golang in 1.13 I believe providing an Unwrap method is the cleanest. Let me know if you think otherwise.

Thanks for the feedback so far.

PR #955, opened 2021-03-08
Add logging on CopyFrom

phemmer:
Logging previously only logged statement results on Exec() and Query(), but not CopyFrom(). This change makes it consistent.

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


jackc:
Merged with one change -- message is "CopyFrom" instead of "Copy".

phemmer:
Ah ok, Thanks. I had used "copy" because I thought it was reflecting the postgres operation name, not the pgx function.

PR #960, opened 2021-03-13
Add comment support when sanitizing SQL queries

psyhatter:
There is a potential error when using comments in a SQL query because they are not processed, for example, if there are quotes, double quotes, or a dollar sign inside the comment.
Added support for comments and example tests.


jackc:
Thanks!

PR #965, opened 2021-03-16
Prevent panics caused by attempting to close an already closed pgxpool.Pool.

schultz-is:
# Summary
Attempting to close a `pgxpool.Pool` instance that was previously closed will result in a panic because the `closeChan` member will have already been closed. This simply checks the if the channel is open by attempting to read from it before calling close. One case in which this change will help is when a function contains both a `defer pool.Close()` and a manual closure (however contrived that might be in the real world.)

# Reproduction Steps
1. Connect a pool.
2. Close the pool.
3. Attempt to close the pool again.
4. Observe the panic due to multiple `close` calls against the `closeChan` member.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -326,8 +326,14 @@ func ParseConfig(connString string) (*Config, error) {
 // Close closes all connections in the pool and rejects future Acquire calls. Blocks until all connections are returned
 // to pool and closed.
 func (p *Pool) Close() {
-	close(p.closeChan)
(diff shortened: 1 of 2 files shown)


jackc:
I merged in the test, but I changed the implementation to use a `sync.Once`. Otherwise concurrent `Close` calls would still have a potential to close the channel twice.

schultz-is:
Ah, that’s a great point. Thanks for reviewing the change!

PR #975, opened 2021-03-26
stdlib: implement Conn.ResetSession

### LINKED ISSUE #974: stdlib: DB.Conn returns a closed connection
`database/sql.DB.Conn` should return a valid connection, but sometimes it doesn't. It can return a closed connection. 

Although there are tests in `sql_test.go` that check for this, it isn't correct behavior. `DB.Conn` has its own retry loop, which is supposed to cooperate with the driver to filter out bad connections. The `driver.ErrBadConn` error is meant for that internal communication, and for `DB.Conn` to work it should be returned by the driver's `Conn.ResetSession` method.

jba:
This prevents closed connections from being returned by `database.sql.DB.Conn`.

Fixes #974.

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -385,6 +385,13 @@ func (c *Conn) CheckNamedValue(*driver.NamedValue) error {
 	return nil
 }
 
+func (c *Conn) ResetSession(ctx context.Context) error {
+	if c.conn.IsClosed() {
+		return driver.ErrBadConn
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

stoksc:
any idea when this would get cut into a released version?

jackc:
Just tagged `v4.12.0`.

PR #982, opened 2021-04-02
stdlib: Do not reuse ConnConfig strings

### LINKED ISSUE #947: RegisterConnConfig reuses data source names for different connection options
Repro:
1. Call `RegisterConnConfig`.
2. Call `UnregisterConnConfig` on the result of #1.
3. Call `RegisterConnConfig` again, but with different connection options.

EXPECTED: The connection string returned by #3 should be different than #1.
ACTUAL: #3 reuses the same connection string used by #1.

This would be easy to fix, by using a separately incrementing counter rather than `len(d.configs)` in `Driver.registerConnConfig`. It would then make building tooling easier. In my particular case, I'm doing some pooling of connections in [https://github.com/cockroachdb/copyist](https://github.com/cockroachdb/copyist). That pooling relies on the reasonable assumption that two connections with the same data source name are connecting to the same backend database. This works great with every other driver I've tried, and even works well with `pgx`, unless `RegisterConnConfig` is used.

I'd be happy to create a PR to fix this, if it's something you're open to merging.

andy-kimball:
Previously, stdlib.RegisterConnConfig would sometimes reuse the same connection
string for different ConnConfig options (specifically, it happened when a connection
was open and then closed, and then a new, different connection was opened). This
behavior interferes with callers that expect that two connections with the same data
source name are connecting to the same backend database in the same way.

This fix updates stdlib.RegisterConnConfig to use an incrementing sequence
counter to uniquify all returned connection strings.

Fixes #947

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -169,6 +169,7 @@ func GetDefaultDriver() driver.Driver {
 type Driver struct {
 	configMutex sync.Mutex
 	configs     map[string]*pgx.ConnConfig
+	sequence    int
 }
 
 func (d *Driver) Open(name string) (driver.Conn, error) {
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #1001, opened 2021-05-06
stdlib: add OptionBeforeConnect and randomizer

dankinder:
Fixes https://github.com/jackc/pgconn/issues/71

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -56,6 +56,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


stdlib/sql.go
```diff
@@ -110,18 +111,57 @@ var (
...
+	if len(connConfig.Fallbacks) == 0 {
+		return nil
+	}
+
+	newFallbacks := append([]*pgconn.FallbackConfig(nil),
+		&pgconn.FallbackConfig{
+			Host:      connConfig.Host,
+			Port:      connConfig.Port,
+			TLSConfig: connConfig.TLSConfig,
+		})
```
jackc, line 143: I don't understand why append is used here. Wouldn't a slice literal be simpler?

dankinder: Good catch, fixed with what I originally meant to put there

PR #1002, opened 2021-05-07
Checking that connection was not closed by the server before reusing it from the pool.

damour:
Checking that conn.write() doesn't return an error during query execution is not enough to prevent errors like EOF for closed connections by the server because write() syscall returns success for the socket that received FIN packet (half-closed TCP connection).
EOF error will be returned only after the read() call but in this case, we cannot distinguish half-closed connection(after the previous query) and read timeout for the new query so query retry is not safe and driver propagates the error to the application.
We use a lightweight non-blocking read from the socket that immediately returns the result in driver.ResetSession() method before a connection is reused for another query so a new connection will be taken from the pool if server initiated connection closing and application can work smoothly.

Implementation was taken from [go mysql driver](https://github.com/go-sql-driver/mysql/pull/934/files).

https://github.blog/2020-05-20-three-bugs-in-the-go-mysql-driver/
https://github.com/golang/go/issues/15735#issuecomment-481519888

### DIFF
--- a/stdlib/conncheck.go
+++ b/stdlib/conncheck.go
@@ -1,0 +1,49 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// +build !windows
+
+package stdlib
+
+import (
+	"errors"
+	"io"
+	"net"
+	"syscall"
+)
+
+var errUnexpectedRead = errors.New("unexpected read from socket")
+
+func connCheck(c net.Conn) error {
+	var (
+		n    int
+		err  error
+		buff [1]byte
+	)
+
+	sconn, ok := c.(syscall.Conn)
+	if !ok {
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


damour:
@jackc Thank you for the extended answer!

I prepared [another PR](https://github.com/jackc/pgx/pull/1005) which provides the ability to define any custom function in ResetSession PTAL.

I'm closing this one.

Just to clarify your point about TLS connection:
> It is incompatible with TLS connections. 
Could you please briefly explain why this solution breaks TLS connections?
We don't use TLS unfortunately and I don't familiar with implementation details. At the first glance connCheck() should work with any socket, doesn't matter the protocol.

jackc:
> Could you please briefly explain why this solution breaks TLS connections?

`tls.Conn` does not implement `syscall.Conn`.

murfffi:
If somebody is here in 2026 and wants to use this feature, check out [this example](https://github.com/murfffi/conncheck/blob/2db680af/examples/sql.go#L27) of implementing a local connection check with TLS and Windows support . The solution builds on the approach proposed in this PR and uses the hook introduced in #1005 .

PR #1005, opened 2021-05-13
ResetSession hook that allows executing arbitrary code before a connection is reused from the pool.

damour:
A generic solution that allows executing arbitrary code before a connection is reused from the pool.

Also allows to fix the issue described in https://github.com/jackc/pgx/pull/1002

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -126,6 +126,15 @@ func OptionAfterConnect(ac func(context.Context, *pgx.Conn) error) OptionOpenDB 
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jackc:
That's the right idea, but it is implemented in the wrong place. The `ResetSession` hook shouldn't be available from regular pgx as it is inert there.

It should be used through only through `stdlib.OpenDB`. See `OptionBeforeConnect` and `OptionAfterConnect` for examples.

damour:
Sorry for the late answer, was on vacation.
@jackc I refactored it to OptionResetSession, please check.

Issue #1004, opened 2021-05-13
circular import between pgx, pgtype bloats go.sum


kevinburke1:
Because pgx imports pgtype and vice versa, each one holds a pointer to the older versions of the other one, going back really far into the project's history. This ends up bloating the go.sum file, and the ability to resolve dependencies, because it needs to hold a pointer to every version of every project that every dependency relies on. This also means the go.sum includes the version of every dependency from very early on in the life of pgx/pgtype...

It would be nice if the shared bits could be pulled out into a separate repo, or separate go.mod file to avoid the circular import problem.

jackc:
It's due to the `pgtype_test` package using pgx. There isn't technically a circular dependency. Sadly, `go.mod` / `go.sum` cascades all the test dependencies, too.

I'm considering pulling some repos apart to resolve this and similar issues but some of them would require a breaking change. Maybe for a possible `v5`.

jackc:
As of `v5` this is resolved.

Issue #977, opened 2021-03-26
Upgrade v4.10.1 => v4.11.0 makes dependency graph bloated


inliquid:
Hi,

have no idea what was in latest update (I'm using GORM with transitive dependency through `gorm.io/driver/postgres`), but when I updated to latest version (v4.10.1 => v4.11.0), my dependency tree becomes over bloated:

Before
```
go mod graph | wc -l
553
```

After
```
go mod graph | wc -l
1256
```

Just with this single line:
```
github.com/jackc/pgx/v4 v4.11.0 // indirect
```

pjediny:
I think the problem is the `go-kit`, not sure why should a db driver use a micro-service framework, in my case it broke build because of the `thrift` dependency that is incompatible with `cadence`. Looking at the code, I see it is because of the logger adapter implementation. In my opinion logger adapters should go out of the repo, or to some other (sub)module, maybe with the minimal adapter (to standard logger) and interface definition staying in this repo.

IIRC, the go-kit logger itself is an abstraction to other (structured) loggers, like zip and logrus, so we have direct adapters implemented here, indirect adapters too and too many of hard dependencies. Maybe author should consider to remove them completely, as they are trivial to implement and the provided implementations are not flexible enough.

Juneezee:
Could it be caused by this commit 9b59dd0346c6f4c93c0d89027ba107f8894541bf?

jackc:
Including log adapters and other optional code in the main repo but in separate packages predates Go modules.

It seems really unfortunate that optional and test dependencies are triggering this behavior (pgtype also has this issue https://github.com/jackc/pgtype/issues/97).

Is there no way around this outside of splitting the repositories? That would technically require a major version release... Not thrilled about inflicting that on everyone.

Also, pgx is already split into several modules (pgx, pgconn, pgtype, pgproto3, puddle, etc.). This already makes testing, changelogs, and releases more awkward than it would be as a monorepo. Not excited about making that even worse either.

Juneezee:
@jackc I think it's all about trade-offs. In my opinion, the benefit of slimming down the dependency tree outweights the drawbacks. Just like what @pjediny has suggested, the base Logger interface definition can stay in this repo.

Also just a small suggestion, you could consider creating an organization on GitHub and move the pgx family into it just to simplify the maintenance a little bit.

inliquid:
[...]

PR #1010, opened 2021-05-15
Add Go modules to reduce dependencies graph

psadac:
I'm not sure that this is the way you want to do that but it passes all the tests and it compiles on Windows 10.
However a script would probably be necessary to update each go.mod file when you modify pgx version, but it is probably not so complicated using `go mod edit`
It should solve #1004, #977


jackc:
This might be the right answer, but TBH I don't know.

I've read https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories several times and it looks like it might be the right approach, but that it also can be tricky to get right.

I'm also not sure if the import paths should or would need to change or not. I see that the new modules include the `v4` in the current location, but the new ones really aren't tagged as `v4`. As far as the Go module system is concerned they would be `v0` because we don't have any of the magic tags for sub-modules (and if we did tag them then they would be `.../v4.../v4`. But I don't know if that's a problem or not...

Also, I'm pretty sure pgxpool should stay in the primary module.

I'm really not sure what to do with this right now. Maybe someone can offer more perspective or an example from how another project handles this.

psadac:
Thank you for the link, didn't know this. If you are looking for an example you should take a look at [etcd](https://github.com/etcd-io/etcd). However I am not sure it will be transposable to your repo, they use release branches and tag versions on theses branches, which is not the way you proceed.

PR #1011, opened 2021-05-16
[proposal] Native interface exposing an API for high-level functions for managing Postgres

henvic:
*pgx.Conn and *pgxpool.Pool share a common set of functions that is what you'll need most of the time apart from opening/closing database connection.
This adds a Native interface with such functions to explicitly define what you likely want to be using on a high-level context.

I'm not 100% confident if this [is a good thing](https://golang.org/doc/effective_go#interfaces_and_types), but I adopted it in my projects and liked how it simplifies things and confident about using pgx safely.

Another possible downside of having this in github.com/jackc/pgx is making it a hard dependency, but the situation doesn't appear to be worse than explicitly adding *pgxpool.Pool.

### DIFF
--- a/protocol.go
+++ b/protocol.go
@@ -1,0 +1,66 @@
+package pgx
+
+import (
+	"context"
+
+	"github.com/jackc/pgconn"
+)
+
+// Native high-level interface for Postgres.
+// It is implemented by *pgx.Conn and *pgxpool.Pool, and you should use it only if you want to
+// delimit the scope of what your package using it should do.
+type Native interface {
+	// Begin starts a transaction. Unlike database/sql, the context only affects the begin command. i.e. there is no
+	// auto-rollback on context cancellation.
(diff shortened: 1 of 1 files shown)


jackc:
I have mixed feelings on adding a common interface. I have considered it in the past but have not pulled the trigger. In my case, I wanted an interface that is also satisfied by `Tx`.

But that's part of the issue, there isn't a perfectly correct answer for what methods this interface should require. And there is the Go rule of thumb to let the consumer define the interface when possible.

And given that a user can define whatever common interface(s) they need leads me to be slightly against this.

henvic:
I agree with both points. I'm closing the pull request for now.

The consumer defining the interface is precisely what I had in mind when I said that this might not be a good thing. Satisfiability by Tx, for **my** use case, would be going down the rabbit hole of abstractions for their own sake. Others with more complex code bases might have a good usage for them (and such an interface wouldn't help them).

Issue #831, opened 2020-09-11
read tcp XXX:YYYY->XXX:5432: i/o timeout


trashhalo:
I've been trying to track down this i/o timeout error. It seems to happen more frequently when the server is under load but one just came in when the db was @ 3% cpu.

This is a go app running in a docker container on an alpine image running in aws ecs fargate.

This comes back when executing queries. In the latest case. `GetContext` call. The entire time it waited was 56.21ms

I've been digging around pgx/pgconn code trying to understand it but running out ideas.

---

digging around the internet this `read tcp ... i/o timeout` issue seems to be a relatively common problem across many different systems. https://www.google.com/search?q=read+tcp+i/o+timeout+golang+site:stackoverflow.com

---
This appears to be the line in go that generates this error.
https://github.com/golang/go/blob/b115207baf6c2decc3820ada4574ef4e5ad940ec/src/net/net.go#L179

might be related too
https://github.com/golang/go/blob/e9ad52e46dee4b4f9c73ff44f44e1e234815800f/src/net/net.go#L573-L580

bkeroackdsc:
I'm seeing a lot of these errors. My unit tests run perfectly but at runtime lots of queries fail with these timeout errors, which I don't see with (for example) pq. I don't know if #701 is the answer, or if I'm doing something incorrectly but it's been frustrating so far.

bkeroackdsc:
For debug purposes I added a goroutine to the AfterConnect hook that pings every connection once a second. I see it pinging successfully, then a query timeout error, and (weirdly) the ping transiently fails with connection busy (presumably while the connection was busy with the query) and then it pings fine again.

Log output:
```
[...]
db: ping connection
db: ping connection
2020/09/18 00:12:16 error getting build: error getting build by id: failed to connect to `host=postgresql user=postgres database=furan`: failed to receive message (read tcp 172.17.0.7:53542->10.103.202.101:5432: i/o timeout)
2020/09/18 00:12:16 error listening for build events: error waiting for notification: 7d68435a-451b-438c-a0c9-bd7f173b5c03: read tcp 172.17.0.7:53538->10.103.202.101:5432: i/o timeout
error pinging connection: conn busy
db: ping connection
db: ping connection
db: ping connection
```

trashhalo:
What I ended up doing to try and reduce customer impact was wrapping our api with envoy which has built in error retry mechanisms. Obviously I'd rather the connection to the DB to be stable, but I'm out of ideas on how fix this.

[...]

PR #1030, opened 2021-06-24
Implement timeout error

MichaelDarr:
These changes address #831. In the issue thread, @atombender [summed up the issue and the solution](https://github.com/jackc/pgx/issues/831#issuecomment-819881537) implemented by these changes:

>The problem is distinguishing the error from a cancellation that occurs for other reasons. pgconn.Timeout() tests for context.Cancelled, which isn't universally a timeout signal.
>
> The only way to use pgconn.Timeout() "safely" is to always use it immediately on errors returned by pgx (or pgconn, etc.). If you wrap or pass the error on to higher-level layers, those layers can no longer know if the error came from pgx.
>
> This would really benefit from being wrapped in some special error type at the origin, in pgx. Otherwise we'll have to do this in every single pgx call:
>
> ```go
> func FetchStuff() (Data, error) {
>   // ...
>   if err := rows.Err(); err != nil {
>     return nil, convertError(err)
>   }
>   // ...
> }
>
> func convertError(err error) error {
>   if pgconn.Timeout(err) {
>     return ErrPostgresTimeout
>   }
>   return err
> }
>
> var ErrPostgresTimeout = errors.New("timeout")
> ```

These changes wrap timeout errors in a new error type, `ErrPostgresTimeout`. Because of error unwrapping, `ErrPostgresTimeout`  is fully compatible with the existing code, including `pgconn.Timeout`. Tests that utilize this function, such as `TestStmtExecContextCancel`, did not have to be altered to be compatible with these changes.

These changes also add a new function to the test suite, `TestTimeoutErr`. This exhaustive test ensures that all timeout errors are detected and wrapped by `ErrPostgresTimeout`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -409,6 +409,7 @@ func (c *Conn) Exec(ctx context.Context, sql string, arguments ...interface{}) (
 
 	commandTag, err := c.exec(ctx, sql, arguments...)
 	if err != nil {
+		err = wrapErrIfTimeout(err)
 		if c.shouldLog(LogLevelError) {
 			c.log(ctx, LogLevelError, "Exec", map[string]interface{}{"sql": sql, "args": logQueryArgs(arguments), "err": err})
 		}
@@ -603,7 +604,7 @@ optionLoop:
 			rows.resultReader = mrr.ResultReader()
 			rows.multiResultReader = mrr
 		} else {
-			err = mrr.Close()
+			err = wrapErrIfTimeout(mrr.Close())
 			rows.fatal(err)
 			return rows, err
 		}
--- a/rows.go
+++ b/rows.go
@@ -161,7 +161,7 @@ func (rows *connRows) CommandTag() pgconn.CommandTag {
 }
 
 func (rows *connRows) Err() error {
-	return rows.err
+	return wrapErrIfTimeout(rows.err)
 }
 
 // fatal signals an error occurred after the query was sent to the server. It
--- a/timeout.go
+++ b/timeout.go
@@ -1,0 +1,55 @@
+package pgx
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+)
+
+// ErrPostgresTimeout occurs when an error was caused by a timeout. Specifically, it is true if err is or was caused
+// by a context.Canceled, context.DeadlineExceeded or an implementer of net.Error where Timeout() is true.
+type ErrPostgresTimeout struct {
+	Err         error
+	isTimeout   bool
+	isTemporary bool
+}
+
+func (e *ErrPostgresTimeout) Error() string { return fmt.Sprint("postgres timeout: " + e.Err.Error()) }
+
+func (e *ErrPostgresTimeout) Unwrap() error { return e.Err }
+
+func (e *ErrPostgresTimeout) Temporary() bool {
+	return e.isTemporary
+}
+
+func (e *ErrPostgresTimeout) Timeout() bool {
+	return e.isTimeout
+}
+
+// wrapErrIfTimeout wraps an error if it was caused by a timeout. Otherwise, the passed error is returned as-is.
+func wrapErrIfTimeout(err error) error {
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return &ErrPostgresTimeout{
+			Err:         err,
+			isTimeout:   true,
+			isTemporary: netErr.Temporary(),
+		}
+	}
+	if errors.Is(err, context.Canceled) {
+		return &ErrPostgresTimeout{
+			Err:         err,
+			isTimeout:   false,
+			isTemporary: false,
+		}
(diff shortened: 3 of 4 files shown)


atombender:
Thanks for this, @MichaelDarr. I'm wondering if it's worth implementing the `net.Timeout` interface here, as it allows a central error handler (e.g. HTTP or gRPC middleware) to distinguish between timeouts and other errors without knowing what they are.

MichaelDarr:
No problem, I was pretty excited to see this issue still up for grabs! I'm pretty new to this sort of thing, and this task looked a) not very difficult and b) legitimately impactful. I spent a lot of yesterday extremely confused by that error message, thinking there was something wrong on the networking layer when it was actually just contexts getting canceled underneath me - whoops!

I don't see a [`net.Timeout` interface](https://golang.org/pkg/net/) - I might be missing something, or maybe you're referring to [`net.Error`](https://golang.org/pkg/net/#Error)? Whatever the case, I'm more than happy to tweak this and make it shine; all feedback is deeply appreciated.

atombender:
D'oh, yes, `net.Error`.

MichaelDarr:
Going to do a little research and push in that change - should be pretty easy!

MichaelDarr:
@atombender I've implemented the `net.Error` interface, but I'm not sure I did it properly - really flying by the seat of my pants on this one. For instance, I guess a `context.Canceled` isn't _really_ a timeout, so those should return false for `Temporary()`? Would definitely appreciate some feedback if you (or someone else who knows this code better than me) could offer it.

atombender:
The goal here, I think, should be to distinguish between genuine connection timeouts, and cancellations. So `context.Canceled` and `context.DeadlineExceeded` should both be passed through, because you can do stuff like:

```go
ctx, cancel := context.WithCancel(context.Background())
go func() {
  time.Sleep(1*time.Second)
  cancel()
}()
rows, err := conn.Query(...)
```

In this case, it's not a timeout.

How we distinguish this, I'm not sure — I haven't looked at the Pgx code recently to see how it enforces TCP timeouts. If it uses contexts internally, that could get tricky, because you'll want to distinguish between context cancellations originating with Pgx from those originating with the caller.

MichaelDarr:
I definitely think this error (`ErrPostgresTimeout`) should be as thin as possible, passing through the context errors like you mentioned. I'm going to push a commit reverting the changes to implement `net.Error`; I don't think the benefits of fulfilling the interface are worth the obfuscation that it layers on top of the underlying data, at least with the current implementation.

I'm sure the maintainers will have something to say about all this when they see it.

jackc:
I've read through original issue and the comments here. I'm not sure I understand the entire issue, but I have a few thoughts.

Someone mentioned that `context.Cancelled` being considered a timeout by `pgconn.Timeout` was a problem. I'm not sure I agree, but I can see the case for that assertion. If that is agreed to, then current behavior of `pgconn.Timeout` could be considered a bug and it could be fixed. Would that be enough to solve the problem? Also, this `ErrPostgresTimeout` continues treating `context.Cancelled` as a timeout. I think this should be decided one way or the other.

The next point was to distinguish a timeout error that occurred in pgx vs one that occurred at a higher level. That seems reasonable. But I think this should be implemented at a lower level in pgconn. A `ErrPostgresTimeout` type error wrapper could be implemented there and `pgconn.Timeout` could be updated to look for that. This would have the advantage of improving the existing interface instead of adding a new one.

MichaelDarr:
Thanks for taking a look @jackc. I'll see if I can implement this at a lower level in pgconn and open up a new PR over there.

atombender:
@jackc I agree, pgconn would be the place to do this.

There are, as you say, two problems:

* `pgconn.Timeout()` returns true even for errors that don't indicate a connection timeout
* When you cancel a context, this bubbles up to the caller as `*net.OpError` such as `read tcp [::1]:51084->[::1]:5432: i/o timeout`, which appears to blame the Postgres connection, when in fact the connection was probably fine.

The latter point has caused confusion among our SREs, who see logs filled with connection timeouts to Postgres, and assume that the database server is having issues.

jackc:
Here's a potential solution:

Always return the context error if the context was cancelled or expired. This is a little tricky due to concurrency but it should be possible. The calling code can directly check for `context.Canceled`, `context.DeadlineExceeded`, or `net.Error` (with `err.Timeout() == true`.

This might remove the need for `pgconn.Timeout()` altogether.

---

Then it becomes a separate matter if pgconn / pgx should wrap all errors with its own error type to allow programmatic determination of whether an error occurred in the database driver or elsewhere.

MichaelDarr:
@jackc @atombender I've opened a [draft PR in `pgconn`](https://github.com/jackc/pgconn/pull/79) with roughly the same functionality as what I proposed in this PR.

If we want to go with the solution proposed above (always returning the context error on cancellation or expiration), I'll close that PR as well.

Issue #748, opened 2020-05-14
Convert time.Duration to interval (with stdlib API)


jhwbarlow:
Is this possible while still retaining `database/sql` API compatibility?

For example, I have the following table:

```sql
  id BIGSERIAL NOT NULL PRIMARY KEY,
  team_name TEXT NOT NULL,
  pipeline_name TEXT NOT NULL,
  job_name TEXT NOT NULL,
  start_time TIMESTAMP NOT NULL,
  end_time TIMESTAMP NOT NULL,
  duration INTERVAL(0) NOT NULL,
  global_build_id BIGINT NOT NULL
```

Using this:

```go
const buildInsertStatement = `
INSERT INTO build_timings (
	team_name,
	pipeline_name,
	job_name,
	start_time,
	end_time,
	duration,
	global_build_id)
VALUES ($1, $2, $3, $4, $5, $6, $7)`

type Notifier struct {
	db *sql.DB
}

func New(db *sql.DB) *Notifier {
	return &Notifier{db}
}

func (n *Notifier) Notify(ctx context.Context, build concourse.Build) error {
	_, err := n.db.ExecContext(ctx,
		buildInsertStatement,
		build.Team().Name(),
		build.Pipeline().Name(),
		build.Job().Name(),
		build.Start(),
		build.End(),
		build.Duration(),
		build.ID())
	if err != nil {
		return fmt.Errorf("inserting build notification into database: %w", err)
	}

	return nil
}
```

Where the call to `build.Duration()` returns a `time.Duration`.

Unfortunately I get the following error :(

```
cannot convert 587120172000 to Interval
```

So it appears to be treating the `time.Duration` as its underlying `int64` type. I notice in `pgtype.Interval` there is a explicit type switch for `time.Duration` but it is clearly not matching:

```go
switch value := src.(type) {
	case time.Duration:
		*dst = Interval{Microseconds: int64(value) / 1000, Status: Present}
	default:
		if originalSrc, ok := underlyingPtrType(src); ok {
			return dst.Set(originalSrc)
		}
		return errors.Errorf("cannot convert %v to Interval", value)
	}
```

Is what I am trying to achieve possible?

Thanks!

cam-per:
Faced the same problem. The only solution I found is to add to switch
```go
case int64:
    *dst = Interval{Microseconds: value / 1000, Status: Present}
```

sashayakovtseva:
+1 for supporting this

bfontaine:
[...]

PR #1041, opened 2021-07-13
Support time.Duration with simple protocol

MichaelDarr:
These changes address #748, which notes that `time.Duration` can't be used as a PostgreSQL interval. The underlying problem isn't `stdlib`, as was postulated there - it's the simple protocol. This is easily rectified by converting `time.Duration` to a string in the [expected PostgreSQL syntax](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT) instead of to an integer.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1061,3 +1061,16 @@ func TestStmtCacheInvalidationTx(t *testing.T) {
 
 	ensureConnValid(t, conn)
 }
+
+func TestInsertDurationInterval(t *testing.T) {
+	testWithAndWithoutPreferSimpleProtocol(t, func(t *testing.T, conn *pgx.Conn) {
+		_, err := conn.Exec(context.Background(), "create temporary table t(duration INTERVAL(0) NOT NULL)")
+		require.NoError(t, err)
+
+		result, err := conn.Exec(context.Background(), "insert into t(duration) values($1)", time.Minute)
+		require.NoError(t, err)
+
+		n := result.RowsAffected()
+		require.EqualValues(t, 1, n)
+	})
+}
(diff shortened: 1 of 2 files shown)


MichaelDarr:
Would you mind taking a peek at this, @jackc? Unless I'm overlooking something, this feels like an easy, low-risk change that meaningfully improves usage.

wlbr:
Hey, would you mind to create a (release) tag, so that we could get the feature simply by using go get / mod?
Thank's in advance.

jackc:
Just released `v4.13.0`.

wlbr:
Works!
Thanks a lot!!

gz-95:
On `v4.13.0` with the following

```
row := db.QueryRow("SELECT INTERVAL '5 minutes';")

var d time.Duration
err := row.Scan(&v)
```

Produces

`sql: Scan error on column index 0, name "interval": converting driver.Value type string ("00:05:00") to a int64: invalid syntax`

BigBoulard:
> On `v4.13.0` with the following
> 
> ```
> row := db.QueryRow("SELECT INTERVAL '5 minutes';")
> 
> var d time.Duration
> err := row.Scan(&v)
> ```
> 
> Produces
> 
> `sql: Scan error on column index 0, name "interval": converting driver.Value type string ("00:05:00") to a int64: invalid syntax`

Do you still have this issue?

sean-:
Yes, as of v5.7.5, this is still a problem if you use the stdlib database interface.  Depending on your application, the right fix is to use pgxpool (or pgx) instead of stdlib sql.DB, at which point this type conversion problem goes away.

PR #1050, opened 2021-07-22
Release the connection in pool.QueryRow

chrisdavies:
This PR fixes a bug I ran into in production. My app hangs after a while, and querying `pg_stat_activity` indicated connections building up over time. They are all running queries with `pool.queryRow`.

I was able to reproduce the issue by setting my local max pool size to 1 and calling `QueryRow` a few times.

Adding `c.Release()` fixed the issue.

Thanks for the great library!.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -458,6 +458,7 @@ func (p *Pool) QueryRow(ctx context.Context, sql string, args ...interface{}) pg
 	if err != nil {
 		return errRow{err: err}
 	}
+	defer c.Release()
 
 	row := c.QueryRow(ctx, sql, args...)
 	return c.getPoolRow(row)


jackc:
I don't think this change is correct. `Release` has to be called before the row is read or there could be a race condition. And it is called automatically by Scan: https://github.com/jackc/pgx/blob/59fa1868a788a860aeec78ace1ee2ade4ee30810/pgxpool/rows.go#L102

My guess is in your application code you are not `Scan`ing the result of `QueryRow`.

chrisdavies:
Ah. You're right. There's one path where I don't care about the result and don't do a scan on the row. Thanks! Well, I'll close this PR, and opt for a helper that always ensures the release in my own codebase, instead!

PR #1058, opened 2021-08-05
Update current dependencies and add cache in workflow

mmorel-35:
HI : I have used dependabot to update the current dependencies.
All tests where passed.


jackc:
So what will this dependabot action do? Will it automatically update dependencies?

To be honest, I'm not sure I like updating dependencies without a concrete reason. In general, I think pgx should depend on the lowest version possible and let the application decide what version to use.

For example, say pgx requires package `foo` `1.1` and the latest version is `1.5`. If someone's application is using `1.3` we would be forcing them to upgrade `foo` when they upgrade `pgx`. Now presumably there is no problem with them using `1.5`. But if they have some reason for using `1.3` we now have forced them to either not upgrade `pgx` or to upgrade `foo`.

Security issues in dependencies are one thing, but outside of that I would rather not force the issue.

---

The caching thing looks cool though. I assume that would make CI tests run faster?

mmorel-35:
IMHO, keeping your dependencies up-to-date help you fight against [technical debt](https://en.wikipedia.org/wiki/Technical_debt). Security is one of the reasons but that's not the only one. 
If you want to know more, you can take a look over [here](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates).
The bot will never modify your project himself, only create pr, you'll have to validate if you want them included in. 

I'm just offering this possibility to your project. I'm not here to change your mind.  Many projects are using this because they don't want to handle the updates of their dependencies manually. 

Concerning the version management, I'm not sure it works this way. The project are always using the version they want even if your project provides a different one. Both will be in the go.sum by the way.

PR #1070, opened 2021-08-19
Fix doc replicated words

smiletrl:
- One small doc fix
- Pgx pool QueryRow misses release https://github.com/jackc/pgx/blob/master/pgxpool/pool.go#L456

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -96,7 +96,7 @@ type Config struct {
 	// AfterConnect is called after a connection is established, but before it is added to the pool.
 	AfterConnect func(context.Context, *pgx.Conn) error
 
-	// BeforeAcquire is called before before a connection is acquired from the pool. It must return true to allow the
+	// BeforeAcquire is called before a connection is acquired from the pool. It must return true to allow the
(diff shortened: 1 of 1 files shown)


jackc:
The connection can't be released until the row is scanned and the release happens when it is scanned. See #1050 for more discussion of the same issue.

smiletrl:
ah, I see. Could you help describe the race condition if release happens before scan?

smiletrl:
Right now, `QueryRow` must be chained by Scan(), otherwise it will eat all available connections. In this case, I feel `QueryRow` should be removed, but replaced with something like `QueryRowScan`, which will always release the connection.

Is this struct `Pool` is implementing some interface, which must include `QueryRow`?

jackc:
Why would `QueryRow` be called without also calling `Scan`?

smiletrl:
ah, personally I don't have a specific case without calling `Scan`. Just feel it doesn't look right that if one function must be chained by another one, otherwise it could break the application, like bug reported at https://github.com/jackc/pgx/pull/1050.

smiletrl:
This release is removed now^ This pr only fixes that doc.

PR #1079, opened 2021-09-10
Add some comments to pgxpool package

gilcrest:
Hi @jackc and other maintainers. I am switching from `lib/pq` to this library and it's been a pleasure so far. Thanks a lot for all the hard work! As I'm reading through the library and trying to understand how it works, I thought I could help by adding a few comments. If it's not helpful, I won't continue, but I wanted to see if this was something you're open to. A lot of the time, I'm copying comments from other functions that have the same or similar meaning and tweaking them slightly (to consider the pool, etc.)...

Thanks again!

Dan


jackc:
LGTM.

PR #1081, opened 2021-09-17
Add context options to zerologadapter

thomasf:
### WithContextFunc
WithContextFunc adds possibility to get request scoped values from the ctx.Context before logging lines.

I usually insert the request id from `zerolog.hlog.IDFromCtx` in the pgx logger and this lets users do that or even get the whole logger from `zerolog/log.Ctx` to get the whole logger from the context if they want to.

### WithoutPGXModule
WithoutPGXModule disables adding module:pgx to the default logger context.

I usually set the caller to pgx instead so I have just removed the module:pgx from my own modified version.

### DIFF
--- a/log/zerologadapter/adapter.go
+++ b/log/zerologadapter/adapter.go
@@ -9,15 +9,42 @@ import (
 )
 
 type Logger struct {
-	logger zerolog.Logger
+	logger     zerolog.Logger
+	withFunc   func(context.Context, zerolog.Context) zerolog.Context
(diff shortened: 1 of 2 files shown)


thomasf:
Maybe the tests are a bit too verbose, they probably go a little bit into example test territory.

I have some variant of this zerologadapter change in every project with a http handler for a while now so I guess attempting to upstream it is a good idea.

jackc:
LGTM

PR #1085, opened 2021-09-25
feat: remove prepare operation in copy from

KennyChenFight:
I think it's not necessary to use prepare operation to set value type before copy from operation. It would produce one RTT and slow down copy from operation.

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"time"
 
-	"github.com/jackc/pgconn"
 	"github.com/jackc/pgio"
 )
 
@@ -99,11 +98,6 @@ func (ct *copyFrom) run(ctx context.Context) (int64, error) {
 	}
 	quotedColumnNames := cbuf.String()
 
-	sd, err := ct.conn.Prepare(ctx, "", fmt.Sprintf("select %s from %s", quotedColumnNames, quotedTableName))
-	if err != nil {
-		return 0, err
-	}
-
 	r, w := io.Pipe()
 	doneChan := make(chan struct{})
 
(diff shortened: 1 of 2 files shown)


jackc:
I think that step is necessary to handle native Go values like `string` and `int32`...

mixmind:
I have same question, 

I'm using extension(pipelinedb), there is no simple select from table that pgx tries to prepare statement.
This extension creates foreign table and simple select from it can only throw error, that it's stream and not a table.
This table only in use for inserts, or creating view from it. 

Removing prepare statements would help for me too.
Can you please suggest some workaround , that I could use COPY too?

jackc:
@mixmind You could drop down to the pgconn level and construct the copy stream manually. See https://pkg.go.dev/github.com/jackc/pgconn#PgConn.CopyFrom.

bfontaine:
@KennyChenFight any thought on Jack’s comment?

alejandrodnm:
Hi @jackc we are using CopyFrom in a write intensive application. I wanted to get rid of the prepare statement, would you accept a PR that exposes a new `CopyFromWithStatementDescription` method that accepts a `pgconn.StatementDescription` as argument?

jackc:
@alejandrodnm I don't mind an improvement to copy from performance, but I'm not convinced that a `CopyFromWithStatementDescription` is the correct approach.

What about `CopyFrom` using the automatic statement or description caching like `Query` does? Then everyone would benefit automatically. Detecting / handling schema changes might be tricky though.

alejandrodnm:
Sounds good @jackc , I'll give it a try. 

Thanks

alejandrodnm:
@jackc I created PR #1412 for what we discussed

rubiagatra:
Is this PR still relevant? @KennyChenFight

PR #1094, opened 2021-10-03
Eager initialize minpoolsize on connect

BornToGrill:
Currently minPoolSize is ignored on connect and the idle connections are only created once the background check runs.
This breaks developer expectations and has a big performance impact on applications that are autoscaled.
Autoscaled applications will immediatly receive lots of traffic and if the pool isn't ready the connections will have to be created the first few requests (lazy) instead of the startup of the application.

Currently our applications use this workaround:
```go
func warmupConnections(ctx context.Context, conn *pgxpool.Pool, targetConnections int) error {
	for i := 0; i < targetConnections; i++ {
		c, err := conn.Acquire(ctx)
		if err != nil {
			return err
		}
		defer c.Release()
	}
	return nil
}
```

But it would be nice if this library would enforce the min pool size at all times.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -225,6 +225,10 @@ func ConnectConfig(ctx context.Context, config *Config) (*Pool, error) {
 	go p.backgroundHealthCheck()
 
 	if !config.LazyConnect {
+		if err := p.createIdleResources(ctx, int(p.minConns)); err != nil {
+			return nil, err
+		}
+
 		// Initially establish one connection
 		res, err := p.p.Acquire(ctx)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


BornToGrill:
Forgot to run with the -race flag. Fixed my tests.
(TestPoolBackgroundChecksMaxConnIdleTime is flakey on my machine even on master).

jackc:
LGTM

jackc:
Hmm... actually it looks like a test is flickering.

https://github.com/jackc/pgx/runs/4055666218?check_suite_focus=true#step:5:885

It's probably not feasible to check which connection attempt failed since they all happen concurrently.

BornToGrill:
First error is indeed not guaranteed. We really only care about it being any error other than a context cancelation.
Made a quickfix #1106

PR #1108, opened 2021-11-10
conn: Wrap the notifications buffer with a lock

tchap:
It seems to me that the way notifications are being buffered is unsafe. That is why I am adding a lock here.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/jackc/pgconn"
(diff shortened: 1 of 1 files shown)


jackc:
I don't understand what case this is guarding against. `*pgx.Conn` doesn't claim to be concurrency safe. i.e. It's designed to be used with a pool when concurrency is needed.

tchap:
@jackc I don't know exactly how this works, but it seems pretty suspicious to me that you register a callback somewhere internally and then you read the same slice via public API when you call `WaitForNotification`. Are you suggesting there is a mechanism that ensures that `bufferNotifications` and `WaitForNotification` are never called at once? I am not sure how that should be enforced... In any case, I can avoid this whole mechanism by using a custom notifications callback, but I would rather fix it here if it is an actual issue...

jackc:
That callback is only triggered by `receiveMessage` https://github.com/jackc/pgconn/blob/162dc65eff6f037c98baa36f1f4c75658408d65a/pgconn.go#L515. That will only be called when reading from the connection which would only happen when the application called `Query`, `Exec`, etc. There is no background reading from the connection.

tchap:
Alright then, sorry to bother you in that case 👼🏻

PR #1111, opened 2021-11-13
Fix *pgxpool.Pool Acquire lock issue if nil is passed as context.

henvic:
Panic immediately if nil is used as value for context.Context.

I'm not sure where it'd be preferred to solve this (or if in both places).

Related: https://github.com/jackc/puddle/pull/13

### DIFF
--- a/pgxpool/pool_test.go
+++ b/pgxpool/pool_test.go
@@ -143,6 +143,24 @@ func TestPoolAcquireFuncReturnsFnError(t *testing.T) {
 	require.EqualError(t, err, "some error")
 }
 
+func TestPoolCloseAfterAcquireNilContextPanics(t *testing.T) {
+	t.Parallel()
+
+	pool, err := pgxpool.Connect(context.Background(), os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+	defer pool.Close()
+
+	c := make(chan struct{}, 1)
+	go func() {
+		defer func() {
+			c <- struct{}{}
+		}()
+		assert.PanicsWithValue(t, "tried to acquire connection with nil context", func() { pool.Acquire(nil) })
+	}()
+	<-c
+	pool.Close()
+}
+
 func TestPoolBeforeConnect(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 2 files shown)


pgxpool/pool_test.go
```diff
@@ -143,6 +143,24 @@ func TestPoolAcquireFuncReturnsFnError(t *testing.T) {
...
 
+func TestPoolCloseAfterAcquireNilContextPanics(t *testing.T) {
+	t.Parallel()
+
+	pool, err := pgxpool.Connect(context.Background(), os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+	defer pool.Close()
+
+	c := make(chan struct{}, 1)
+	go func() {
```
henvic, line 154: It doesn't really have to be on a separate goroutine. It can be on the same one. This was just a way to show that there is a resource leak that remains.

jackc:
Can you explain this a little more? What is it that happens when on a nil context? And what is the resource leak you refer to in the test?

henvic:
On `*Pool.doAcquire` there is the following:

```go
func (p *Pool) doAcquire(ctx context.Context, block bool) (*Resource, error) {
startNano := nanotime()
p.cond.L.Lock()
if doneChan := ctx.Done(); doneChan != nil {
	select {
	case <-ctx.Done():
		p.canceledAcquireCount += 1
		p.cond.L.Unlock()
		return nil, ctx.Err()
	default:
	}
}

// ...
```

Notice the `p.cond.L.Lock()` call in the very beginning. This lock isn't unlocked by a defer statement.
Therefore, if a panic happens, it's never going to be unlocked.

One cause of panic I've identified is when you make the mistake of passing `nil` as a context. So, my suggestion is to panic immediately if `ctx == nil` as this already happens and you shouldn't be using nil as a context anyways.

Maybe it's also desired to use `defer p.cond.L.Unlock()` just after locking for the first time, but this is more sensitive and I took the easy way out.

A secondary benefit of panicking with this message is that the library user becomes aware of what caused the failure faster.

henvic:
If you leave the test code but comment out the `if ctx == nil { panic(...) }` block I wrote on the application code, you'll see that it already panics, but with a less useful message, and leaving behind the lock locked.

jackc:
Okay. I think I understand the issue. I think this would be better handled in puddle. I'll respond there.

PR #1116, opened 2021-11-21
Change constant definition syntax so they show up in godoc

Jille:
for TxAccessMode, TxDeferrableMode and TxIsoLevel

After this commit, Godoc understands these are the valid values of these types and shows them together in the documentation.


jackc:
LGTM

Issue #1127, opened 2021-12-14
PGX Greatly Increases Memory Usage (compared to pq)


noahschumacher:
I recently switched a couple of my services from PQ to PGX. The first service I switched over I didn't notice anything. For the second service however I quickly noticed a sharp increase in memory consumption. For awhile I thought this was a memory leak issue due to my implementation, however I noticed that if I let the service run long enough the memory would stabilize are 8x what it was with PQ. 

When I realized this wasn't a memory leak I checked the first smaller (less traffic) service and noticed it had experienced that same  rise in memory but less so (around 4x). Looking at the `pprof` `top` output:

```
Showing top 10 nodes out of 147
      flat  flat%   sum%        cum   cum%
   35.64MB 17.78% 17.78%    35.64MB 17.78%  google.golang.org/grpc/internal/transport.newBufWriter
   33.53MB 16.73% 34.51%    43.03MB 21.47%  google.golang.org/grpc/internal/transport.(*http2Server).operateHeaders
   17.64MB  8.80% 43.31%    17.64MB  8.80%  github.com/jackc/chunkreader/v2.(*ChunkReader).newBuf
   15.50MB  7.73% 51.04%    15.50MB  7.73%  context.WithValue
   14.94MB  7.45% 58.49%    14.94MB  7.45%  bufio.NewReaderSize
    9.51MB  4.74% 63.24%     9.51MB  4.74%  runtime.allocm
    9.50MB  4.74% 67.98%     9.50MB  4.74%  bytes.(*Buffer).String
    8.54MB  4.26% 72.24%     8.54MB  4.26%  golang.org/x/net/http2/hpack.(*headerFieldTable).addEntry
       5MB  2.50% 74.73%        5MB  2.50%  github.com/grpc-ecosystem/go-grpc-middleware/tags.(*mapTags).Set
       4MB  2.00% 76.73%        4MB  2.00%  github.com/uber/jaeger-client-go.simpleSpanAllocator.Get
```

I noticed a few instances were mentioning buffer which led me to [this file](https://github.com/jackc/chunkreader/blob/master/chunkreader.go) and specifically this comment:
```
// ChunkReader is a io.Reader wrapper that minimizes IO reads and memory allocations. It allocates memory in chunks and
// will read as much as will fit in the current buffer in a single call regardless of how large a read is actually
// requested. The memory returned via Next is owned by the caller. This avoids the need for an additional copy.
//
[...]

PR #1134, opened 2021-12-29
Allocate connRows on demand instead of preallocating in bulk

jackc:
The 64 element preallocatedRows may be pinning memory from previous
queries.

See https://github.com/jackc/pgx/issues/1127


jackc:
I suppose one extra allocation a query is a fair trade to avoid the risk of pinning memory.

PR #1143, opened 2022-01-20
Add configurable log filter to be able to skip certain logs

yaronius:
Hi Jack!

First and foremost, thank you for this great project, it's very useful! 🙇🏻 

While using this library, I found it inconvenient that I cannot skip logging some internal queries. For example, I use `conn.Ping()` for health checks and it results in plenty of `Exec` logs with the query being `;` (which is the default empty query used for pinging). I could switch the log level but I do want to log the actual queries. So I was thinking of adding an (optional) filter function that can be customised and used for this purpose - to skip certain logs depending on their data rather than log level. I'm not absolutely sure it's the best approach but it's quite straightforward and simple with minimal changes required. What do you think? Looking forward to your reply 🙏🏻

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -19,8 +19,9 @@ import (
 // then it can be modified. A manually initialized ConnConfig will cause ConnectConfig to panic.
 type ConnConfig struct {
 	pgconn.Config
-	Logger   Logger
-	LogLevel LogLevel
+	Logger    Logger
+	LogLevel  LogLevel
+	LogFilter LogFilter
 
 	// Original connection string that was parsed into config.
 	connString string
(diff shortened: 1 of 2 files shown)


jackc:
That's a useful feature, but it can already easily be done without any modification to pgx.

Notice that the signature of `LogFilter` is very similar to the signature of the `Log` method of the `Logger` interface. 

Instead of a filter as separate concept a filter would be a type that implemented the `Logger` interface and wrapped another logger. It would only pass on the log messages that shouldn't be filtered.

yaronius:
That's good news, thank you for explaining the possible workaround, I'll close this PR and will try your solution 🙇🏻

Issue #467, opened 2018-10-14
rows.Values() panics with index out of range


atedja:
Sample code:
```
	connConfig := pgx.ConnConfig{
		Database: "postgres",
		Host:     "localhost",
		Password: "local123",
		Port:     uint16(5432),
		User:     "postgres",
	}

	db, err := pgx.Connect(connConfig)
	if err != nil {
		panic(err.Error())
	}

	rows, err := db.Query("SELECT * FROM sample")
	if err != nil {
		panic(err.Error())
	}

	values, err := rows.Values()
	if err != nil {
		panic(err.Error())
	}

	fmt.Println(len(values))

	rows.Close()
```

Result:
```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/jackc/pgx.(*Rows).nextColumn(0xc420120940, 0x0, 0x2, 0xc420127120, 0x0, 0x2)
        .../go/src/github.com/jackc/pgx/query.go:178 +0x124
github.com/jackc/pgx.(*Rows).Values(0xc420120940, 0x6d43cb, 0x14, 0x0, 0x0, 0x0)
        ../go/src/github.com/jackc/pgx/query.go:283 +0xd4
main.main()
        .../go/src/testpg/main.go:27 +0xf0
exit status 2
```

If I try to `fmt.Println(rows)`, I'd get this: `&{0xc4200e8000 <nil> <nil> [] [{id 16386 1 23 4 int4 -1 1} {name 16386 2 25 -1 text -1 1}] 0 0 <nil> {13756467921915557982 5028934 0x82cba0} SELECT * FROM sample [] true false}`, which doesn't look like the rows are corrupted as the `err` field is `<nil>`

Using `rows.Next()` and iteratively call `rows.Scan()` works as expected.

jackc:
`rows.Next()` must be called before `rows.Values()` just like `rows.Scan`. It is always a bug if this is not done, so a panic is the correct behavior.

PR #1147, opened 2022-01-29
Clarify that Values() and Scan() require Next() to have been called on the rows object

djsavvy:
I ran into the same issue as the author of https://github.com/jackc/pgx/issues/467 and didn't find the disclaimer where I'd expect to see this info in the documentation. Therefore, I made this PR to put it there. 

I am of course willing to reword this to @jackc's satisfaction. Thanks so much for creating this library, `pgx` is pretty damn awesome.


jackc:
👍

PR #1153, opened 2022-02-11
exposing stdlib DB connector

jimenez:
Hello,

Following https://github.com/jackc/pgx/issues/481 
I'm having an issue while using `gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql` with `pgx`, specifically, I'm looking for a way to set the `BeforeConnect` while still using sqltrace. 

Allowing access to the connector will allow me to do this:

```
// add stdlib.OptionBeforeConnect to opts here 

// pass connector with BeforeConnect callback to be used in sqltrace
 sqltrace.OpenDB(stdlib.GetConnector(*cfg, opts...))
```

Is there is another way of doing this?

Thanks,

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -163,7 +163,7 @@ func RandomizeHostOrderFunc(ctx context.Context, connConfig *pgx.ConnConfig) err
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


jackc:
It seems reasonable to me.

PR #1154, opened 2022-02-11
ref: remove anchored code

sshaplygin:
Remove not used code


cristaloleg (commented):
You cannot just remove public API without a major release.

PR #1160, opened 2022-03-02
Fix single line comment for line endings in mac when sanitizing.

psyhatter:
Line endings on macs are \r, so sanitizing single-line comments did not correctly parse query arguments up to \n


jackc:
Thanks!

PR #1200, opened 2022-05-07
feat: add time duration to error query and copy and implement batch logging

EpicStep:
Hi, these changes will help in writing tracing modules.
But I think that in future versions it is necessary to add the start and end times, instead of duration.


gvencadze (approved):

EpicStep:
@jackc maybe i need to make these changes in v5?

jackc:
> maybe i need to make these changes in v5?

Don't worry about it. I merge v4 into v5 every now and then and I suspect this should be a clean merge.

Also, in the future please run the tests. This change broke a few tests. I fixed it in dc0ad04ff58f72f4819289f54745a36124cdbec3.

gvencadze:
@jackc hello, can you tell how soon we will see new version with this changes?

jackc:
There is usually a new release every 2 or 3 months unless an urgent bug fix requires an immediate point release.

Issue #1061, opened 2021-08-07
How to improve logging and tracing support in future release?


jackc:
The original design goal for logging was simply to log what queries are being run and how long they take. 

There have been a few requests to support additional logging / tracing points or new ways of configuring what is currently logged (#853, #998). The current logging system works fine for me, but as I'm considering the possibility of a new major release I'd like to open the door to breaking changes to the logging / tracing interface.

As I do not have a first-hand need for this I do not have any opinion on what these changes might be, and I would probably not implement them myself, but I wanted to bring up the possible opportunity for those who may have the interest and ability while there is a window open for possible breaking changes.

Tochemey:
Hello @jackc thanks for such feature. That will help a lot

HakShak:
Some references:

https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5
https://github.com/XSAM/otelsql

Tochemey:
@jackc any progress on this.

jackc:
> @jackc any progress on this.

I have not heard of anyone working on this.

obitech:
I'm wondering how practical it would be to provide some sort of hooks or middleware chains which could be triggered before  or after an `Exec` is called on the underlying `PgConn`. Similar to the `BeforeAcquire` functionality in a Pool. These hooks could just receive a context which is used for span propagation. WDYT @jackc ?

jackc:
> I'm wondering how practical it would be to provide some sort of hooks or middleware chains which could be triggered before or after an Exec is called on the underlying PgConn. Similar to the BeforeAcquire functionality in a Pool. These hooks could just receive a context which is used for span propagation. WDYT @jackc ?

I like the idea of hooks, but there would need to be a lot more than before / after `pgconn.Exec`. There's also `pgconn.ExecPrepared` and `pgconn.ExecParams`. Then those methods only send the queries. They don't read the results. There would need to be something after the results are received. And maybe when the first result comes back. Also, there are batch requests.

I could see something like:

```go
type BeforeQueryHook interface {
  BeforeQuery(...)
}
```

Every hook point could have its own interface. The logger or tracer could implement as many or as few hooks as it wanted to. This would allow adding or even removing hooks in a backwards compatible way.

obitech:
Couple more questions since I'm not very familiar with the inner workings of pgx:

- Would there be interfaces for every command like `BeforeQueryHook`, `AfterQueryHook`, `BeforeQueryRowHook`, and so on? 
[...]

PR #1212, opened 2022-05-26
Increase level of connection open/close log messages

lzap:
Hey, this might be a bit bold, but when our app starts I see bunch of open/close messages logged at info level and I feel like as an app developer I am not too interested in these. What matters are SQL statements typically, these are unfortunately also at info level.

So I propose to increase level of these connection-related messages to trace (or at least debug) if you don't mind. Also, I've noticed that most of the logging messages start with capital letter.

Cheers!

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -216,8 +216,8 @@ func connect(ctx context.Context, config *ConnConfig) (c *Conn, err error) {
 		}
 	}
 
-	if c.shouldLog(LogLevelInfo) {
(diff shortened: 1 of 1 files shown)


jackc:
I'm sympathetic to the issue, but I'd rather not change behavior in a minor release. But I'm hoping for a much more flexible system for `v5` where each log message or trace point can be controlled individually. See https://github.com/jackc/pgx/issues/1061.

lzap:
Sure, if this could be also changed I'd appreciate it. Thanks!

PR #1232, opened 2022-06-21
Add logger func wrapper for adapter creation

ubiuser:
This PR adds a convenience logger function wrapper to make logger adapter creation easier. This solution is analogous to how a logger is added to an AWS config using aws-sdk-go-v2, see https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config#WithLogger. Implementation is based on https://github.com/aws/smithy-go/blob/main/logging/logger.go.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1,7 +1,9 @@
 package pgx_test
 
 import (
+	"bytes"
 	"context"
+	"log"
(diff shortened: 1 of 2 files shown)


jackc:
LGTM

PR #1261, opened 2022-07-25
Ignore cancellation in puddle constructor

### LINKED ISSUE #1259: Dial is affected by query context cancellation
We noticed a lot of: `failed to connect to ...: dial error (dial tcp ...: operation was canceled)` and this seems to stem from the fact that puddle does `p.constructResourceValue(ctx)` and passes along the context from the original query. If the query is cancelled (user closed the tab, etc), we don't necessarily want to throw away the connection.

While I do think there's value in passing a context (which might contain helpful information on what originated the request and tracing information) maybe the puddle constructor (in pgxpool) should be overwriting the original context's cancellation? There are some implementations of how to ignore the cancellation out there in https://github.com/golang/go/issues/40221 (Under Existing Implementations). This would make acquire-initiated connection creations match pool-health-initiated connection creattions (which just send context.Background() to `p.p.CreateResource(ctx)`).

What was happening to us is that there were hundreds of these `dial error ... operation was canceled` and we were DDOSing the DB without creating any useful connections. This was during a period where the DB was already struggling and was taking longer to create connections. This issue further exacerbated the problem by causing even more load on the DB.

jameshartig:
Fixes #1259

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -70,6 +70,16 @@ func (cr *connResource) getPoolRows(c *Conn, r pgx.Rows) *poolRows {
 	return pr
 }
 
+// detachedCtx wraps a context and will never be canceled, regardless of if
+// the wrapped one is cancelled. The Err() method will never return any errors.
+type detachedCtx struct {
+	context.Context
+}
+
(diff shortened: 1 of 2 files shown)


jackc:
LGTM

PR #1262, opened 2022-07-25
pgxpool: kill connections after max configured life time.

stefanbeller:
We ran into issues of a web request erring out after 15 seconds due to the underlying DB request taking more than 15 seconds. Investigating the issue lead me to believe the destructor should also respect the configured timeout.

Looking at the introduction, 39b096d0 ("pgxpool waits for connection cleanup to finish before making room in pool", 2020-08-20), I am unsure how to add a proper test.


jackc:
I don't quite understand the underlying problem or how this should solve it.

The destructor should be run in a different goroutine from any application code. i.e. a timeout in the destructor should never be able to affect a query.

PR #1270, opened 2022-08-03
pgxpool: Make BeginTx success case clearer

PeterFeicht:
`err` is nil at this point, so might as well return `nil` as the error to make that fact obvious.


jackc:
LGTM

PR #1278, opened 2022-08-10
Create package level variable ErrNoResult

adrubesh:
This is a very simple, non-breaking change that brings the `no result` error in as a package variable. This simplifies scanning batch results to something like:

```go
for {
  i, err := br.Query()
  if err != nil {
    if errors.Is(err, ErrNoResult) {
      break
    }

    return err
   }
  ... 
}
```

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -61,6 +61,9 @@ type batchResults struct {
 	closed bool
 }
 
+// ErrNoResult occurs on a batch operation when no result is available.
+var ErrNoResult = errors.New("no result")
+
 // Exec reads the results from the next query in the batch as if the query has been sent with Exec.
 func (br *batchResults) Exec() (pgconn.CommandTag, error) {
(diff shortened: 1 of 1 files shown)


jackc:
Is this an error that an application should handle? It seems to me that the only way this error occurs is when there is an application bug. If that is the case I don't see the need for a named error as the application should never be checking for it.

adrubesh:
> Is this an error that an application should handle? It seems to me that the only way this error occurs is when there is an application bug. If that is the case I don't see the need for a named error as the application should never be checking for it.

It seems that way to me,  for instance in the example I gave this allows you to itetate over a Batch Result until err is not nil and distinguish between there being no more rows left to read (which is what we want)  and a different error(which we handle as all other errors)

jackc:
What I mean is this implies you don't know how many are queries in the batch. But you should always know that.

PR #1281, opened 2022-08-14
Make FieldDescription.Name a string to avoid corruption

kirk-anchor:
I sometimes get corrupted column names returned from v5's `func (r *Rows) Columns() []string` method. The `Name` field on `FieldDescription` is getting overwritten.

The `Name` field can be changed to a `string`, which would implicitly copy the bytes and avoid corruption. If it will remain a `[]byte`, the name should be copied from the buffer so the name will not get corrupted if the buffer is overwritten.

### DIFF
--- a/pgproto3/row_description.go
+++ b/pgproto3/row_description.go
@@ -34,7 +34,7 @@ func (fd FieldDescription) MarshalJSON() ([]byte, error) {
 		TypeModifier         int32
 		Format               int16
 	}{
-		Name:                 string(fd.Name),
+		Name:                 fd.Name,
 		TableOID:             fd.TableOID,
 		TableAttributeNumber: fd.TableAttributeNumber,
 		DataTypeOID:          fd.DataTypeOID,
--- a/pgproto3/row_description.go
+++ b/pgproto3/row_description.go
@@ -14,7 +14,7 @@ const (
 )
 
 type FieldDescription struct {
-	Name                 []byte
+	Name                 string
 	TableOID             uint32
 	TableAttributeNumber uint16
 	DataTypeOID          uint32
@@ -70,7 +70,7 @@ func (dst *RowDescription) Decode(src []byte) error {
 		if idx < 0 {
 			return &invalidMessageFormatErr{messageType: "RowDescription"}
 		}
-		fd.Name = src[rp : rp+idx]
+		fd.Name = string(src[rp : rp+idx])
 		rp += idx + 1
 
 		// Since buf.Next() doesn't return an error if we hit the end of the buffer
@@ -152,7 +152,7 @@ func (dst *RowDescription) UnmarshalJSON(data []byte) error {
 	dst.Fields = make([]FieldDescription, len(msg.Fields))
 	for n, field := range msg.Fields {
 		dst.Fields[n] = FieldDescription{
-			Name:                 []byte(field.Name),
+			Name:                 field.Name,
 			TableOID:             field.TableOID,
 			TableAttributeNumber: field.TableAttributeNumber,
 			DataTypeOID:          field.DataTypeOID,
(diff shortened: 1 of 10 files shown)


kirk-anchor:
I was able to reproduce the issue. This fixed it for me. I had to `copy()` each `Name` field as well.

kirk-anchor:
Reverting this commit would be another option https://github.com/jackc/pgx/commit/8d43b38287d2efd59774e43b80ad3bbb7ca33b54

kirk-anchor:
The commit I mentioned reverting, https://github.com/jackc/pgx/commit/8d43b38287d2efd59774e43b80ad3bbb7ca33b54 has another issue, it cuts off the last character of the column name. I pushed a new commit to change FieldDescription.Name to a string and confirmed that I do not get any corrupted names.

pgproto3/row_description.go
```diff
@@ -70,7 +70,7 @@ func (dst *RowDescription) Decode(src []byte) error {
 		if idx < 0 {
 			return &invalidMessageFormatErr{messageType: "RowDescription"}
 		}
-		fd.Name = src[rp : rp+idx]
+		fd.Name = string(src[rp : rp+idx])
```
kirk-anchor, line 73: If you would prefer to keep `Name` as a `[]byte` instead of a `string`, this is all that's needed to fix the bug:
```golang
        name := src[rp : rp+idx]
        fd.Name = make([]byte, len(name))
        copy(fd.Name, name)
```

jackc:
Usability wise, I much prefer it as a `string`. But the reason it is currently `[]byte` is to avoid allocations in a common path. The lowest level `PgConn.ExecParams()` can execute a query in as little as 3 allocations. This change would cause an additional allocation for every column in a result set whether or not the application needed it. Making a copy of the name as a `[]byte` has the same issue.

Are you sure you are not reading from `FieldDescriptions` after you have closed the Rows? The documentation for Rows specifies that that is not allowed.

Even with this change it still may not be safe. The whole `[]pgproto3.FieldDescription` may be reused by future queries.

...

I just read your original post again. You are using `database/sql` not the native pgx interface. But that should work the same way. https://pkg.go.dev/database/sql#Rows.Columns should return an error if the Rows is closed. And if it is not closed it shouldn't be possible to for there to be corruption from another query...

Hmmm...

Have you run your code with the race detector? Maybe there is something else going on?

kirk-anchor:
Thanks for looking into this. I can provide some background. Yes, I am using the `database/sql` interface because I am using `jmoiron/sqlx`. I use sqlx's `GetContext` and `SelectContext`, which does the Scan and Close for me. It uses the column names to populate my struct with the row data returned by the query. This is the error I am hitting https://github.com/jmoiron/sqlx/blob/28212d434cdd87418ecd0cb81173690ce7ac6ab6/sqlx.go#L780

An example of one of the errors is `missing destination name 1ZIn_type in *provider.data`. The sqlx library gets the column names before it scans the first row, so my guess is that the column names are being overwritten by the row data of the same query. If I use pgx v4, I do not have this issue. I tested changing `Name` to a string as well as leaving it as a `[]byte` and copying the bytes, and both of them fixed the issue on pgx v5.

If there's another way it can be fixed, such as preventing the column names in the buffer from being overwritten before the first row is scanned, we could avoid having to copy them. I have reverted to using pgx v4 until the issue is fixed.

jackc:
Do you have an example I can run that demonstrates this issue? I've tried to reproduce with `database/sql` and also tried with `sqlx` and was unable to reproduce it.

jackc:
I just managed to reproduce the issue down at the pgconn level. Still considering if there is a solution that can avoid the allocations.

jackc:
Underlying issue should be fixed in ae65a8007b3078346959e4b80d1f37d3730b92df.

I ended up going for the string conversion, but at the pgconn level rather than the pgproto3 level. This allows proxies and other direct users of pgproto3 to avoid the allocations.

I tested a few ideas to avoid the allocations in general. One was string interning for the column names and the other was caching the entire RowDescription message. But I was unable to get a meaningful improvement so I reverted them both.

If any real world use cases are negatively impacted by this change we can revisit it, but I suspect it will only be noticeable in synthetic microbenchmarks.

kirk-anchor:
Thank you! I tested your fix and I do not get corrupted column names anymore.

PR #1302, opened 2022-09-08
Add DisableNestedTransactions as a configurable flag

minight:
These changes introduce an optional connect flag `disable_nested_transactions` and the config value DisableNestedTransactions (default is False, preserving savepoint behaviour).

Enabling this will cause nested transactions to behave as no-ops (begin, commit, rollback inside an existing txn), relying on the parent transaction to commit/rollback any changes.

The motivation behind these changes were:
* we were observing the performance impacts of having long running nested transactions on our postgres database. (yes, long running transactions is a problem we gradually are fixing)
* our code is structured such that nested helper functions also invoked transactions
* we concluded that allowing the parent transaction to handle the commit/rollback of txns would be acceptable, and we didn't require the savepoint functionality in our codebase.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -37,6 +37,13 @@ type ConnConfig struct {
 	// QueryExOptions.SimpleProtocol.
 	PreferSimpleProtocol bool
 
+	// DisableNestedTransactions will disable savepoint transactions. By default using Begin() while already inside a
+	// transaction will result in a savepoint being created. This disables that behaviour and Begin() while already
+	// inside a transaction behaves as a no-op. This was developed as result of realising that long-running nested
+	// transactions has noticable negative performance impact on postgres.
+	// https://about.gitlab.com/blog/2021/09/29/why-we-spent-the-last-month-eliminating-postgresql-subtransactions/
+	DisableNestedTransactions bool
+
 	createdByParseConfig bool // Used to enforce created by ParseConfig rule.
 }
 
@@ -127,6 +134,9 @@ func ConnectConfig(ctx context.Context, connConfig *ConnConfig) (*Conn, error) {
 //
 //	prefer_simple_protocol
 //		Possible values: "true" and "false". Use the simple protocol instead of extended protocol. Default: false
+//
+//  disable_nested_transactions
+//      Possible values: "true" and "false". Disable savepoint nested txns when executing inside a txn. Default: false
 func ParseConfig(connString string) (*ConnConfig, error) {
 	config, err := pgconn.ParseConfig(connString)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


jackc:
I'm concerned about this option introducing potential foot-guns. I don't see any good way to handle rollback. Either it doesn't roll anything back and risks committing something it shouldn't if the outer `Tx` commits (I think this is what this PR does) or it rolls back the parent transaction. Either one could be dangerous.

> our code is structured such that nested helper functions also invoked transactions

I assume this means the helper functions are receiving an interface rather than a `*pgx.Conn` or `pgx.Tx`. What about doing a type assertion to `Tx` and not calling `Begin()` if you are already in a transaction. You could even make a function like `DoAtomic()` or `EnsureTx` to wrap that whole concept.

And this approach makes obvious that different transaction behavior is in play at the site of the code that is affected by it instead of spooky action at a distance.

minight:
> I'm concerned about this option introducing potential foot-guns. I don't see any good way to handle rollback. Either it doesn't roll anything back and risks committing something it shouldn't if the outer Tx commits (I think this is what this PR does) or it rolls back the parent transaction. Either one could be dangerous.

I think this is a fair concern. I reasoned through this thinking that the parent transaction would catch an error in the child transaction, then rollback to the very beginning, as if there was no nested transaction. I understand that there probably are more nuanced usages where people rely on checkpoints, and as such would expect a potentially different behaviour. 

I'm not sure entirely how it would commit something it shouldn't, as a failed operation in the txn (e.g. bad syntax, or failed constraint), would result in postgres telling us the transaction has failed and has to be rolled back. I'm guessing it could happen though with a transient network error that is only client-sided, then the caller ignoring errors returned from a query, thus continuing on to commit the whole txn? 

> I assume this means the helper functions are receiving an interface rather than a *pgx.Conn or pgx.Tx. What about doing a type assertion to Tx and not calling Begin() if you are already in a transaction. You could even make a function like DoAtomic() or EnsureTx to wrap that whole concept.

I think this is probably a better idea that I can implement without modifying pgx. Thanks!

For completeness, our code pattern is something like:
```golang

type Queryer interface {
	// Begin starts a pseudo nested transaction.
	Begin(ctx context.Context) (pgx.Tx, error)
	// BeginFunc starts a pseudo nested transaction and executes f. If f does not return an err the pseudo nested
	// transaction will be committed. If it does then it will be rolled back.
	BeginFunc(ctx context.Context, f func(pgx.Tx) error) (err error)
	Query(ctx context.Context, sql string, args ...interface{}) (pgx.Rows, error)
	QueryRow(ctx context.Context, query string, args ...interface{}) pgx.Row
	Exec(ctx context.Context, sql string, arguments ...interface{}) (commandTag pgconn.CommandTag, err error)
}

func helper(db iface.Queryer) error {
	return db.BeginFunc(context.TODO(), func(tx pgx.Tx) error {
		if err := foo(tx); err != nil {
			return errors.Wrap(err, "failed to help")
		}
        })
}

func parent(db iface.Queryer) error {
	return db.BeginFunc(context.TODO(), func(tx pgx.Tx) error {
		if err := helper(tx); err != nil {
			return errors.Wrap(err, "failed to parent")
		}
        })
}
```

In this case, we avoid the footgun you describe, as we have a pretty simple usage of transactions, and don't rely on savepoints for any application behaviour or guarantees.

jackc:
👍

PR #1314, opened 2022-09-24
Remove pgerrcode from the README

henvic:
Now that it was merged into the pgconn package.


henvic:
Oops. Disregard this entirely.

PR #1315, opened 2022-09-24
Rewrite context watcher to use single background goroutine

amenzhinsky:
This removes the need to start a new gouroutine each query supplied with a non-`context.Background` context, instead it maintains a background routine that handles all `Watch` requests.

This can save some cpu cycles for applications that pass down `net/http.(*Request).Context` or grpc-generated Context to every single sql query.

Here is `benchcmp` for `ctxwatch` package:

```
benchmark                                  old ns/op     new ns/op     delta
BenchmarkContextWatcherUncancellable-8     27.3          13.6          -50.18%
BenchmarkContextWatcherCancelled-8         599           454           -24.20%
BenchmarkContextWatcherCancellable-8       420           319           -24.04%

benchmark                                  old allocs     new allocs     delta
BenchmarkContextWatcherUncancellable-8     0              0              +0.00%
BenchmarkContextWatcherCancelled-8         4              3              -25.00%
BenchmarkContextWatcherCancellable-8       1              0              -100.00%

benchmark                                  old bytes     new bytes     delta
BenchmarkContextWatcherUncancellable-8     0             0             +0.00%
BenchmarkContextWatcherCancelled-8         208           176           -15.38%
BenchmarkContextWatcherCancellable-8       32            0             -100.00%
```

Unfortunately go benchmarks don't include info about goroutines used since they're handled by the runtime.

### DIFF
--- a/pgconn/internal/ctxwatch/context_watcher.go
+++ b/pgconn/internal/ctxwatch/context_watcher.go
@@ -14,6 +14,7 @@ type ContextWatcher struct {
 	watchInProgress uint32
 	watchChan       chan context.Context
 	unwatchChan     chan struct{}
+	watchDoneChan   chan struct{}
 }
 
 // NewContextWatcher returns a ContextWatcher. onCancel will be called when a watched context is canceled.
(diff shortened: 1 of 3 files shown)


jackc:
I have a few concerns about this change.

I'm not sure atomics as used provide the same synchronization that the mutex did. Maybe it does... But it is definitely harder to reason about. My experience with atomics has led me to avoid them unless there is a very large performance gain. I'd expect that this is a low contention lock where the mutex would be pretty fast.

But regarding the persistent background goroutine overall -- have you benchmarked how this would affect entire queries? I went down this path myself several years ago and found at best no significant improvements. IIRC, the additional synchronization cost offset the savings from starting a goroutine. Maybe things have changed since then but unless there are major benefits I'd rather stay with the simpler approach we have now.

Issue #1335, opened 2022-10-17
nil pointer dereference on upgrading from 5.0.1 to 5.0.3


starudream:
**To Reproduce**

`docker run -d --name pg -e POSTGRES_PASSWORD=p1ssw0rd -p 5432:5432 postgres`

```go
package main

import (
	"context"
	"fmt"

	"github.com/jackc/pgx/v5"
	"github.com/lib/pq"
)

func main() {
	conn, err := pgx.Connect(context.Background(), "postgres://postgres:p1ssw0rd@127.0.0.1:5432/postgres?sslmode=disable&default_query_exec_mode=cache_describe&timezone=Asia/Shanghai")
	if err != nil {
		panic(err)
	}
	defer conn.Close(context.Background())

	_, err = conn.Exec(context.Background(), "CREATE TABLE IF NOT EXISTS test (status SMALLINT);")
	if err != nil {
		panic(err)
	}

	rows, err := conn.Query(context.Background(), "SELECT status FROM test WHERE status = ANY($1);", pq.Array([]int16{2}))
	if err != nil {
		panic(err)
	}
	defer rows.Close()

	for rows.Next() {
		var status int
		err = rows.Scan(&status)
		if err != nil {
			panic(err)
		}
		fmt.Printf("%d\n", status)
	}
}
```

**Actual behavior**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x98 pc=0x102fa2588]

goroutine 1 [running]:
github.com/jackc/pgx/v5/pgtype.TryWrapStructEncodePlan({0x0?, 0x0})
	/go/pkg/mod/github.com/jackc/pgx/v5@v5.0.3/pgtype/pgtype.go:1803 +0x88
github.com/jackc/pgx/v5/pgtype.(*Map).PlanEncode(0xc0000d64e0, 0x15, 0x1, {0x0?, 0x0})
	/go/pkg/mod/github.com/jackc/pgx/v5@v5.0.3/pgtype/pgtype.go:1321 +0x2b8
github.com/jackc/pgx/v5/pgtype.(*ArrayCodec).PlanEncode(0xc0000c4270, 0xc0000d64e0, 0x3ed, 0x1, {0x1037df3e0, 0xc0000bbcf8})
	/go/pkg/mod/github.com/jackc/pgx/v5@v5.0.3/pgtype/array_codec.go:61 +0xa8
[...]

PR #1336, opened 2022-10-17
Check nil pointer before reflect.TypeOf

starudream:
#1335


deleted:
Just wanted to call attention to `pgtypes.TryWrapDerefPointerEncodePlan()`, which is also affected by this nil pointer issue.  There may be other `Try...EncodePlan()` functions that need to be fixed as well.

jackc:
@starudream Sorry, I was working through my emails oldest to newest and didn't see your PR until I had already fixed the issue.

@deleted AFAIK, `TryWrapDerefPointerEncodePlan()` should be fixed as of f2e7c8144d5e997517767355619215cf7a91189d.

PR #1337, opened 2022-10-17
fix: bump text package for CVE-2022-32149

TheJefe:
https://security.snyk.io/vuln/SNYK-GOLANG-GOLANGORGXTEXTLANGUAGE-3043869


jackc:
pgx isn't using the vulnerable function, but I suppose it is worth placating snik.

PR #1344, opened 2022-10-18
Fix RowToStructByPos on structs with multiple anonymous sub-structs

### LINKED ISSUE #1343: Destination fields are wrong when using RowToStructByPos on a struct with multiple nested anonymous structs
**Describe the bug**
When using `RowToStructByPos` with a `struct` that contains more than one anonymous sub-`struct`s, we get the error "`number of field descriptions must equal number of destinations`" even with a correct number of fields.

**To Reproduce**

Create a simple table with three int columns and a couple rows:
```bash
sudo -u postgres createuser dummypgx
sudo -u postgres createdb dummypgx
sudo -u postgres psql dummypgx
```

Then:
```sql
ALTER USER dummypgx WITH ENCRYPTED PASSWORD 'dummypgx';
GRANT ALL PRIVILEGES ON DATABASE dummypgx TO dummypgx;
CREATE TABLE t ( a int, b int, c int );
INSERT INTO t (a, b, c) VALUES (1, 1, 1), (2, 2, 2), (3, 3, 3) ;
GRANT SELECT ON t TO dummypgx;
```

Code:
```go
package main

import (
	"context"
	"github.com/jackc/pgx/v5"
	"log"
)

type AHolder struct{ A int }
type BHolder struct{ B int }
type CHolder struct{ C int }

type ABC struct {
	AHolder
	BHolder
	CHolder
}

func main() {
	ctx := context.Background()

	conn, err := pgx.Connect(ctx,
		"user=dummypgx password=dummypgx host=localhost port=5432 dbname=dummypgx")
	if err != nil {
		panic(err)
	}
	defer conn.Close(ctx)

	rows, err := conn.Query(ctx, "SELECT * FROM t;")
	if err != nil {
		panic(err)
	}
	defer rows.Close()

	abcs, err := pgx.CollectRows(rows, pgx.RowToStructByPos[ABC])
	if err != nil {
		panic(err)
	}

	for _, abc := range abcs {
		log.Printf("%+v\n", abc)
	}
}
```
I created a [minimal repo](https://github.com/bfontaine/dummypgx) with the code. [This is the diff](https://github.com/bfontaine/dummypgx/commit/dca031d2e7436cc0fe00a0cec94e8af567d03048) between a version that work and one that doesn’t.

**Expected behavior**
It should run and print:
```
2022/10/18 18:05:11 {AHolder{A:1} BHolder{B:1} CHolder{C:1}}
2022/10/18 18:05:11 {AHolder{A:2} BHolder{B:2} CHolder{C:2}}
2022/10/18 18:05:11 {AHolder{A:3} BHolder{B:3} CHolder{C:3}}
```

**Actual behavior**
it fails with:
```
panic: number of field descriptions must equal number of destinations, got 3 and 7

goroutine 1 [running]:
main.main()
	/.../main.go:37 +0x256
exit status 2
```

**Version**
 - Go: `go version go1.19.2 linux/amd64`
 - PostgreSQL: `PostgreSQL 12.12 (Ubuntu 12.12-0ubuntu0.20.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit`
 - pgx: `v5.0.3`

**Additional context**
In my real-world example, the number of destinations found by pgx differ based on the order of the fields. My parent struct has three child-structs: T with 5 fields, M with 4 and N with 2 (total: 11). With `struct { T M N }` pgx finds 30 destination fields, but with `struct { T N M }` it finds only 28 fields. With `struct { T M }` it finds 16 fields (instead of 9); with `struct { T N }` it finds 16 fields (instead of 7).

bfontaine:
Fixes #1343.

See the issue for a description of the issue.

The new test fails on the current `master`, but succeeds with the patch.

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -530,9 +530,9 @@ func (rs *positionalStructRowScanner) appendScanTargets(dstElemValue reflect.Val
 	for i := 0; i < dstElemType.NumField(); i++ {
 		sf := dstElemType.Field(i)
 		if sf.PkgPath == "" {
-			// Handle anoymous struct embedding, but do not try to handle embedded pointers.
+			// Handle anonymous struct embedding, but do not try to handle embedded pointers.
 			if sf.Anonymous && sf.Type.Kind() == reflect.Struct {
-				scanTargets = append(scanTargets, rs.appendScanTargets(dstElemValue.Field(i), scanTargets)...)
+				scanTargets = rs.appendScanTargets(dstElemValue.Field(i), scanTargets)
 			} else {
 				scanTargets = append(scanTargets, dstElemValue.Field(i).Addr().Interface())
 			}
--- a/rows_test.go
+++ b/rows_test.go
@@ -354,6 +354,34 @@ func TestRowToStructByPosEmbeddedStruct(t *testing.T) {
 	})
 }
 
+func TestRowToStructByPosMultipleEmbeddedStruct(t *testing.T) {
+	type Sandwich struct {
+		Bread string
+		Salad string
+	}
+	type Drink struct {
+		Ml int
+	}
+
+	type meal struct {
+		Sandwich
+		Drink
+	}
+
+	defaultConnTestRunner.RunTest(context.Background(), t, func(ctx context.Context, t testing.TB, conn *pgx.Conn) {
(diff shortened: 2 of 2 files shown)


jackc:
Thanks! LGTM.

PR #1363, opened 2022-10-30
feat: support row to struct by tag or use snake case name for default field name

KennyChenFight:
Hi, @jackc 
I think use `pgx.RowToStructByPos` to scan struct is kind of inconvenient：when query `select *`, struct field order should follow real db schema order.
So, I think we can support `pgx.RowToStructByTag`, and use snake case name as default field name, not use position to scan.
It may have some performance problem, but it's useful.


KennyChenFight:
Hi @jackc 
I fixed the test. Could you approve me to run workflows ?

jackc:
Well, this is an unusual situation. Two PRs for the same feature before I have a chance to review either. See #1366. 🤔

KennyChenFight:
> Well, this is an unusual situation. Two PRs for the same feature before I have a chance to review either. See #1366. 🤔

Oh, but I think my spec is a little different since I use struct tag and snake case.
Since I used to use go-pg when I migrate to pgx, I think scan by struct tag and snake case is convenient.

jackc:
I ended up going for #1366. I wanted to avoid the external dependency for the snake case mapping, and I wasn't really sure about the magic name mapping overall.

PR #1366, opened 2022-11-03
implement `RowToStructByName` and `RowToAddrOfStructByName`

pashagolub:


jackc:
Well, this is an unusual situation. Two PRs for the same feature before I have a chance to review either. See #1363. 🤔

pashagolub:
Wow :-) It's up to you what to choose. Just a reminder, if one uses `db/sql` mappers or migrates from `db/sql` then `db` tag used by default.
But I'm sure we can have both. That's exactly why I used `const structTagKey = "db"`. Just make that global and allow user to specify what exactly tag to use

pashagolub:
Plus, I'm not sure if you want to add another dependency `iancoleman/strcase`

jackc:
This has been manually rebased and merged. I added a bit more documentation and tests.

I thought about relaxing the requirement that all public struct fields must be filled, but elected to leave it the way it is and see what the user experience is like. It can always be relaxed later.

PR #1372, opened 2022-11-07
Set tracer in 1 LoC

Imm0bilize:
Added a global private variable (_defaultQueryTracer_) and setter for this variable (_SetDefaultTracer_) for more convenient work with tracing. The old method remains working.

Example of the new method:
```golang
ctx, cancel := context.WithTimeout(context.Background(), time.Second * 10)
defer cancel()


// allows you set your tracer (custom log, otel, etc.) in 1 LoC
//    ↓           ↓            ↓           ↓ 
pgx.SetDefaultTracer(tracing.NewPgxTracer())

connect, err := pgx.Connect(ctx, "")
if err != nil{
    log.Fatal(err)
}
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -181,6 +181,7 @@ func ParseConfig(connString string) (*ConnConfig, error) {
 
 	connConfig := &ConnConfig{
 		Config:                   *config,
+		Tracer:                   defaultQueryTracer,
 		createdByParseConfig:     true,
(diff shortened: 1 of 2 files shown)


jackc:
Sorry, I'd prefer to avoid adding global state unless there is no other way. In this case, the same functionality could be provided by an application function that creates connections or configs.

PR #1377, opened 2022-11-10
Add DeallocateAll to pgx.Conn to clear prepared statement cache

bodokaiser:
Please provide feedback to the PR. Related [discussion](https://github.com/jackc/pgx/discussions/1370).

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -326,6 +326,13 @@ func (c *Conn) Deallocate(ctx context.Context, name string) error {
 	return err
 }
 
+// Deallocate all released prepared statements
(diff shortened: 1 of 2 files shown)


jackc:
The statement cache would still have references to the prepared statements removed by `deallocate all`. You need to reset `c.statementCache` as well as explicitly prepared statements.

The documentation should also clarify that it deallocates explicitly prepared statements and resets the statement cache.

Also, even though it is not strictly necessary, it probably should reset `c.descriptionCache`. Then this function could be used when the database structure changes.

PR #1379, opened 2022-11-13
perf: use uuid module

S-YOU:
I was trying to benchmark some projects and figure it encodeUUID is using too much memory.

image

Looks like fmt.Sprintf is not performant and output of uuid string has fixed length.
I was trying to roll my own but looks google/uuid doing best on this purpose. 
Also it is compatible with pgx datatype which is [16]byte.

Benchmark results is here (I am not included those, since I am not sure, I should even add google/uuid to the pgx)

```
BenchmarkEncodeUUID-16                   1000000              2482 ns/op             552 B/op         21 allocs/op
BenchmarkGoogleEncodeUUID-16             2829258               417.9 ns/op           144 B/op          3 allocs/op
```

And the results are good. I am personally ok with custom data type that pgx provided, so I don't mind even if you not accepting this kind of PR.

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/jackc/pgx/v5
 go 1.18
 
 require (
+	github.com/google/uuid v1.3.0
 	github.com/jackc/pgpassfile v1.0.0
 	github.com/jackc/pgservicefile v0.0.0-20200714003250-2b9c44734f2b
 	github.com/jackc/puddle/v2 v2.1.2
--- a/go.sum
+++ b/go.sum
@@ -2,12 +2,12 @@ github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ3
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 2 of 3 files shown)


jackc:
Have you tried using https://github.com/vgarvardt/pgx-google-uuid ?

S-YOU:
Thanks @jackc , I didn't know that package, looks like it is already using google/uuid, so problem should be fixed there.

S-YOU:
I just tried pgx-google-uuid, but pgtype.encodeUUID is still getting called. I am doing something wrong?  @jackc 
I followed usage instructions on README.

S-YOU:
Also tried to benchmark DecodeUUID (Parse) method, and Google's one is kind of faster with no memory allocation.
Not sure anyone needed such micro benchmark, but I like fast stuff, so just sharing benchmark results. 

```
BenchmarkParseUUID-16                    2271021           541.7 ns/op        96 B/op          3 allocs/op
BenchmarkParseGoogleUUID-16             14152162            83.26 ns/op        0 B/op          0 allocs/op
```
(not that one op has 4 UUIDs in inner loop, so actual values supposed to be 4 times faster)

jackc:
Also, I'm guessing you using database/sql? I think this entire path doesn't happen with the native pgx interface.

My guess is [github.com/vgarvardt/pgx-google-uuid](https://github.com/vgarvardt/pgx-google-uuid) could implement `DecodeDatabaseSQLValue()`. It appears that method is delegated to the pgx UUID code.

Beyond that, I don't mind merging a performance improvement to the built-in `encodeUUID()`, but not one that adds a dependency. One of the big improvements of v5 over v4 was shedding a bunch of dependencies. I don't want to regress there.

S-YOU:
> Also, I'm guessing you using database/sql? I think this entire path doesn't happen with the native pgx interface.

I am using pgxpool, but I could be wrong on usage.

> Beyond that, I don't mind merging a performance improvement to the built-in encodeUUID(), but not one that adds a dependency. One of the big improvements of v5 over v4 was shedding a bunch of dependencies. I don't want to regress there.

I agree. fewer dependencies is better. I tried to write similar encode function in pgx directly, but I checked google uuid source, and I can't unseen and don't want to copy paste their code here, so was sharing my two cents. I'll move on.

PR #1384, opened 2022-11-17
removed unnecessary name argument from DeallocateAll

bodokaiser:
Hey Jack,

In our last commit, I forgot to remove the `name` argument (copied from `Deallocate`), which does not make any sense for `DeallocateAll`.

This PR removes it.

Best wishes,
Bodo


jackc:
Whoops. I should of noticed that too.

PR #1398, opened 2022-11-25
perf(tx): use strings.Builder 

kscooo:
use strings.Builder to avoid the overhead of []byte -> string conversion

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/jackc/pgx/v5/pgconn"
 )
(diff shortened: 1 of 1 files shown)


iClaus21:
Hi @jackc , hi @kscooo.
I think that these changes don't carry any useful and only Worse performance relative to my previous solution.
To confirm my words I attach the written benchmarks.

[database_test.txt](https://github.com/jackc/pgx/files/10176313/database_test.txt)

Short results are here:
`BenchmarkNew
BenchmarkNew-12    	20182904	        59.25 ns/op	      64 B/op	       1 allocs/op`
`BenchmarkOld
BenchmarkOld-12    	26423922	        45.38 ns/op	      64 B/op	       1 allocs/op`

As a result, performance degrated from 45 ns\op to 60 ns\op (~33%).
I use Go version 1.19.3.

What do you think about these changes?

Issue #1388, opened 2022-11-22
Cannot unmarshal string into Go struct field * of type pgtype.Range[github.com/jackc/pgx/v5/pgtype.Date]


emaborsa:
I have problems to scan json objects which include daterange types:

Let's assume these tables:
```
CERATE TABLE first_table (
  id int PRIMARY KEY,
  name text
);

CERATE TABLE second_table (
  id int PRIMARY KEY,
  title text,
  period daterange NOT NULL,
  a_id int NOT NULL REFERENCES first_table (id)
);
```

The structs are 
```
type A struct {
  ID int
  Name *string
  SomeBs []B
}

type B struct {
  ID int
  Title *string
  Period  pgtype.Range[pgtype.Date]
}
```

The query:
```
SELECT 
    a.id,
    a.name,
    bs.*
FROM first_table a
LEFT JOIN LATERAL (
	SELECT 
	COALESCE(json_agg(json_build_object(
		'id', id,
		'title', title,
		'period', period)), '[]')
	FROM second_table b 
	WHERE a.id = b.a_id
) AS bs ON true 
```

I get the error `can't scan into dest[2]: json: cannot unmarshal string into Go struct field B.Period of type pgtype.Range[github.com/jackc/pgx/v5/pgtype.Date]`

As writte [here](https://github.com/jackc/pgx/issues/1374), scanning it without json_agg and json_build_object, it works.
If for testing purpose I changed the type B.Period to string, the query works and the value is something like `[2020-02-01,2023-02-03)`. I tried to change the type to `date_range.DateRange` or `pgtype.Range[differentTypes]`, nothing worked.

How can I scan the values in `date_range.DateRange` or `pgtype.Range[pgtype.Date]`? I would prefer the latter.

jackc:
pgtype.Range doesn't implement `json.Unmarshaler`. No reason it shouldn't -- just no one has implemented it yet. Happy to merge if someone does.

acim:
I have started working on this.

https://github.com/acim/pgx/tree/pgtype-range-json

PR #1408, opened 2022-12-01
#1388 pgtype.Range json marshal and unmarshal

acim:


jackc:
Is it possible to reuse some more of the existing text encoding and parsing code? I haven't looked at it in detail, but naively I would have assumed that `MarshalJSON` would basically delegate to the existing text format encoding code and `UnmarshalJSON` would basically delegate to the existing text format parsing code.

acim:
I could reuse `parseUntypedTextRange` in `UnmarshalJSON`, but reusing `Encode` method of `encodePlanRangeCodecRangeValuerToText` struct looks not an easy thing to do because it's using type OID's for which I don't see the way to get in `func (src Range[T]) MarshalJSON() ([]byte, error)` since T in Range is of type any. There are no further bounds, so getting an OID seems impossible.

```
		lowerPlan := plan.m.PlanEncode(plan.rc.ElementType.OID, TextFormatCode, lower)
		if lowerPlan == nil {
			return nil, fmt.Errorf("cannot encode %v as element of range", lower)
		}

		buf, err = lowerPlan.Encode(lower, buf)
		if err != nil {
			return nil, fmt.Errorf("failed to encode %v as element of range: %v", lower, err)
		}
		if buf == nil {
			return nil, fmt.Errorf("Lower cannot be NULL unless LowerType is Unbounded")
		}
```
and the similar piece of code for upper bound in case of json could be simple replaced by marshal. So the only useful code from there would be before and after these mentioned.

acim:
Actually, I found a way to reuse the `encodePlanRangeCodecRangeValuerToText.Encode` but I have to modify it to:

```
type encodePlanRangeCodecRangeValuerToText struct {
	rc Codec
	m  interface {
		PlanEncode(oid uint32, format int16, value any) EncodePlan
	}
}
```

I implemented tests for Range[Date] and Range[Numeric] and then found out that Numeric didn't have json.Unmarshaler implementation so I did that also. I couldn't run all tests because lot of them require database connection. @jackc, could you please allow CI to run for this PR?

rubiagatra:
Hi @acim is this PR still relevant? Can we close it?
cc: @jackc

acim:
No, you can close it. :)

PR #1409, opened 2022-12-01
Config cleanup

JCPrice0024:
In pgconn/config.go I found a function with a complexity score of 44 and reduced it to under 15. I did this by breaking up the complex if statements, inside of the function, into smaller helper functions each with comments. I also made a sonarqube properties file and added .scannerwork to gitignore.

here is the before and after sonarqube scan:
image
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ _testmain.go
 
 .envrc
 /.testdb
+.scannerwork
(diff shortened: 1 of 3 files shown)


jackc:
Sorry, but IMO this is not an improvement. One large function that can be read top to bottom can be easier to understand than multiple smaller functions where one must jump around in the code to follow the execution flow.

But in particular this is the case here where the smaller functions are impure (i.e. they are mutating the config argument). This also obscures any temporal dependencies that exists in the process (i.e. steps that depend on previous steps).

PR #1412, opened 2022-12-05
Add QueryExecMode to CopyFrom

alejandrodnm:
### Use DefaultQueryExecMode in CopyFrom 
 
 CopyFrom had to create a prepared statement to get the OIDs of the data
types that were going to be copied into the table. Every COPY operation
required an extra round trips to retrieve the type information. There
was no way to customize this behavior.

By leveraging the QueryExecMode feature, like in `Conn.Query`, users can
specify if they want to cache the prepared statements, execute
them on every request (like the old behavior), or bypass the prepared
statement relying on the pgtype.Map to get the type information.

The `QueryExecMode` behave exactly like in `Conn.Query` in the way the
data type OIDs are fetched, meaning that:

- `QueryExecModeCacheStatement`: caches the statement.
- `QueryExecModeCacheDescribe`: caches the statement and assumes they do
  not change.
- `QueryExecModeDescribeExec`: gets the statement description on every
  execution. This is like to the old behavior of `CopyFrom`.
- `QueryExecModeExec` and `QueryExecModeSimpleProtocol`: maintain the
  same behavior as before, which is the same as `QueryExecModeDescribeExec`.
  It will keep getting the statement description on every execution

The `QueryExecMode` can only be set via
`ConnConfig.DefaultQueryExecMode`, unlike `Conn.Query` there's no
support for specifying the `QueryExecMode` via optional arguments
in the function signature.

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -85,6 +85,7 @@ type copyFrom struct {
 	columnNames   []string
 	rowSrc        CopyFromSource
 	readerErrChan chan error
+	mode          QueryExecMode
 }
 
 func (ct *copyFrom) run(ctx context.Context) (int64, error) {
@@ -105,9 +106,29 @@ func (ct *copyFrom) run(ctx context.Context) (int64, error) {
 	}
 	quotedColumnNames := cbuf.String()
 
-	sd, err := ct.conn.Prepare(ctx, "", fmt.Sprintf("select %s from %s", quotedColumnNames, quotedTableName))
-	if err != nil {
-		return 0, err
+	var sd *pgconn.StatementDescription
+	switch ct.mode {
+	case QueryExecModeExec, QueryExecModeSimpleProtocol:
+		// These modes don't support the binary format. Before the inclusion of the
+		// QueryExecModes, Conn.Prepare was called on every COPY operation to get
+		// the OIDs. These prepared statements were not cached.
+		//
+		// Since that's the same behavior provided by QueryExecModeDescribeExec,
+		// we'll default to that mode.
+		ct.mode = QueryExecModeDescribeExec
+		fallthrough
+	case QueryExecModeCacheStatement, QueryExecModeCacheDescribe, QueryExecModeDescribeExec:
+		var err error
+		sd, err = ct.conn.getStatementDescription(
+			ctx,
+			ct.mode,
+			fmt.Sprintf("select %s from %s", quotedColumnNames, quotedTableName),
+		)
+		if err != nil {
+			return 0, fmt.Errorf("statement description failed: %w", err)
+		}
+	default:
+		return 0, fmt.Errorf("unknown QueryExecMode: %v", ct.mode)
 	}
 
 	r, w := io.Pipe()
@@ -208,6 +229,7 @@ func (c *Conn) CopyFrom(ctx context.Context, tableName Identifier, columnNames [
 		columnNames:   columnNames,
 		rowSrc:        rowSrc,
 		readerErrChan: make(chan error),
+		mode:          c.config.DefaultQueryExecMode,
 	}
 
 	return ct.run(ctx)
--- a/copy_from_test.go
+++ b/copy_from_test.go
@@ -14,6 +14,129 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestConnCopyWithAllQueryExecModes(t *testing.T) {
+	for _, mode := range pgxtest.AllQueryExecModes {
+		t.Run(mode.String(), func(t *testing.T) {
+			t.Parallel()
+
+			cfg := mustParseConfig(t, os.Getenv("PGX_TEST_DATABASE"))
+			cfg.DefaultQueryExecMode = mode
+			conn := mustConnect(t, cfg)
+			defer closeConn(t, conn)
+
+			mustExec(t, conn, `create temporary table foo(
+			a int2,
+			b int4,
+			c int8,
+			d text,
+			e timestamptz
+		)`)
+
+			tzedTime := time.Date(2010, 2, 3, 4, 5, 6, 0, time.Local)
+
+			inputRows := [][]any{
+				{int16(0), int32(1), int64(2), "abc", tzedTime},
+				{nil, nil, nil, nil, nil},
+			}
+
+			copyCount, err := conn.CopyFrom(context.Background(), pgx.Identifier{"foo"}, []string{"a", "b", "c", "d", "e"}, pgx.CopyFromRows(inputRows))
+			if err != nil {
+				t.Errorf("Unexpected error for CopyFrom: %v", err)
+			}
+			if int(copyCount) != len(inputRows) {
+				t.Errorf("Expected CopyFrom to return %d copied rows, but got %d", len(inputRows), copyCount)
+			}
+
+			rows, err := conn.Query(context.Background(), "select * from foo")
+			if err != nil {
+				t.Errorf("Unexpected error for Query: %v", err)
+			}
+
+			var outputRows [][]any
+			for rows.Next() {
+				row, err := rows.Values()
+				if err != nil {
+					t.Errorf("Unexpected error for rows.Values(): %v", err)
+				}
+				outputRows = append(outputRows, row)
+			}
+
+			if rows.Err() != nil {
+				t.Errorf("Unexpected error for rows.Err(): %v", rows.Err())
+			}
+
+			if !reflect.DeepEqual(inputRows, outputRows) {
+				t.Errorf("Input rows and output rows do not equal: %v -> %v", inputRows, outputRows)
+			}
+
+			ensureConnValid(t, conn)
+		})
+	}
+}
+
+func TestConnCopyWithKnownOIDQueryExecModes(t *testing.T) {
+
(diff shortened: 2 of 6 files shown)


cristaloleg:
Hi, thanks for the PR, are there any performance improvements in benchmarks?

jackc:
I like the direction this is going, but I am concerned about the change in the `CopyFrom()` signature.  It won't directly break any calling code, but it does break any interfaces defined with `Conn`, `Tx`, etc. in mind. e.g.

```go
type DB interface {
	Query(ctx context.Context, sql string, optionsAndArgs ...interface{}) (pgx.Rows, error)
	QueryRow(ctx context.Context, sql string, optionsAndArgs ...interface{}) pgx.Row
	CopyFrom(ctx context.Context, tableName pgx.Identifier, columnNames []string, rowSrc pgx.CopyFromSource) (int64, error)
}
```

An application or library using interface `DB` would break.

Would it be sufficient to use the default query exec mode without the ability to override? AFAIK, overriding the mode per query is extremely rare. I only included it on `Query`, `QueryRow`, and `Exec` for completeness and testing.

If it really needs to be controlled per call, perhaps a new optional interface could be created that `rowSrc` could implement. This interface could have a method to get the query exec mode. A wrapper struct could be created that wrapped the original `CopyFromSource` and added the new method.

alejandrodnm:
@jackc I really need to override it to avoid the the prepare statement and the cache 😅 

We are using pgx in [timescale/promscale](https://github.com/timescale/promscale) to ingest metrics and traces, we insert the data using COPY. Basically we have a table per metric, and our COPY operations are always of the same 3 columns. 

We want to get rid of the extra roundtrip per copy. In our use case the cache would be a lot of unnecessary memory since we are always sending (time, int64 and float64), which the Type map resolves correctly.

> If it really needs to be controlled per call, perhaps a new optional interface could be created that rowSrc could implement. This interface could have a method to get the query exec mode. A wrapper struct could be created that wrapped the original CopyFromSource and added the new method.

I'll try to give ☝️ a shot.

alejandrodnm:
I'm going to try our use case with allowing setting the Exec mode only via `ConnConfig.DefaultQueryExecMode`.

I reverted the signature changes to `CopyFrom`.

tracelog/tracelog_test.go
```diff
@@ -217,7 +217,7 @@ func TestLogCopyFrom(t *testing.T) {
 		return config
 	}
 
-	pgxtest.RunWithQueryExecModes(context.Background(), t, ctr, nil, func(ctx context.Context, t testing.TB, conn *pgx.Conn) {
+	pgxtest.RunWithQueryExecModes(context.Background(), t, ctr, pgxtest.KnownOIDQueryExecModes, func(ctx context.Context, t testing.TB, conn *pgx.Conn) {
```
alejandrodnm, line 220: @jackc this is the tests that was making the CI fail. It only fails on cockroachDB, and when retrieving the types from the typeMap instead of a prepared statement.

When using a prepare statement it fails because there's a type mismatch between the input which is a `string` and the OID from the prepared statement. 

When using the types from the typeMap, the OID matches the input, but the input doesn't match the table definition.  but no error is returned.

copy_from.go
```diff
@@ -105,9 +106,17 @@ func (ct *copyFrom) run(ctx context.Context) (int64, error) {
...
-		return 0, err
+	var oids []uint32
+	switch ct.mode {
+	case QueryExecModeCacheStatement, QueryExecModeCacheDescribe, QueryExecModeDescribeExec:
+		var err error
+		oids, err = ct.oidsFromStatementDescription(ctx, quotedColumnNames, quotedTableName)
+		if err != nil {
+			return 0, err
+		}
+	case QueryExecModeExec, QueryExecModeSimpleProtocol:
```
jackc, line 117: I would have expected this to cause test failures, so maybe I'm missing something, but I don't see how this will work for `QueryExecModeExec` and `QueryExecModeSimpleProtocol`. `oids` will be `nil`.

tracelog/tracelog_test.go
jackc: Maybe this has something to do with my comment above.

alejandrodnm: The CI hadn't finished and I was seeing the previous error 🤦

copy_from.go
```diff
@@ -166,20 +175,32 @@ func (ct *copyFrom) run(ctx context.Context) (int64, error) {
...
 		if len(values) != len(ct.columnNames) {
 			return false, nil, fmt.Errorf("expected %d values, got %d values", len(ct.columnNames), len(values))
 		}
 
+		if oids == nil {
+			oids, err = ct.oidsFromValues(values)
+			if err != nil {
+				return false, nil, fmt.Errorf("couldn't get OIDs from values: %w", err)
+			}
+		}
```
alejandrodnm, line 199: Here I check if OIDs were pass as arguments, if they weren't, which is the case for `case QueryExecModeExec, QueryExecModeSimpleProtocol` then It defaults to matching the types of the values against the `pgtype.Map`. I need to do this here because I can't peek at the values from `rowSrc` without moving the iterator. 

I'll add some comments for this path to make it more explicit.

copy_from.go
alejandrodnm: I need to pass `oids = nil` so that `buildCopyBuf` looks for the OIDs from the values. 

Here's more context https://github.com/jackc/pgx/pull/1412/files#r1045710050 I'll add some extra comments to avoid confusion.

copy_from_test.go
```diff
@@ -14,6 +14,129 @@ import (
...
+			}
+
+			ensureConnValid(t, conn)
+		})
+	}
+}
+
+func TestConnCopyWithKnownOIDQueryExecModes(t *testing.T) {
+
+	for _, mode := range pgxtest.KnownOIDQueryExecModes {
```
alejandrodnm, line 79: This tests is only for the `pgxtest.KnownOIDQueryExecModes` because we can't resolved between `varchar` and `text`, or between `date` and `timestampz`, with the exec modes that rely on the `pgtype.Map`

copy_from.go
alejandrodnm: I added the extra comments.

jackc: Ah. I don't think this will work for types that are not exactly 1 to 1 for Go to PostgreSQL. e.g. a Go `time.Time` and a PostgreSQL `date` or Go `int64` and PostgreSQL `smallint`.

It previously would run a prepare even in `QueryExecModeExec` and `QueryExecModeExec` to avoid these problems. I don't think that can safely be removed.

alejandrodnm: You mean because of: https://github.com/jackc/pgx/blob/e58381ac9409172a674f76c1801c3c606ac41b42/extended_query_builder.go#L148-L160

I just need a way to specify the OIDs for the copy operation. I was thinking about:

> If it really needs to be controlled per call, perhaps a new optional interface could be created that rowSrc could implement. This interface could have a method to get the query exec mode. A wrapper struct could be created that wrapped the original CopyFromSource and added the new method.

It would be ideal if I can extend the `CopyFromSource` with a method that returns the OIDs of the values, but that would mean changing the interface. We could try the wrapper that you mentioned, but then at some point in the code I'll need to cast the `CopyFromSource` to check if it's a wrapper struct, and then call the function, which might not be the best solution. Something like changing 

https://github.com/jackc/pgx/pull/1412/files#diff-89f7d5816796e2870b38b63109c1f9a3f9a382420f245dffcada921d3878760cR109

For:

```go
var oids []uint32
w, ok := (ct.rowSrc).(withOIDsWrapper)
if ok {
         oids = w.OIDs()
} else {
	switch ct.mode {
	case QueryExecModeExec, QueryExecModeSimpleProtocol:
                 ct.mode = QueryExecModeDescribeExec
                 fallthrough
	case QueryExecModeCacheStatement, QueryExecModeCacheDescribe, QueryExecModeDescribeExec:
		var err error
		oids, err = ct.oidsFromStatementDescription(ctx, quotedColumnNames, quotedTableName)
		if err != nil {
			return 0, err
		}
	default:
		return 0, fmt.Errorf("unknown QueryExecMode: %v", ct.mode)
}
```

If this is something that you don't want to add, I can modify the PR to always use the defaultExecMode and to handle the cases of `QueryExecModeExec` and `QueryExecModeSimpleProtocol`, default to use `QueryExecModeDescribeExec` because it's the same behaviour as before. For our use case, we could just construct our own CopyFrom using `pgconn.CopyFrom` directly.

Let me know what you think, or if you have other ideas. Thanks :)

jackc: I think this is the correct approach. i.e. `QueryExecModeDescribeExec`, `QueryExecModeExec`, `QueryExecModeSimpleProtocol` would all call `Prepare()` like they do now and only `QueryExecModeCacheStatement` and `QueryExecModeCacheDescribe` would use the new path.

That would finish this unit of work. Then in another optional PR we could work out that extra interface for `CopyFromSource` to manually control this behavior.

alejandrodnm: I'll try to get this donde by the end of the week, otherwise it'll have to wait until next year. Thanks for the input :)

jackc:
LGTM! Thanks!

PR #1416, opened 2022-12-06
Add pgtype.Uint64 to support pgtype.PgLSN type (closes #1407)

### LINKED ISSUE #1407: Add pgtype.Uint64 to support pg_lsn type
I'd like to add `pgtype.Uint64` to support `pg_lsn` type.

I read https://github.com/jackc/pgx/issues/462#issuecomment-526838739 and noticed type LSN is defined at https://github.com/jackc/pglogrepl/blob/1e972958ba0339bc189fe5d5b190ce97f0258e3f/pglogrepl.go#L48.

However I think it's more consistent to add `pgtype.Uint64` since `pgtype.Uint32` was added to replace CID, OID, OIDValue, and XID at
https://github.com/jackc/pgx/commit/eec82c9433368eaf9855ad6ee0a4f36418c6318a

My attempt to add pgtype.Uint64 to support pg_lsn type is at https://github.com/jackc/pgx/commit/e925c2eebcd02dd16ad702da194c79c85000b51b

With this change, I can define `PgLSN` type like below just as described at https://pkg.go.dev/github.com/jackc/pgx/v5@v5.1.1/pgtype#hdr-Extending_Existing_Type_Support

```
type PgLSN struct {
	LSN   uint64
	Valid bool
}

func (n *PgLSN) ScanUint64(v pgtype.Uint64) error {
	*n = PgLSN{
		LSN:   v.Uint64,
		Valid: v.Valid,
	}
	return nil
}

func (n PgLSN) Uint64Value() (pgtype.Uint64, error) {
	return pgtype.Uint64{
		Uint64: n.LSN,
		Valid:  true,
	}, nil
}

func (src *PgLSN) String() string {
	if !src.Valid {
		return ""
	}
	return fmt.Sprintf("%X/%X", src.LSN>>32, uint32(src.LSN))
}

type PgLSNCodec struct {
	pgtype.Uint64Codec
}

func (PgLSNCodec) DecodeValue(tm *pgtype.Map, oid uint32, format int16, src []byte) (interface{}, error) {
	if src == nil {
		return nil, nil
	}

	var target PgLSN
	scanPlan := tm.PlanScan(oid, format, &target)
	if scanPlan == nil {
		return nil, fmt.Errorf("PlanScan did not find a plan")
	}

	err := scanPlan.Scan(src, &target)
	if err != nil {
		return nil, err
	}

	return target, nil
}
```

hnakamur:
Discussed in [Add pgtype.Uint64 to support pg_lsn type · Issue #1407 · jackc/pgx](https://github.com/jackc/pgx/issues/1407)


jackc:
The tests of `pg_lsn` need to skip CockroachDB for CI.

I also don't see how the text format could be working. `TestConnQuerPgLSN` is not testing encoding at all and is only testing decoding binary. That test should be in the `pgtype` package and use `pgxtest.RunValueRoundTripTests`.

jackc:
Looks like Cockroach is still getting triggered somewhere.

I ran the tests locally and it seems to work -- though TBH I still don't see exactly how. The text format is different for pg_lsn than a normal integer and I don't see a scan plan or an encode plan to the text format. I guess `ScanText` and `TextValue` or doing the job. I suppose it works, but that's not the normal path.

Also, I noticed that the database/sql implementations for `Value()` are returning `uint64`. This is not in spec (only allowed types are listed here: https://pkg.go.dev/database/sql/driver#Value).

hnakamur:
> Looks like Cockroach is still getting triggered somewhere.

It was `TestConnQuerPgLSN` in query_test.go. I removed it at https://github.com/jackc/pgx/pull/1416/commits/349d80d3c72471584b3c23c0102f5e51a325c001.

> I ran the tests locally and it seems to work -- though TBH I still don't see exactly how. The text format is different for pg_lsn than a normal integer and I don't see a scan plan or an encode plan to the text format. I guess ScanText and TextValue or doing the job. I suppose it works, but that's not the normal path.

Yes, I added debug logs and confirmed `ScanText` and `TextValue` are called.
It's my first time to try to add a type to pgx, and honestly I'm not familiar with the normal path.
Could you point me to an example or document that I can learn the normal path?

> Also, I noticed that the database/sql implementations for Value() are returning uint64. This is not in spec (only allowed types are listed here: https://pkg.go.dev/database/sql/driver#Value).

Thanks I didn't know that. I changed `Value()` to return int64 instead of uint64 at https://github.com/jackc/pgx/pull/1416/commits/99e8ae54befd89e68919ab7ab1ed3fc447f5d5d0.
I also changed `Scan()` to handle `int64` as well.

jackc:
> Could you point me to an example or document that I can learn the normal path?

I would expect the LSN codec to define its own `PlanScan` and `PlanEncode`. They would reuse the binary format code from uint64 but use their own paths for the text format. Check out `jsonb` code. It has a similar relationship to `json` that `pg_lsn` has the `uint64`. The only difference is that for json/jsonb it is the binary format that differs and for pg_lsn/uint64 it is the text format that differs.

> Thanks I didn't know that. I changed Value() to return int64 instead of uint64 at https://github.com/jackc/pgx/commit/99e8ae54befd89e68919ab7ab1ed3fc447f5d5d0.
I also changed Scan() to handle int64 as well.

Sorry to keep nitpicking, but that won't work in all cases either. What happens if the number is greater than the max int64? It will need to encode the value into a string to be safe.

hnakamur:
> I would expect the LSN codec to define its own PlanScan and PlanEncode. They would reuse the binary format code from uint64 but use their own paths for the text format. Check out jsonb code. It has a similar relationship to json that pg_lsn has the uint64. The only difference is that for json/jsonb it is the binary format that differs and for pg_lsn/uint64 it is the text format that differs.

Thanks for your explanation. I rewrote PgLSNCodec at https://github.com/jackc/pgx/pull/1416/commits/5ffe8df4adf9e82cb02fe53c874f767e53090f48.

> Sorry to keep nitpicking, but that won't work in all cases either. What happens if the number is greater than the max int64? It will need to encode the value into a string to be safe.

Thanks. I deleted `Scan()` since `TestPgLSNCodec` passes without it.

pgtype/pg_lsn.go
```diff
@@ -0,0 +1,206 @@
...
+		} {
+			if wrapperPlan, nextValue, ok := f(value); ok {
+				if nextPlan := c.PlanEncode(m, oid, format, nextValue); nextPlan != nil {
+					wrapperPlan.SetNext(nextPlan)
+					return wrapperPlan
+				}
+			}
+		}
+	}
+
```
jackc, line 86: I don't think this block is necessary here. That is a special case for json.

pgtype/uint64_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/jackc/pgx/v5/pgxtest"
+)
+
+func TestUint64Codec(t *testing.T) {
+	pgxtest.RunValueRoundTripTests(context.Background(), t, defaultConnTestRunner, pgxtest.KnownOIDQueryExecModes, "pg_lsn", []pgxtest.ValueRoundTripTest{
```
jackc, line 12: This needs to skip CockroachDB as well.

But it also makes me wonder if this is actually testing `Uint64Codec`. It's using the `pg_lsn` type. That would mean that `PgLSNCodec` is being used. We'd need to use another real uint64 PG type for this test... Hmm... That makes me wonder again about whether Uint64 should actually exist. I looked in the  PostgreSQL `pg_type` table and your example of `FullTransactionId` isn't exposed there. I don't think there are any other uint64 types in PostgreSQL. (Run `select * from pg_type where typlen = 8;` in PostgreSQL to see all candidates).

Unless we can find another uint64 type in PostgreSQL I think pg_lsn is all that should exist.

hnakamur: > I don't think there are any other uint64 types in PostgreSQL. (Run select * from pg_type where typlen = 8; in PostgreSQL to see all candidates).

Thanks. I looked at source code of `macaddr8` and `xid8` and confirmed that they are not backed with `uint64`.

> Unless we can find another uint64 type in PostgreSQL I think pg_lsn is all that should exist.

I am convinced now and am closing this pull request and the issue [Add pgtype.Uint64 to support pg_lsn type · Issue #1407 · jackc/pgx](https://github.com/jackc/pgx/issues/1407).

Sorry for taking your time and thanks again for your review.

PR #1425, opened 2022-12-12
Enable some CopyFrom tests for cockroachDB

alejandrodnm:
CockroachDB added support for COPY in version 20.2.

https://www.cockroachlabs.com/docs/v20.2/copy-from

There are some limitations on the implementation, that's why not all the existing tests were enabled.


jackc:
LGTM

PR #1436, opened 2022-12-16
pgtype Int fix minimum error message.

mwchambers:
Previously on the minimum condition the error would be:

  `is greater than maximum`

Also add encoding/json import into the .erb template as the import was missing after running rake generate.


jackc:
Thanks!

Issue #1437, opened 2022-12-17
`MaxConnLifetimeJitter` setting doesn't actually jitter


benweint:
We're seeing an issue wherein the `MaxConnLifetimeJitter` doesn't actually splay out DB reconnections as intended. Under high enough throughput, connections within the pool wind up being marked as expired at pretty much `MaxConnLifetime`, meaning that they all get closed and re-created in quick succession, which can have performance impacts for clients using the pool

## To reproduce

Run the following program, which sets both `MaxConnLifetime` and `MaxConnLifetimeJitter` to 10s, and issues queries every 10ms:

```go
package main

import (
	"context"
	"fmt"
	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/pgxpool"
	"os"
	"time"
)

func main() {
	// urlExample := "postgres://username:password@localhost:5432/database_name"
	ctx := context.Background()

	cfg, err := pgxpool.ParseConfig(os.Getenv("DATABASE_URL"))
	if err != nil {
		panic(err)
	}

	cfg.MinConns = 5
	cfg.MaxConns = 5
	cfg.MaxConnLifetime = 10 * time.Second
	cfg.MaxConnLifetimeJitter = 10 * time.Second

	lastNewConn := time.Now()
	cfg.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
		elapsed := time.Since(lastNewConn)
		lastNewConn = time.Now()
		fmt.Printf("new conn after %v\n", elapsed)
		return nil
	}

	pool, err := pgxpool.NewWithConfig(context.Background(), cfg)
	if err != nil {
		panic(err)
	}
	defer pool.Close()

	ticker := time.NewTicker(10 * time.Millisecond)
	secondTicker := time.NewTicker(1 * time.Second)

	for {
		select {
		case <-ticker.C:
			runOnce(ctx, pool)
		case <-secondTicker.C:
			fmt.Printf("\n")
		}
	}
}

func runOnce(ctx context.Context, pool *pgxpool.Pool) {
	result, err := pool.Query(ctx, "SELECT 1")
	if err != nil {
		fmt.Printf("got error: %v\n", err)
		return
	}
	result.Close()
}
```

## Expected behavior

Given the description of `MaxConnLifetimeJitter`, I would expect the log lines noting the creation of new DB connections to be uniformly distributed over a 10s window.

## Actual behavior

Even after running for a long time, new connection creations remain tightly clustered together, for example:

```
[...]

PR #1443, opened 2022-12-19
Make MaxConnLifetimeJitter setting actually jitter (addresses #1437)

benweint:
This change addresses the issue described in https://github.com/jackc/pgx/issues/1437 by caching the connection expiry time on the connection resource itself, rather than re-calculating it every time a connection is released.


jackc:
LGTM, thanks!

Issue #1426, opened 2022-12-13
sql.NullFloat64 scan bug


agandreev:
### Description
When I try to scan numeric value from postgres db table like this one `-0.01`, I receive error like that on scan level `can't scan into dest[3]: converting driver.Value type string (\"0.-1\") to a float64: invalid syntax`. It means that somehow postgres numeric value `-0.01` transfers into `"0.-1\"`.

### Reproduction

1. Prerequisites
`Postgres 12` with such sql table for just one value
``` sql
CREATE TABLE one_value_table
(
    val NUMERIC
);

INSERT INTO one_value_table(val) VALUES (-0.01);
```
2. Code bug reproduction
Simple code with pgxpool creation and running sql query.
`go1.19.2 darwin/amd64`, `github.com/jackc/pgx/v5 v5.2.0`
``` golang
package main

import (
	"context"
	"database/sql"
	"fmt"
	"os"

	"github.com/jackc/pgx/v5/pgxpool"
)

func main() {
	ctx := context.Background()

	cfg := scanConfig()
	db, err := pool(ctx, cfg.User, cfg.Password, cfg.Port, cfg.DBName)
	if err != nil {
		panic(err)
	}

	rows, err := db.Query(ctx, `SELECT * FROM one_value_table`)
	if err != nil {
		panic(err)
	}

	oneVals := make([]oneValueDBView, 0)
	for rows.Next() {
		var oneVal oneValueDBView
		if err = rows.Scan(&oneVal.Val); err != nil {
			panic(err)
		}
		oneVals = append(oneVals, oneVal)
	}

	fmt.Println(oneVals)
}

func pool(ctx context.Context, user, password, port, dbname string) (*pgxpool.Pool, error) {
	connection := fmt.Sprintf(
		"postgres://%s:%s@localhost:%s/%s?sslmode=disable",
		user,
		password,
		port,
		dbname,
	)
	return pgxpool.New(ctx, connection)
}

func scanConfig() config {
	return config{
		User:     os.Getenv("DB_USER"),
		Password: os.Getenv("DB_PASSWORD"),
		Port:     os.Getenv("DB_PORT"),
		DBName:   os.Getenv("DB_NAME"),
	}
}

type config struct {
	User     string
	Password string
	Port     string
	DBName   string
}

type oneValueDBView struct {
	Val sql.NullFloat64 `db:"val"`
}
```
3. Run
-  
``` bash
[...]

PR #1459, opened 2023-01-04
pgtype.Numeric numberTextBytes() bug for negative 0.0X

mwchambers:
I'm not sure what the best fix is. The first commit demonstrates in the tests an issue with MarshalJSON and Scan() as per https://github.com/jackc/pgx/issues/1426.

Second commit as a quick workaround.


jackc:
LGTM

PR #1461, opened 2023-01-05
Adds CollectOneRowOK

manniwood:
`CollectOneRowOK()` is a "comma OK idom" wrapper around `CollectOneRow()`.

Compare `CollectRows()` to `CollectOneRow()`: when `CollectRows()` finds no rows, it does not return an error; it's just that the returned slice is length 0. But when `CollectOneRow()` finds no rows, it returns an error (`ErrNoRows`).

`CollectOneRowOK()` provides a way to also return without error when no rows are found. Whereas a caller of `CollectRows()` can check if the returned slice is length > 0 to determine if rows were found, a caller of CollectOneRowOK can check the second return value (usually named `ok`) to see if a row was found.

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -449,6 +449,23 @@ func CollectOneRow[T any](rows Rows, fn RowToFunc[T]) (T, error) {
 	return value, rows.Err()
 }
 
+// CollectOneRowOK is CollectOneRow with the "comma OK idiom": think 'val, ok := myMap["foo"]'
+// or os.GetEnv (no comma OK idiom) versus os.LookupEnv (follows comma OK idiom).
+// If no rows are found, the second return value is false (instead of returning error ErrNoRows
+// as CollectOneRow does). If a row is found, the second return value is true.
+func CollectOneRowOK[T any](rows Rows, fn RowToFunc[T]) (T, bool, error) {
+	var value T
+	var err error
+	value, err = CollectOneRow(rows, fn)
+	if err != nil {
+		if err == ErrNoRows {
+			return value, false, nil
+		}
+		return value, false, err
+	}
+	return value, true, nil
+}
+
 // RowTo returns a T scanned from row.
 func RowTo[T any](row CollectableRow) (T, error) {
 	var value T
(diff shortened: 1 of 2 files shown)


jackc:
Can you provide some examples of how this simplifies calling code? I don't see how it is better than using `CollectOneRow` directly. If you want to consider no rows as a success case if can be done with this error check

```go
if err != nil && err != pgx.ErrNoRows {
// handle err
}
```

manniwood:
Hello! Thanks for considering my PR.
Thanks for reminding me of this perspective: "how does this simplify calling code?"
I don't think it simplifies calling code. I think it clarifies calling code.
I think it clarifies calling code in two ways.

1) It's not an error.
Let's say I am at a `psql` prompt.
If I type `select name from users`, I expect to get every row in the table. But what if the table is empty? I get no rows. But I don't get an error.
If I type `select name from users where id = 1` and there is a unique index on id, I expect to get one row. But what if there is no user with id 1? I get no rows. But I don't get an error.
So pgx returning an error when Postgres had no error is a bit surprising.

2) Consistency. I could see a version of pgx where `CollectRows()` checks to see if there are any rows to read, and, if not, returns `pgx.ErrNoRows`. But it does not. But `CollectOneRow()` does when no rows are found. This behavior provides an inconsistent answer to the question "is no rows returned an error condition?" In fact, now that I think about it, my PR should include a `CollectRowsErr()` so that users who prefer an error on no rows found can get an error whether collecting multiple rows or one row! Now you've really made me think. :-)

Anyway, these are my thoughts. To scratch my own itch, I have created https://github.com/manniwood/pgxtras so that I can have this functionality for myself and share it with whoever else may like it. Please let me know if I have credited your pgx work enough in my spin-off project: I want to give credit where it's due!

jackc:
I like the idea of new ideas being proven in other packages for a while. For what it's worth the whole `Collect*` concept was first tested in my own pgx grab bag package pgxutil.

Let's validate these helpers functions in your package for a while before making any decision on adding it to the pgx core.

manniwood:
That seems reasonable. Anybody desiring this behavior can now find it in my little project.

And yes, one of my favorite things about pgx5 is how it incorporated pgxutil. I was quite late in appreciating what pgxutil brought to pgx --- `Exec()`, `Query()` and `QueryRow()` worked so well for me I didn't explore alternatives. But now I'm all in on these utility functions.

PR #1471, opened 2023-01-15
Remove unused test struct, replace deprecated ioutil, return err fix, and spelling.

mwchambers:
Done as separate commits should you want to cherry-pick.


jackc:
LGTM

PR #1490, opened 2023-02-07
Change query in getCompositeType to ignore dropped attributes

### LINKED ISSUE #1489: LoadType fails if attribute was deleted from composite type
**Describe the bug**
Using `LoadType` to register a composite fails, if an attribute was deleted.

The error is in the `Conn.getCompositeFields` function. To be more specific:

```sql
select attname, atttypid
from pg_attribute
where attrelid=$1
order by attnum
```
returns

| attname  | atttypid |
|----------|---------|
| attr1 | 111111 |
| attr2 | 222222 |
|........pg.dropped.3........| 0 |
| attr3 | 333333 |

if the 3rd attribute was deleted and then added again.

**Expected behavior**
Deleted attributes should be ignored.

**Actual behavior**
`LoadType` fails with the error `unknown composite type field OID: 0`

**Version**
 - Go: go version go1.19.5 darwin/amd64
 - PostgreSQL: PostgreSQL 14.4 (Debian 14.4-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
 - pgx: v5.2.0

Alisaien:
Fixes #1489 and adds a simple test case.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1282,7 +1282,7 @@ func (c *Conn) getCompositeFields(ctx context.Context, oid uint32) ([]pgtype.Com
 	var fieldOID uint32
 	rows, _ := c.Query(ctx, `select attname, atttypid
 from pg_attribute
-where attrelid=$1
+where attrelid=$1 and not attisdropped
 order by attnum`,
 		typrelid,
 	)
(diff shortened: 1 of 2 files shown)


jackc:
LGTM

PR #1495, opened 2023-02-09
replace erroneous reflect.New with reflect.Zero in TryWrapStructScanPlan

Alisaien:
Fix as said in the title. This only had effect in conjunction with `TryPointerPointerScanPlan`, where nextDest is a nil pointer.
The `TestMapScanPointerToNilStructDoesNotCrash` still passes, but the error thrown seems different than what test is intended for.

Also adds a test case.


jackc:
Thanks! LGTM.

Issue #1501, opened 2023-02-13
panic error still occurs in v5.3.0 and v4.18.0 when two versions of pgx/stdlib are imported


bon10:
**Describe the bug**

I know this issue has already been resolved by  https://github.com/jackc/pgx/pull/1498
However, the problem still occurs in my situation.

**To Reproduce**

A simple example is when I have two files and call different versions of pgx/stdlib for each file.
In practice, this occurs when a module added with go mod uses a different version of pgx/stdlib than the app.

File1
```go
package main

import (
	"database/sql"
	"fmt"

	_ "github.com/jackc/pgx/v5/stdlib"
)

func main() {
	dsn := fmt.Sprintf("postgresql://postgres:@password:localhost/test_db?sslmode=disable")
	db, err := sql.Open("pgx", dsn)
	if err != nil {
		panic(err)
	}
	defer db.Close()
	InitSecond()  // call function of File2
}
```

File2
```go
package main

import (
	"database/sql"
	"fmt"

	_ "github.com/jackc/pgx/v4/stdlib"
)

func InitSecond() {
	dsn := fmt.Sprintf("postgresql://postgres:@password:localhost/test_db?sslmode=disable")
	db, err := sql.Open("pgx", dsn)
	if err != nil {
		panic(err)
	}
	defer db.Close()
}
```
**Expected behavior**

No error

**Actual behavior**

```
panic: sql: Register called twice for driver pgx

goroutine 1 [running]:
database/sql.Register({0x1005c9579, 0x3}, {0x100705498, 0x140000a02d0})
	/usr/local/go/src/database/sql/sql.go:51 +0x124
github.com/jackc/pgx/v4/stdlib.init.0()
	/Users/bon/work/go/pkg/mod/github.com/jackc/pgx/v4@v4.18.0/stdlib/sql.go:99 +0x144
```

**Version**

 - Go: go version go1.20 darwin/arm64
 - PostgreSQL: PostgreSQL 15.0 (Debian 15.0-1.pgdg110+1) on aarch64-unknown-linux-gnu, comp
iled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
 - pgx:  v5.3.0 / v4.18.0

**Additional context**

I changed the code and it works fine.
However, I am unable to prepare a test that reproduces this situation.

```diff
- if i := sort.SearchStrings(sql.Drivers(), "pgx"); len(drivers) >= i || drivers[i] != "pgx" {
[...]

PR #1504, opened 2023-02-13
simplify duplicate `pgx` registration guard

prochac:
The binary search is overkill here.
Readability first.

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -66,7 +66,6 @@ import (
 	"math"
 	"math/rand"
 	"reflect"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


prochac:
Also it should fix https://github.com/jackc/pgx/issues/1501

I have a reason to think the previous guard is wrong.
https://github.com/jackc/pgx/issues/1501#issuecomment-1428442392

But I admin, I can be wrong, because I don't clearly understand the documentation of 
https://pkg.go.dev/sort#SearchStrings
In some way, it's another reason for simple for loop.

I can provide benchmark, if requested :wink:

prochac:
> @prochac @jackc Oops, I apologize for the bad testing.

Don't worry, possibly it would take me much longer to make the PR without your contribution :D

Issue #1509, opened 2023-02-15
Add MarshalJSON/UnmarshalJSON methods to v5/pgtype.Array


carlmjohnson:
pgtype.Array is a useful type, but it marshals to/from JSON poorly. If there is any interest, I can open a PR to add MarshalJSON/UnmarshalJSON methods to it.

jackc:
You specifically mean `pgtype.Array` no `pgtype.FlatArray`? It's fine with me, but it might be a little awkward to round trip the array dimensions.

carlmjohnson:
See https://github.com/kyleconroy/sqlc/issues/2083 for why I want to change pgtype.Array. :-) FlatArray has basically fine JSON serialization other than the usual Go "nil slice becomes null JSON" thing.

jackc:
I don't know the context for the sqlc thing, but I'm a little surprised whenever I see `pgtype.FlatArray` or especially `pgtype.Array` used directly. Typically, pgx can directly work with `[]T`, `[][]T`, etc. I only anticipated `pgtype.Array` being used directly if the application actually needed to work with non-standard starting indexes.

carlmjohnson:
sqlc used []T directly in its pgx/v4 generator, so maybe the change to pgtype.Array with v5 is a regression that they should fix.

kyleconroy:
I posted this comment on https://github.com/kyleconroy/sqlc/pull/2090, but thought this would also be a good place to discuss.

@jackc If I have a nullable text column, I'm assuming that scanning those values into a `[]string` will work, but if there's a null value, the scanning will fail.

carlmjohnson:
I don't think those can be scanned into a pgtype.Array either:

```
$ go doc pgtype.Array
package pgtype // import "github.com/jackc/pgx/v5/pgtype"

type Array[T any] struct {
        Elements []T
        Dims     []ArrayDimension
        Valid    bool
}
    Array represents a PostgreSQL array for T. It implements the ArrayGetter and
    ArraySetter interfaces. It preserves PostgreSQL dimensions and custom lower
    bounds. Use FlatArray if these are not needed.
```

The elements are `[]T`. ArrayDimensions lets you have multidimensional arrays and starting offsets other than 0, but don't have anything for null.

jackc:
Neither `[]T`, `pgtype.FlatArray[T]`, or `pgtype.Array[T]` have any effect on whether `T` can be `NULL`. When `T` is `string` none of them will work if a `NULL` element is scanned.

[...]

PR #1511, opened 2023-02-15
pgtype.Array: add MarshalJSON

carlmjohnson:
Refs #1509

### DIFF
--- a/pgtype/array.go
+++ b/pgtype/array.go
@@ -3,6 +3,7 @@ package pgtype
 import (
 	"bytes"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"reflect"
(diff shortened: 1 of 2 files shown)


jackc:
Not sure how I feel about this. Any non-standard dimensions won't survive a JSON marshalling. OTOH, this is a more convenient representation for the typical use case. 🤔

carlmjohnson:
When you say it won't surive marshalling, do you mean just because there's no UnmarshalJSON method yet or because there's no equivalent for the offset thing?

jackc:
No equivalent for the offset. i.e. marshalling to JSON is a lossy operation.

carlmjohnson:
Closing following discussion and changes to SQLC.

Issue #1482, opened 2023-01-28
Non-blocking I/O on Windows


jackc:
pgx uses non-blocking I/O on *nix. However, it falls back to simulating non-blocking I/O by using short deadlines on Windows. Even a short read deadline waiting to do nothing can have a significant negative performance impact in certain circumstances such as `CopyFrom` (see #1481). 

I do not have a Windows dev box nor do I have the Windows expertise to easily implement it. So if any Windows developers work on this that would be great. The *nix implementation is in internal/nbconn/nbconn_real_non_block.go.

jackc:
I did a little research and here are a few notes that may be useful when this is implemented.

There is code that could probably be adapted in the Go tests for `syscall.RawConn`: https://github.com/golang/go/blob/23c0121e4eb259cc1087d0f79a0803cbc71f500b/src/net/rawconn_windows_test.go#L13-L46

And this appears to be the Windows documentation for the API to use for reading: https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recv

jackc:
Branch [windows-non-blocking-io](https://github.com/jackc/pgx/tree/windows-non-blocking-io) has a candidate implementation. However, my Windows development environment and expertise is still very limited. It *seems* to work as well as the fake non-blocking IO. But I was unable to run the complete test suite even before this change. Not sure if I didn't have my Windows environment set up right, if some tests don't work on Windows, or if there actually is broken functionality on Windows. It'd be helpful if those who actually use Windows could test this. I've reached the limit of how much time I am going to invest in a platform I don't even use.

PR #1525, opened 2023-02-25
Try to make Windows non-blocking I/O

### LINKED ISSUE #1523: PGX v5 Pool poor performance on Windows
**Describe the bug**
When using the pgx pool on Windows, the query time is too long (e.g. `SELECT 1` query can take up to 120ms)

**To Reproduce**
Run code example below

<details>
<summary>Code (click me)</summary>

```go
package main

import (
	"context"
	"fmt"
	pgxv4 "github.com/jackc/pgx/v4"
	pgxv4pool "github.com/jackc/pgx/v4/pgxpool"
	pgxv5 "github.com/jackc/pgx/v5"
	pgxv5pool "github.com/jackc/pgx/v5/pgxpool"
	"os"
	"time"
)

func main() {
	connStr := os.Getenv("DATABASE_URL")

	fmt.Printf("pgx v5 conn:\n")
	testPgxV5Conn(connStr)

	fmt.Printf("\npgx v5 pool:\n")
	testPgxV5Pool(connStr)

	fmt.Printf("\npgx v4 conn:\n")
	testPgxV4Conn(connStr)

	fmt.Printf("\npgx v4 pool:\n")
	testPgxV4Pool(connStr)
}

func testPgxV5Pool(connStr string) {
	cfg, err := pgxv5pool.ParseConfig(connStr)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Unable to parse pool config: %v\n", err)
		os.Exit(1)
	}

	cfg.MaxConns = 1

	conn, err := pgxv5pool.NewWithConfig(context.Background(), cfg)
	if err != nil {
		os.Exit(1)
	}
	defer conn.Close()

	if err := conn.Ping(context.Background()); err != nil {
		fmt.Fprintf(os.Stderr, "pool ping failed: %v\n", err)
		os.Exit(1)
	}

	makeAttemptPool(conn, 0)

	time.Sleep(time.Second)

	makeAttemptPool(conn, 1)

	time.Sleep(time.Second)

	makeAttemptPool(conn, 2)
}

func testPgxV5Conn(connStr string) {
	conn, err := pgxv5.Connect(context.Background(), connStr)
	if err != nil {
		os.Exit(1)
	}

	makeAttempt(conn, 0)

	time.Sleep(time.Second)

	makeAttempt(conn, 1)

	time.Sleep(time.Second)

	makeAttempt(conn, 2)
}

func testPgxV4Pool(connStr string) {
	cfg, err := pgxv4pool.ParseConfig(connStr)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Unable to parse pool config: %v\n", err)
		os.Exit(1)
	}

	cfg.MaxConns = 1

	conn, err := pgxv4pool.ConnectConfig(context.Background(), cfg)
	if err != nil {
		os.Exit(1)
	}
	defer conn.Close()

	if err := conn.Ping(context.Background()); err != nil {
		fmt.Fprintf(os.Stderr, "pool ping failed: %v\n", err)
		os.Exit(1)
	}

	makeAttemptPoolV4(conn, 0)

	time.Sleep(time.Second)

	makeAttemptPoolV4(conn, 1)

	time.Sleep(time.Second)

	makeAttemptPoolV4(conn, 2)
}

func testPgxV4Conn(connStr string) {
	conn, err := pgxv4.Connect(context.Background(), connStr)
	if err != nil {
		os.Exit(1)
	}

	makeAttemptV4(conn, 0)

	time.Sleep(time.Second)

	makeAttemptV4(conn, 1)

	time.Sleep(time.Second)

	makeAttemptV4(conn, 2)
}

func makeAttempt(conn *pgxv5.Conn, attempt int) {
	start := time.Now()

	var id int64

	err := conn.QueryRow(context.Background(), "select 1 AS id").Scan(&id)
	if err != nil {
		fmt.Fprintf(os.Stderr, "QueryRow failed: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("v5 conn time took #%d = %s\n", attempt, time.Since(start))
}

func makeAttemptPool(conn *pgxv5pool.Pool, attempt int) {
	start := time.Now()

	var id int64

	err := conn.QueryRow(context.Background(), "select 1 AS id").Scan(&id)
	if err != nil {
		fmt.Fprintf(os.Stderr, "QueryRow failed: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("v5 pool time took #%d = %s\n", attempt, time.Since(start))
}

func makeAttemptV4(conn *pgxv4.Conn, attempt int) {
	start := time.Now()

	var id int64

	err := conn.QueryRow(context.Background(), "select 1 AS id").Scan(&id)
	if err != nil {
		fmt.Fprintf(os.Stderr, "QueryRow failed: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("v4 conn time took #%d = %s\n", attempt, time.Since(start))
}

func makeAttemptPoolV4(conn *pgxv4pool.Pool, attempt int) {
	start := time.Now()

	var id int64

	err := conn.QueryRow(context.Background(), "select 1 AS id").Scan(&id)
	if err != nil {
		fmt.Fprintf(os.Stderr, "QueryRow failed: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("v4 pool time took #%d = %s\n", attempt, time.Since(start))
}
```
</details>

**Expected behavior**
pgx pool shouldn't add so much delay

**Actual behavior**
pgx pool adds extra delay
```
pgx v5 conn:
v5 conn time took #0 = 1.2867ms
v5 conn time took #1 = 0s
v5 conn time took #2 = 86.7µs

pgx v5 pool:
v5 pool time took #0 = 505.3µs
v5 pool time took #1 = 111.0746ms
v5 pool time took #2 = 109.5005ms

pgx v4 conn:
v4 conn time took #0 = 701.8µs
v4 conn time took #1 = 639.5µs
v4 conn time took #2 = 130.9µs

pgx v4 pool:
v4 pool time took #0 = 744.9µs
v4 pool time took #1 = 0s
v4 pool time took #2 = 135.3µs
```

Here is timings for Ubuntu (go 1.20):
```
pgx v5 conn:
v5 conn time took #0 = 440.4µs
v5 conn time took #1 = 292.9µs
v5 conn time took #2 = 395.5µs

pgx v5 pool:
v5 pool time took #0 = 503.7µs
v5 pool time took #1 = 338.3µs
v5 pool time took #2 = 346µs

pgx v4 conn:
v4 conn time took #0 = 7.0436ms
v4 conn time took #1 = 282.7µs
v4 conn time took #2 = 271.4µs

pgx v4 pool:
v4 pool time took #0 = 424.5µs
v4 pool time took #1 = 383.5µs
v4 pool time took #2 = 284.3µs
```

**Version**
 - Go: `go version go1.19 windows/amd64` and also tested on `1.20` (win, x64)
 - PostgreSQL: `PostgreSQL 14.5, compiled by Visual C++ build 1914, 64-bit`
 - pgx: `v5.3.0`

codercms:
Resolves #1523 #1482 

Do not merge: Possible broken tests


codercms:
@jackc this PR works in real app, but I can't run tests for nbconn (`nbconn_test.go`) on my machine, because of it continuously eat RAM (10GB+) and I don't have enough RAM. Do you have any ideas?

jackc:
It looks like `TestInternalNonBlockingWrite/Pipe` is hanging on master too. So I don't think that has anything to do with this change. Not sure why it's only failing on Windows though. There doesn't seem to be any platform specific code.

jackc:
I fixed `TestInternalNonBlockingWrite/Pipe` on master. There are still a few failures on `TestInternalNonBlockingWrite/*` but it doesn't hang now. Looking into those errors now.

jackc:
Looks good to me. I'm inclined to merge this as is.

Regarding the `TestInternalNonBlockingWrite` tests, I think I have the resource exhaustion you experienced fixed on master. But I still get a `local error: tls: bad record MAC` failure on a couple tests. And as crazy as it sounds, I think it is actually a bug in Go or in Windows. I spent a number of hours investigating and producing a test case. I filed an issue about it https://github.com/golang/go/issues/58764.

If you get a chance could you help me with that error by:

1. Run the test case (https://github.com/jackc/golang-windows-dupreads) I produced for that issue and see if you get the same error.
2. Review my report https://github.com/golang/go/issues/58764. My report was very understandably greeted with a little skepticism. I would be skeptical too. But as far as I can tell, the bug has to be in the net package or below (i.e. Go runtime or Windows).

codercms:
@jackc I tested the master branch, `nbconn_test.go`, everything is fine except two test cases, I got the same errors as you:
```
TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO
        	            	local error: tls: bad record MAC

TestInternalNonBlockingWrite/TLS_over_TCP_with_Real_Non-blocking_IO
        	            	local error: tls: bad record MAC
```

P.S. If you need to test something on WIndows, I'm ready to help
P.P.S. I'm also ran tests for `nbconn_test.go` on Ubuntu 22.04 (WSL 2, Go 1.18.1) and all tests pass, it looks like this bug only occurs on Windows.

codercms:
@jackc I have the same errors as you in this test repo - https://github.com/jackc/golang-windows-dupreads
Tried in different ways, but the error is always the same.
Look like a bug in Go's stdlib.

drakkan:
> @jackc I have the same errors as you in this test repo - https://github.com/jackc/golang-windows-dupreads Tried in different ways, but the error is always the same. Look like a bug in Go's stdlib.

@codercms please read the latest comment in the upstream issue. Can you check if anything changes using Go 1.19.5? If you can't, I'll try it myself next weekend (I don't have a Windows test environment right now)

codercms:
@drakkan same error with Go 1.19.5 & 1.18.10 & 1.16 & 1.15

drakkan:
> @drakkan same error with Go 1.19.5 & 1.18.10 & 1.16 & 1.15

Thanks for your testing. Not sure how to try to help. Based on [this comment](https://github.com/googleapis/google-cloud-go/issues/1581#issuecomment-531817165) `bad record MAC` is a very internal error. I hope Go team could take a look.

What are the pratical implications for pgx? Can we have sporadic errors in Windows with TLS enabled?

jackc:
See https://github.com/golang/go/issues/58764 for my longer description of the problem. But I'm pretty sure the problem is **not** in the tls package. I made wrapper conns that sit below the tls conn and record the traffic. They are seeing the erroneous duplication of traffic. That is what breaks the tls connection. This also suggests that the problem is not exclusive to tls connections, it's just that's the only way I have managed to trigger the issue.

As far as its affect for pgx on Windows, it's concerning but I think it would be very difficult to trigger in real world usage. We can only duplicate this issue when at least 2.7 MB of writes are buffered on the client side and the server has simultaneously written a large amount of data. The only way I can foresee that happening is if a `CopyFrom` with a lot of data is executed and there is something on the server that causes a **huge** amount of data to be sent back while it is receiving the data (e.g. an insert trigger that did a lot of `raise notice`).

jackc:
I've decided to merge this now. The failing test predates this change and appears to be a Go or Windows issue. This change should be an improvement regardless.

internal/nbconn/nbconn_real_non_block_windows.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return nil
+}
+
+// realNonblockingWrite does a non-blocking write. readFlushLock must already be held.
+func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
+	if c.nonblockWriteFunc == nil {
+		c.nonblockWriteFunc = func(fd uintptr) (done bool) {
+			// Make sock non-blocking
+			if err := setSockMode(fd, sockModeNonBlocking); err != nil {
```
drakkan, line 47: I don't have a test case so feel free to ignore this comment. 
What if another goroutine ends a write and set the socket to blocking mode after this method ends and before calling `WSASend`?

drakkan: > @drakkan Connection isn't considered as thread (goroutine) safe , it should be managed through connection pool, that guarantees that only one goroutine can work with connection.
> 
> Here it is - https://github.com/jackc/pgx/wiki/Getting-started-with-pgx#using-a-connection-pool

Ok thanks. I was just trying to figure out if it could be related to #1552, #1548

codercms: @drakkan hmm, #1552 looks really strange. I started debugging this case and found that there is inconsistent `setSockMode` calls:
```go
var currSockMode = sockModeBlocking

func setSockMode(fd uintptr, mode sockMode) error {
	if mode == currSockMode {
		fmt.Printf("double setSockMode=%d\n", mode)
	}

	currSockMode = mode

	res, _, err := ioctlsocket.Call(fd, uintptr(FIONBIO), uintptr(unsafe.Pointer(&mode)))
	// Upon successful completion, the ioctlsocket returns zero.
	if res != 0 && err != nil {
		return err
	}

	return nil
}
```

Here is output while using CopyFrom with large amount of data:
```
double setSockMode=1
double setSockMode=0
double setSockMode=1
double setSockMode=0
double setSockMode=1
double setSockMode=0
double setSockMode=0
double setSockMode=0
double setSockMode=0
double setSockMode=0
double setSockMode=1
double setSockMode=0
double setSockMode=0
double setSockMode=0
double setSockMode=0
double setSockMode=1
(hangs forever)
```

codercms: Added goroutine ids to output:
```
double setSockMode=1 (id=1)
double setSockMode=0 (id=1)
double setSockMode=0 (id=25)
double setSockMode=0 (id=1)
double setSockMode=0 (id=25)
double setSockMode=1 (id=25)
```

Looks like there is race condition

codercms: Just commented out these lines of code:
```go
// Make sock blocking again
if err := setSockMode(fd, sockModeBlocking); err != nil {
	c.nonblockWriteErr = err
	return true
}
```

And CopyFrom completed successfully

codercms: @drakkan I see two possible solutions:
1. Make socket to be always in non-blocking mode (but I'm not sure how it can affect other operations)
2. Wrap Windows non-blocking read/write into critical section, e.g.:
```go
var guard = make(chan struct{}, 1)

func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
	if c.nonblockWriteFunc == nil {
		c.nonblockWriteFunc = func(fd uintptr) (done bool) {
			guard <- struct{}{}
// .....

	<-guard
        return true
```

codercms: @jackc do you have any ideas?

drakkan: > Added goroutine ids to output:
> 
> ```
> double setSockMode=1 (id=1)
> double setSockMode=0 (id=1)
> double setSockMode=0 (id=25)
> double setSockMode=0 (id=1)
> double setSockMode=0 (id=25)
> double setSockMode=1 (id=25)
> ```
> 
> Looks like there is race condition

@codercms  Thanks for confirming. I did a test with this change

```
diff --git a/internal/nbconn/nbconn_real_non_block_windows.go b/internal/nbconn/nbconn_real_non_block_windows.go
index ed1662e0..de63c244 100644
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -6,6 +6,7 @@ import (
        "errors"
        "golang.org/x/sys/windows"
        "io"
+       "sync"
        "syscall"
        "unsafe"
 )
@@ -21,6 +22,10 @@ var dll = syscall.MustLoadDLL("ws2_32.dll")
 // );
 var ioctlsocket = dll.MustFindProc("ioctlsocket")
 
+var (
+       mu sync.Mutex
+)
+
 type sockMode int
 
 const (
@@ -41,6 +46,9 @@ func setSockMode(fd uintptr, mode sockMode) error {
 
 // realNonblockingWrite does a non-blocking write. readFlushLock must already be held.
 func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
+       mu.Lock()
+       defer mu.Unlock()
+
        if c.nonblockWriteFunc == nil {
                c.nonblockWriteFunc = func(fd uintptr) (done bool) {
                        // Make sock non-blocking
@@ -92,6 +100,9 @@ func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
 }
 
 func (c *NetConn) realNonblockingRead(b []byte) (n int, err error) {
+       mu.Lock()
+       defer mu.Unlock()
+
        if c.nonblockReadFunc == nil {
                c.nonblockReadFunc = func(fd uintptr) (done bool) {
                        // Make sock non-blocking
```

If I run the test cases I have this

```
Z:\pgx>go test -v --count 1 -timeout 30s -run ^TestInternalNonBlockingWrite$ github.com/jackc/pgx/v5/internal/nbconn
=== RUN   TestInternalNonBlockingWrite
=== RUN   TestInternalNonBlockingWrite/Pipe
=== RUN   TestInternalNonBlockingWrite/TCP_with_Fake_Non-blocking_IO
=== RUN   TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO
    nbconn_test.go:303:
                Error Trace:    Z:\pgx\internal\nbconn\nbconn_test.go:303
                                                        Z:\pgx\internal\nbconn\nbconn_test.go:143
                Error:          Received unexpected error:
                                local error: tls: bad record MAC
                Test:           TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO
=== RUN   TestInternalNonBlockingWrite/TCP_with_Real_Non-blocking_IO
=== RUN   TestInternalNonBlockingWrite/TLS_over_TCP_with_Real_Non-blocking_IO
--- FAIL: TestInternalNonBlockingWrite (0.40s)
    --- PASS: TestInternalNonBlockingWrite/Pipe (0.17s)
    --- PASS: TestInternalNonBlockingWrite/TCP_with_Fake_Non-blocking_IO (0.03s)
    --- FAIL: TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO (0.16s)
    --- PASS: TestInternalNonBlockingWrite/TCP_with_Real_Non-blocking_IO (0.02s)
    --- PASS: TestInternalNonBlockingWrite/TLS_over_TCP_with_Real_Non-blocking_IO (0.02s)
FAIL
FAIL    github.com/jackc/pgx/v5/internal/nbconn 0.939s
FAIL
```

so also the bad record MAC seems fixed for the `TLS_over_TCP_with_Real_Non-blocking_IO` test (fake non blocking IO still fails). @jackc what do you think about? Thanks

codercms: @drakkan @jackc I found this article - https://scarff.id.au/blog/2022/go-concurrency-idioms-dont-work-for-duplex-sockets/ about Go non-blocking I/O, and it seems like concurrent read/write from net.Conn is not safe.
Then I found this package - https://github.com/lesismal/nbio (Windows is not supported)

See implementation details:
* Read - https://github.com/lesismal/nbio/blob/master/conn_unix.go#L63
* Write - https://github.com/lesismal/nbio/blob/master/conn_unix.go#L146

The both read and write operations using shared mutex to prevent concurrent read & write.

What do you think?  
Should be concurrency limited in PGX for UNIX systems too? (like one of possible solutions for Windows)

drakkan: > @drakkan @jackc I found this article - https://scarff.id.au/blog/2022/go-concurrency-idioms-dont-work-for-duplex-sockets/ about Go non-blocking I/O, and it seems like concurrent read/write from net.Conn is not safe. Then I found this package - https://github.com/lesismal/nbio (Windows is not supported)
> 
> See implementation details:
> 
> * Read - https://github.com/lesismal/nbio/blob/master/conn_unix.go#L63
> * Write - https://github.com/lesismal/nbio/blob/master/conn_unix.go#L146
> 
> The both read and write operations using shared mutex to prevent concurrent read & write.
> 
> What do you think? Should be concurrency limited in PGX for UNIX systems too? (like one of possible solutions for Windows)

@codercms on Linux if I print the goroutine id I have a different id for reads and writes. I assumed the same for Windows and since on Windows we set the blocking mode in every read and write this will explain the reported issues (at least the hangs). This is the reason I asked [this](https://github.com/jackc/pgx/pull/1525#discussion_r1139224288).
I'm not familiar enough with pgx code to understand if reading and writing from different goroutines can cause issues also on Linux/Unix

jackc: I don't have much expertise with Windows network programming, but a bit of doc and Google searching seems to indicate that it *should* be safe to for one thread to read and one thread to write from the safe socket. However, it appears that some of the winsock API is not thread safe. Perhaps that includes the `ioctlsocket` function.

...

But now that I think about it a bit more I think I have a more likely cause for the deadlock. pgx will read and write concurrently under certain conditions (such as `bufferNonblockingRead` when a write is blocked). Both of these functions set non-blocking mode, read or write, and set blocking mode.

What happens if the following sequence of events occurs:

1. Writer sets non-blocking mode
2. Reader sets non-blocking mode
3. Writer writes
4. Writer sets blocking mode
5. Reader reads

The socket is unexpectedly in blocking mode for the read. If nothing is ever received it hangs.

---

As far as Unix goes, I think that since we are not changing the socket options, we are only reading and writing, it is safe.

---

I **really** wish the Go stdlib included non-blocking I/O. This system is so complicated...

I'm half tempted to try using goroutines and normal Go I/O again. I tried it a few years ago and the performance hit was significant, but maybe modern Go has higher performance...

codercms: @jackc You're absolutely right about that socket can be used for concurrent read/write (socket is full duplex).
I think implementing critical section (e.g. mutex) for Windows non-blocking I/O will be simplest solution. What do you think will this fix the problem? But additional benchmark is needed, because I/O actually will be not full duplex (you can only write or only read at the same time).

jackc: That would be the simplest approach.

But it seems it should be possible to switch to non-blocking mode at the beginning of a sequence of read or writes and turn it off at the end. Not sure how messy that would be though.

codercms: @jackc switching to non-blocking mode looks good too, because it doesn't hit performance.
I think non-blocking conn could expose `setBlocking(blocking bool)` method or something like this, for Windows it will call `ioctlsocket`, for Unix it will do nothing.

codercms: @jackc I made draft PR for setting connection non-blocking mode explicitly - https://github.com/jackc/pgx/pull/1555 . Please have a look on it

PR #1538, opened 2023-03-04
Add QueuedQuery.Result()

roelvandergoot:
This allows the user to have access to the pgconn.PgError early on instead of after BatchResults.Close(). This allows for error parsing in the QueuedQuery context. With the existing functions, QueuedQuery.Query(), QueuedQuery.QueryRow() and QueuedQuery.Exec(), the associated QueuedQuery has to be deduced.

Another reason to use QueuedQuery.Result() is that the other functions result (potentially) in multiple kinds of error types: pgconn.PgError and the error types returned from the functions passed in through QueuedQuery.Query(), QueuedQuery.QueryRow() and QueuedQuery.Exec()

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -57,6 +57,12 @@ func (qq *QueuedQuery) Exec(fn func(ct pgconn.CommandTag) error) {
 	}
 }
 
+// Result sets fn to be called when the response to qq is received and gives
+// access to the pgx error immediately (unlike Query, QueryRow or Exec).
+func (qq *QueuedQuery) Result(fn func(br BatchResults) error) {
+	qq.fn = fn
+}
+
 // Batch queries are a way of bundling multiple queries together to avoid
 // unnecessary network round trips. A Batch must only be sent once.
 type Batch struct {


jackc:
I'm sorry, I don't understand what you are trying to do or what this Result function is doing. Can you provide additional explanation and/or example usage?

roelvandergoot:
I am trying to implement a DSL (YAML based) from which I would like to generate a  [JSON:API](https://jsonapi.org) server. Specifically, for the [atomic operation extension](https://jsonapi.org/ext/atomic) I have to translate Postgres errors back to JSON:API errors. Each atomic operation results in one or more SQL statements which I am collecting in a pgx.Batch. The `batch.Queue(stmt, args...).Exec(fn)` doesn't allow me to inspect postgres errors in `fn` where I know what operation I am working on. I only get a postgres error after `batchResults.Close()` and now I have to try and figure out which operation the error belongs to which is quite hard. By using `QueuedQuery.Result(fn)` I can handle the postgres error in `fn`, which greatly facilitates error processing.

An additional (smaller) problem with the existing QueuedQuery receiver functions is that if `fn` returns an error I will have to type cast the error returned from `batchResult.Close()` using a switch, which adds inefficiencies.

Let me know if I have to elaborate more.

roelvandergoot:
Also, if there is more I have to do to make my code or this merge request comply with your preferences, please let me know. And, of course, thank you for maintaining this great library! :)

jackc:
Okay, I think I see what you are trying to do now. Have you considered using:

```go
batch.Queue(stmt, args...).Query(func(rows Rows) error) {
  rows.Close()
  err := rows.Err()
  // do your error check and handling here
  return err
})

**Comment by roelvandergoot** [comment]
Hi Jack,

Sorry for getting back so late.

I looked at your suggested solution and it solves my direct problem I created a
merge request for. Thank you! However, It appears that the Query function I
provide appears it doesn't get called when the database reports an error (for
example a foreign key violation). I was hoping to be able to do the following:

```go
func (op Operation) Prepare(b *Batch) {
    b.Queue(stmt, args...)Query(op.Query)
}

func (op Operation) Query(rows pgx.Rows) error {
    rows.Close()
    if err := rows.Err(); err != nil {
        var pgErr *pgconn.PgError
        if errors.As(err, &pgErr) {
            switch pgErr.Code {
            case "23503":
                return MyError{PgErr: pgErr}
            // ...
            }
        }
    }
    ct := rows.CommandTag()
    if ct.RowsAffected() == 0 {
        return ResourceNotFoundError{}
    }
    return nil
}
```

In other words, I am trying to morph the Postgres error into an error more
convenient to the user. But I am also introducing extra errors by interpreting
the Postgres response (through `CommandTag()`).

Maybe my problem lies in how I am executing the batch:
```go
    if err := tx.SendBatch().Close(); err != nil {
        // err appears to be the original Postgres error and neither a MyError
        // or a ResourceNotFound error.
    }
```

[Generally, I call `SendBatch` receiver function on a `pgx.Tx`, but if the batch
only contains one SQL statement, I call the `SendBatch` receiver function on
`pgx.Pool`.]

Could you tell me whether I can do something like this in your `pgx` library? Or
do I have to resort to interpreting the error that comes out of the
`pgx.BatchResults`? That would somewhat complicate my design though.

Cheers,
Roel.

roelvandergoot:
I will keep on debugging and update this comment with my findings.

The `Queue`d statement makes it to the database and the database generates the expected foreign key constraint violation. The `BatchResult` appears to contain a `pgconn.PgError` after calling `Close()`. Obviously, the Query callback function didn't get called. Time to learn and use Delve!

jackc:
I think we could resolve this by making `Query` always call `fn`. e.g.

```go
func (qq *QueuedQuery) Query(fn func(rows Rows) error) {
	qq.fn = func(br BatchResults) error {
		rows, _ := br.Query()
		defer rows.Close()

		err := fn(rows)
		if err != nil {
			return err
		}
		rows.Close()

		return rows.Err()
	}
}
```

This would be a slight change in behavior, but wouldn't be a breaking change and it would avoid adding a new public method.

roelvandergoot:
That would work for me.

jackc:
Done in 7af80ae8a6c42716d3be04914ce77d70eaaaa9fb.

roelvandergoot:
Thank you!

roelvandergoot:
Jack,

Is there a reason for the `&& br.err == nil` condition?
```go
    // Read and run fn for all remaining items
    for br.err == nil && !br.closed && br.b != nil && br.qqIdx < len(br.b.queuedQueries) {
        if br.b.queuedQueries[br.qqIdx].fn != nil {
            err := br.b.queuedQueries[br.qqIdx].fn(br)
            if err != nil && br.err == nil {
                br.err = err
            }
        } else {
            br.Exec()
        }
    }
```

jackc:
> Is there a reason for the `&& br.err == nil` condition?

Yes. Once an error has occurred it doesn't make sense to try to run any more callbacks.

roelvandergoot:
I understand that you can't carry on once an error occurs and am not asking to change that behavior. I was hoping we could drop the `&& br.err == nil` because it allows the query callback from altering the error and getting that error back through `err := barchResult.Close()`. I think it would make for a nicer API.

jackc:
Hmm. Looking a little closer it appears the ` && br.err == nil` is unnecessary because the outer `for br.err == nil && !br.closed && br.b != nil && br.qqIdx < len(br.b.queuedQueries) {` should have stopped it from ever being called.

So I guess it could / should be removed... but are you saying it actually would change behavior?

roelvandergoot:
The behavior would change and change for the better according to me. By omitting the ` && br.err == nil`, a query callback function can now replace the `pgErr` with a personalized error (by returning it from the query callback function) and that personalized error will be returned in the `err := batchResult.Close()`.  I for one would welcome the removal of the ` && br.err == nil` condition because I would like to personalize the error.

jackc:
I think I see what you mean now. The case that this would change is when an error is encountered before the callback function is called. Since `Query` callbacks are now always called it has a chance to override the error.

Changed in be79f1c8f598cdc93eb77158fe631e96bc591b7d.

roelvandergoot:
Thank you, kindly!

Issue #1552, opened 2023-03-14
The copy from function runs incorrectly on some machines


linlexing:
Some computers running the same version of the Windows operating system will prompt a 22P04 error when running copyfrom, indicating that the number of data fields received is not the number in the table, and the number of error fields prompted is different each time. Bad computers always go wrong, and good computers always work.
Spent a lot of time testing each version and found that version 898891a6eea762a65bb94b383fb9424ebbcc9d88 caused the problem.
It was found that the computer with the problem, with the earlier version, can run the copyfrom function, but the execution speed is obviously slower than normal. The submitted code is about time control. It is guessed that the death line time is too short, and the network data is not transmitted completely.

copy from source code:

```go
		conn, err := t.db.Conn(context.Background())
		if err != nil {
			return err
		}
		defer conn.Close()
		if err := conn.Raw(func(driverConn interface{}) error {
			pgxConn := driverConn.(*stdlib.Conn).Conn() // conn is a *pgx.Conn
			cnt, err := pgxConn.CopyFrom(context.Background(), pgx.Identifier{strings.ToLower(tabName)},
				cols, pgx.CopyFromRows(t.cacheRows))
			if err != nil {
				// saveErrorData()
				return err
			}
			if cnt != int64(len(t.cacheRows)) {
				return fmt.Errorf("copy from %d rec,but success %d rec", len(t.cacheRows), cnt)
			}
			return nil
		}); err != nil {
			return err
		}
```
**Version**
 - windows:windows2012 server r2 (6.3.9600)
 - Go: go1.20.2 windows/amd64
 - PostgreSQL: postgresql-15.1-1-windows-x64
 - pgx: 898891a6eea762a65bb94b383fb9424ebbcc9d88

jackc:
Try the master branch. That has true non-blocking IO on Windows instead of the fake non-blocking IO in previous versions.

That said, I really can't figure how 898891a could cause the error you are receiving. A timeout or a deadlock perhaps, but that error implies a corrupted data stream.

linlexing:
In this version c09ddaf4408eabc36322c8696cd93a646e93e976 , the first database operation hangs and does not respond.

linlexing:
In this version c09ddaf4408eabc36322c8696cd93a646e93e976, the first database operation hangs and does not respond.
In this version 9ae852eb583d2dced83b1d2ffe1c8803dda2c92e(last master), the copy from operation hangs and does not respond.

jackc:
[...]

PR #1555, opened 2023-03-18
[WIP] Fix Windows non-blocking I/O

codercms:
Created based on discussion here: https://github.com/jackc/pgx/pull/1525#pullrequestreview-1344511991

Fixes https://github.com/jackc/pgx/issues/1552

### DIFF
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -131,14 +131,32 @@ func (c *NetConn) SetBlockingMode(blocking bool) error {
 		return nil
 	}
 
+	// Prevent concurrent SetBlockingMode calls
+	c.nbOperMu.Lock()
+	defer c.nbOperMu.Unlock()
+
+	// Guard against negative value (which should never happen in practice)
+	if c.nbOperCnt < 0 {
+		c.nbOperCnt = 0
+	}
+
 	if blocking {
-		// Not ready to exit from non-blocking mode, there are pending non-blocking operations
-		if c.nbOperCnt.Add(-1) > 0 {
+		// Socket is already in blocking mode
+		if c.nbOperCnt == 0 {
+			return nil
+		}
+
+		c.nbOperCnt--
+
+		// Not ready to exit from non-blocking mode, there is pending non-blocking operations
+		if c.nbOperCnt > 0 {
 			return nil
 		}
 	} else {
-		// Socket is already in non-blocking state
-		if c.nbOperCnt.Add(1) > 1 {
+		c.nbOperCnt++
+
+		// Socket is already in non-blocking mode
+		if c.nbOperCnt > 1 {
 			return nil
 		}
 	}
@@ -162,11 +180,13 @@ func (c *NetConn) SetBlockingMode(blocking bool) error {
 
 		// Revert counters inc/dec in case of error
 		if blocking {
-			c.nbOperCnt.Add(1)
+			c.nbOperCnt++
+			//c.nbOperCnt.Add(1)
 
 			return fmt.Errorf("cannot set socket to blocking mode: %w", retErr)
 		} else {
-			c.nbOperCnt.Add(-1)
+			c.nbOperCnt--
+			//c.nbOperCnt.Add(-1)
 
 			return fmt.Errorf("cannot set socket to non-blocking mode: %w", retErr)
 		}
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -44,7 +44,7 @@ func setSockMode(fd uintptr, mode sockMode) error {
 }
 
 func (c *NetConn) isDeadlineSet(dl time.Time) bool {
-	return !dl.IsZero() && dl != NonBlockingDeadline && dl != disableSetDeadlineDeadline
+	return !dl.IsZero() && !dl.Equal(NonBlockingDeadline) && !dl.Equal(disableSetDeadlineDeadline)
 }
 
 func (c *NetConn) isWriteDeadlineExpired() bool {
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -63,7 +63,7 @@ func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
 	c.nonblockWriteN = 0
 	c.nonblockWriteErr = nil
 
-	if c.isDeadlineSet(c.writeDeadline) && time.Now().After(c.writeDeadline) {
+	if c.isDeadlineSet(c.writeDeadline) && !time.Now().Before(c.writeDeadline) {
 		c.nonblockWriteErr = errors.New("i/o timeout")
 
 		return 0, c.nonblockWriteErr
@@ -109,7 +109,7 @@ func (c *NetConn) realNonblockingRead(b []byte) (n int, err error) {
 	c.nonblockReadN = 0
 	c.nonblockReadErr = nil
 
-	if c.isDeadlineSet(c.readDeadline) && time.Now().After(c.readDeadline) {
+	if c.isDeadlineSet(c.readDeadline) && !time.Now().Before(c.readDeadline) {
 		c.nonblockReadErr = errors.New("i/o timeout")
 
 		return 0, c.nonblockReadErr
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -23,6 +23,8 @@ var dll = syscall.MustLoadDLL("ws2_32.dll")
 // );
 var ioctlsocket = dll.MustFindProc("ioctlsocket")
 
+var deadlineExpErr = errors.New("i/o timeout")
+
 type sockMode int
 
 const (
@@ -45,6 +47,20 @@ func (c *NetConn) isDeadlineSet(dl time.Time) bool {
 	return !dl.IsZero() && dl != NonBlockingDeadline && dl != disableSetDeadlineDeadline
 }
 
+func (c *NetConn) isWriteDeadlineExpired() bool {
+	c.writeDeadlineLock.Lock()
+	defer c.writeDeadlineLock.Unlock()
+
+	return c.isDeadlineSet(c.writeDeadline) && !time.Now().Before(c.writeDeadline)
+}
+
+func (c *NetConn) isReadDeadlineExpired() bool {
+	c.readDeadlineLock.Lock()
+	defer c.readDeadlineLock.Unlock()
+
+	return c.isDeadlineSet(c.readDeadline) && !time.Now().Before(c.readDeadline)
+}
+
 // realNonblockingWrite does a non-blocking write. readFlushLock must already be held.
 func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
 	if c.nonblockWriteFunc == nil {
@@ -63,8 +79,8 @@ func (c *NetConn) realNonblockingWrite(b []byte) (n int, err error) {
 	c.nonblockWriteN = 0
 	c.nonblockWriteErr = nil
 
-	if c.isDeadlineSet(c.writeDeadline) && !time.Now().Before(c.writeDeadline) {
-		c.nonblockWriteErr = errors.New("i/o timeout")
+	if c.isWriteDeadlineExpired() {
+		c.nonblockWriteErr = deadlineExpErr
 
 		return 0, c.nonblockWriteErr
 	}
@@ -109,8 +125,8 @@ func (c *NetConn) realNonblockingRead(b []byte) (n int, err error) {
 	c.nonblockReadN = 0
 	c.nonblockReadErr = nil
 
-	if c.isDeadlineSet(c.readDeadline) && !time.Now().Before(c.readDeadline) {
-		c.nonblockReadErr = errors.New("i/o timeout")
+	if c.isReadDeadlineExpired() {
+		c.nonblockReadErr = deadlineExpErr
 
 		return 0, c.nonblockReadErr
 	}
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -132,7 +132,7 @@ func (c *NetConn) SetBlockingMode(blocking bool) error {
 	}
 
 	if blocking {
-		// No ready to exit from non-blocking mode, there are pending non-blocking operations
+		// Not ready to exit from non-blocking mode, there are pending non-blocking operations
 		if atomic.AddInt32(&c.nbOperCnt, -1) > 0 {
 			return nil
 		}
@@ -141,8 +141,6 @@ func (c *NetConn) SetBlockingMode(blocking bool) error {
 		if atomic.AddInt32(&c.nbOperCnt, 1) > 1 {
 			return nil
 		}
-
-		//fmt.Println("socket reverting to blocking mode")
 	}
 
 	mode := sockModeNonBlocking
--- a/internal/nbconn/nbconn_real_non_block_windows.go
+++ b/internal/nbconn/nbconn_real_non_block_windows.go
@@ -4,6 +4,7 @@ package nbconn
 
 import (
 	"errors"
+	"fmt"
 	"golang.org/x/sys/windows"
 	"io"
 	"syscall"
@@ -154,19 +155,20 @@ func (c *NetConn) SetBlockingMode(blocking bool) error {
 	})
 
 	if ctrlErr != nil || err != nil {
+		retErr := ctrlErr
+		if retErr == nil {
+			retErr = err
+		}
+
 		// Revert counters inc/dec in case of error
 		if blocking {
 			c.nbOperCnt.Add(1)
+
+			return fmt.Errorf("cannot set socket to blocking mode: %w", retErr)
 		} else {
 			c.nbOperCnt.Add(-1)
-		}
 
-		if ctrlErr != nil {
-			return ctrlErr
-		}
-
-		if err != nil {
-			return err
+			return fmt.Errorf("cannot set socket to non-blocking mode: %w", retErr)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


jackc:
It seems unfortunate for the concept of non-blocking mode to leak out of nbconn into the higher layers (e.g. the copy protocol support).

Would it be possible for this logic to be internal? I think that the only methods that could be affected from the outside are `Read`, `Flush`, and `BufferReadUntilBlock` (`Write` is not because writes are always buffered). It seems that at any given read or write call nbconn already knows whether it should be a blocking or non-blocking call and it could change mode or do nothing depending on the current mode.

codercms:
@jackc I have updated the PR according to you comment. Please check it again.
* `CopyFrom` works perfect with large amount of data
* But some of tests won't complete (just hangs forever):
<details>
<summary>Click Me</summary>

image
</details>

drakkan:
> **UPD**: I decided to use atomic counter to determine should socket be set to blocking or non-blocking mode, but it might not work on some CPUs or in some circumstances. Do you have any thoughts on this?

32 bit atomic should work on any CPU, anyway you can use [atomic.Int32](https://tip.golang.org/doc/go1.19#atomic_types).

codercms:
@drakkan thanks for the tip, I've updated PR

jackc:
This logic is tricky and there's a few cases I am concerned about.

What if `SetBlockingMode(false)` is called more times than `SetBlockingMode(true)`? It would seem that `nbOperCnt` could go negative and have unexpected behavior.

---

If `SetBlockingMode` can be entered concurrently then I think there may be some sort of race condition. What if there are two concurrent `SetBlockingMode(true)` calls? It would seem the following sequence of actions is possible:

1. G1 increments `nbOperCnt` from 0 to 1 and continues.
2. G2 increments `nbOperCnt` from 1 to 2 and returns.
3. G2 uses the socket thinking it is in non-blocking mode.
4. G1 calls setSockMode.

I suspect that the whole `SetBlockingMode` function needs to be guarded by a mutex.

---

It's possible that neither of these cases actually will occur based on how and where `SetBlockingMode` is called, but at the very least that would be leaving a landmine for the next person to update nbconn.

codercms:
@jackc I see your point, I've updated PR to use mutex instead of atomic

drakkan:
@codercms even with this patch applied some test cases hang as you also noticed

```
Z:\pgx>go test -v -timeout 30s -run ^TestInternalNonBlockingWriteWithDeadline$ github.com/jackc/pgx/v5/internal/nbconn
=== RUN   TestInternalNonBlockingWriteWithDeadline
=== RUN   TestInternalNonBlockingWriteWithDeadline/Pipe
=== RUN   TestInternalNonBlockingWriteWithDeadline/TCP_with_Fake_Non-blocking_IO
=== RUN   TestInternalNonBlockingWriteWithDeadline/TLS_over_TCP_with_Fake_Non-blocking_IO
=== RUN   TestInternalNonBlockingWriteWithDeadline/TCP_with_Real_Non-blocking_IO
    nbconn_test.go:320:
                Error Trace:    Z:\pgx\internal\nbconn\nbconn_test.go:320
                                                        Z:\pgx\internal\nbconn\nbconn_test.go:143
                Error:          An error is expected but got nil.
                Test:           TestInternalNonBlockingWriteWithDeadline/TCP_with_Real_Non-blocking_IO
=== RUN   TestInternalNonBlockingWriteWithDeadline/TLS_over_TCP_with_Real_Non-blocking_IO
panic: test timed out after 30s
running tests:
        TestInternalNonBlockingWriteWithDeadline (30s)
        TestInternalNonBlockingWriteWithDeadline/TLS_over_TCP_with_Real_Non-blocking_IO (30s)

goroutine 14 [running]:
testing.(*M).startAlarm.func1()
        C:/Program Files/Go/src/testing/testing.go:2241 +0x3b9
created by time.goFunc
        C:/Program Files/Go/src/time/sleep.go:176 +0x32

goroutine 1 [chan receive]:
testing.(*T).Run(0xc0000356c0, {0x4b90e3?, 0x23eb93?}, 0x4dc9c8)
        C:/Program Files/Go/src/testing/testing.go:1630 +0x405
testing.runTests.func1(0x6cde00?)
        C:/Program Files/Go/src/testing/testing.go:2036 +0x45
testing.tRunner(0xc0000356c0, 0xc00006fc88)
        C:/Program Files/Go/src/testing/testing.go:1576 +0x10b
testing.runTests(0xc0000530e0?, {0x6c7c80, 0xd, 0xd}, {0xc00007af68?, 0x100c00006fd10?, 0x6cd400?})
        C:/Program Files/Go/src/testing/testing.go:2034 +0x489
testing.(*M).Run(0xc0000530e0)
        C:/Program Files/Go/src/testing/testing.go:1906 +0x63a
main.main()
        _testmain.go:73 +0x1aa

goroutine 6 [chan receive]:
testing.(*T).Run(0xc000035860, {0x4b830a?, 0xc00000f2a8?}, 0xc00009c048)
        C:/Program Files/Go/src/testing/testing.go:1630 +0x405
github.com/jackc/pgx/v5/internal/nbconn_test.testVariants(0xc000049f60?, 0x4dc9c0)
        Z:/pgx/internal/nbconn/nbconn_test.go:107 +0x165
github.com/jackc/pgx/v5/internal/nbconn_test.TestInternalNonBlockingWriteWithDeadline(0x0?)
        Z:/pgx/internal/nbconn/nbconn_test.go:310 +0x25
testing.tRunner(0xc000035860, 0x4dc9c8)
        C:/Program Files/Go/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
        C:/Program Files/Go/src/testing/testing.go:1629 +0x3ea

goroutine 20 [runnable, locked to thread]:
syscall.SyscallN(0x0?, {0xc0000958f0?, 0x1e4db5acd98?, 0x1e4b5800eb8?})
        C:/Program Files/Go/src/runtime/syscall_windows.go:557 +0x109
syscall.Syscall9(0xc00001e4e0?, 0xc00001e450?, 0xc00001eb10?, 0xc00001eba0?, 0xc0000403b0?, 0xc00001ee10?, 0x20a93f?, 0xc0000520a0?, 0xc000035d40?, 0x0, ...)
        C:/Program Files/Go/src/runtime/syscall_windows.go:507 +0x78
syscall.WSASend(0x1b8?, 0xc000095a28?, 0x1, 0xc000095a1c?, 0x0, 0x26b185?, 0x0?)
        C:/Program Files/Go/src/syscall/zsyscall_windows.go:1311 +0xb9
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).realNonblockingWrite.func1(0xc000130780?)
        Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:51 +0x78
internal/poll.(*FD).RawWrite(0xc000130780, 0xc0000ac2c0)
        C:/Program Files/Go/src/internal/poll/fd_windows.go:1096 +0xbd
net.(*rawConn).Write(0xc000088038, 0xc0004d4000?)
        C:/Program Files/Go/src/net/rawconn.go:55 +0x45
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).realNonblockingWrite(0xc000002180, {0xc000ace000?, 0x471900?, 0x723160?})
        Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:61 +0x122
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).nonblockingWrite(0x524840?, {0xc000ace000?, 0x524840?, 0x723160?})
        Z:/pgx/internal/nbconn/nbconn.go:408 +0x27
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).flush(0xc000002180)
        Z:/pgx/internal/nbconn/nbconn.go:317 +0x212
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).Flush(0x6cd400?)
        Z:/pgx/internal/nbconn/nbconn.go:289 +0xa5
github.com/jackc/pgx/v5/internal/nbconn.(*TLSConn).Flush(0x525a08?)
        Z:/pgx/internal/nbconn/nbconn.go:533 +0x1d
github.com/jackc/pgx/v5/internal/nbconn_test.TestInternalNonBlockingWriteWithDeadline.func1(0x525a08?, {0x527ac0, 0xc00006e0f0}, {0x0?, 0x0?})
        Z:/pgx/internal/nbconn/nbconn_test.go:319 +0x11e
github.com/jackc/pgx/v5/internal/nbconn_test.testVariants.func1(0xc000142820?)
        Z:/pgx/internal/nbconn/nbconn_test.go:143 +0x534
testing.tRunner(0xc000142820, 0xc00009c048)
        C:/Program Files/Go/src/testing/testing.go:1576 +0x10b
created by testing.(*T).Run
        C:/Program Files/Go/src/testing/testing.go:1629 +0x3ea

goroutine 13 [runnable, locked to thread]:
syscall.SyscallN(0x1e4db5fa9b8?, {0xc00156bcb0?, 0x1e4b580a398?, 0x1e4b5800a28?})
        C:/Program Files/Go/src/runtime/syscall_windows.go:557 +0x109
syscall.Syscall9(0x6375b0?, 0x6d528?, 0x1c78?, 0x11b48?, 0xc00156bd58?, 0x1f1719?, 0x723ea0?, 0x0?, 0xc00156bd98?, 0x0, ...)
        C:/Program Files/Go/src/runtime/syscall_windows.go:507 +0x78
syscall.WSARecv(0x1e4db5fa9b8?, 0x1c00156be10?, 0x1, 0x0?, 0xc00156bde8?, 0x26b125?, 0x67?)
        C:/Program Files/Go/src/syscall/zsyscall_windows.go:1295 +0xb8
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).realNonblockingRead.func1(0xc000130780?)
        Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:91 +0x87
internal/poll.(*FD).RawRead(0xc000130780, 0xc0000ac000)
        C:/Program Files/Go/src/internal/poll/fd_windows.go:1067 +0xda
net.(*rawConn).Read(0xc000088038, 0xc00001e840?)
        C:/Program Files/Go/src/net/rawconn.go:43 +0x45
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).realNonblockingRead(0xc000002180, {0xc0003f6000?, 0xc000002180?, 0xc00156bee8?})
        Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:101 +0x122
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).nonblockingRead(0x2000?, {0xc0003f6000?, 0x524840?, 0x723160?})
        Z:/pgx/internal/nbconn/nbconn.go:441 +0x27
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).BufferReadUntilBlock(0xc000002180)
        Z:/pgx/internal/nbconn/nbconn.go:358 +0xaf
github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).bufferNonblockingRead.func1()
        Z:/pgx/internal/nbconn/nbconn.go:382 +0x3b
created by github.com/jackc/pgx/v5/internal/nbconn.(*NetConn).bufferNonblockingRead
        Z:/pgx/internal/nbconn/nbconn.go:380 +0x10a
FAIL    github.com/jackc/pgx/v5/internal/nbconn 30.596s
FAIL
```

take a look here

```
Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:91 +0x87
```

and

```
Z:/pgx/internal/nbconn/nbconn_real_non_block_windows.go:51 +0x78
```

should we avoid concurrent reads and writes on Windows?

codercms:
@jackc @drakkan I've added some deadline simulation, please have a look on it  
All tests passed, except 2 cases:
* TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO (Error:      	Received unexpected error: local error: tls: bad record MAC)
* TestInternalNonBlockingWriteWithDeadline/TCP_with_Real_Non-blocking_IO (Error:      	An error is expected but got nil.)

internal/nbconn/nbconn_real_non_block_windows.go
```diff
@@ -39,36 +43,48 @@ func setSockMode(fd uintptr, mode sockMode) error {
 	return nil
 }
 
+func (c *NetConn) isDeadlineSet(dl time.Time) bool {
+	return !dl.IsZero() && dl != NonBlockingDeadline && dl != disableSetDeadlineDeadline
```
drakkan, line 47:
```
return !dl.IsZero() && !dl.Equal(NonBlockingDeadline) && !dl.Equal(disableSetDeadlineDeadline)
```

drakkan:
> @jackc @drakkan I've made one more update and `TestInternalNonBlockingWriteWithDeadline/TCP_with_Real_Non-blocking_IO` now runs successfully.
> 
> **UPD:** @jackc should c.readDeadline / c.writeDeadline readings be guarded by mutex?

It looks good, thanks! If test cases cover all possible usage we should be fine now. The only failing test case is `TestInternalNonBlockingWrite/TLS_over_TCP_with_Fake_Non-blocking_IO` but this failure is unrelated to this PR.

To be honest, I'm not happy with all this platform-specific, complicated and hard-to-maintain code (in general, not only related to this PR), it can hide subtle and hard-to-diagnose bugs (e.g. the remaining failed test case)

internal/nbconn/nbconn_real_non_block_windows.go
codercms: @drakkan is this necessary?  This vars declared as special values here - https://github.com/jackc/pgx/blob/master/internal/nbconn/nbconn.go#L35

And there's also direct comparison with this special values in other parts of code - https://github.com/jackc/pgx/blob/master/internal/nbconn/nbconn.go#L259

codercms:
@drakkan I agree, that platform specific code isn't a good approach, but unfortunately GoLang's stdlib doesn't have non-blocking I/O for Windows.

It's actually normal in C/C++ world to have platform specific code, especially for sockets, because Windows sockets API is different from sockets API in UNIX (and behavior is different too). In UNIX sockets are files (actually in UNIX everything is file), but in Windows files are only files and nothing more.

Projects like cURL, nginx, boost.asio (async I/O library for C++) have different network layer implementation for Windows.

UNIX systems have advanced socket polling mechanisms like kqueue, epoll, but Windows doesn't have kqueue nor epoll, it have [WaitForMultipleObjects](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects).
UNIX systems have `fcntl`/`ioctl` syscalls to set socket to non-blocking mode, Windows have `ioctlsocket` syscall.
In general, there are many differences between sockets in UNIX-like systems and Windows.

internal/nbconn/nbconn_real_non_block_windows.go
drakkan: > @drakkan is this necessary? This vars declared as special values here - https://github.com/jackc/pgx/blob/master/internal/nbconn/nbconn.go#L35
> 
> And there's also direct comparison with this special values in other parts of code - https://github.com/jackc/pgx/blob/master/internal/nbconn/nbconn.go#L259

it does not matter if the location is the same, see [here](https://pkg.go.dev/time#Time.Equal). The best practice is to use Equal

drakkan:
> @drakkan I agree, that platform specific code isn't a good approach, but unfortunately GoLang's stdlib doesn't have non-blocking I/O for Windows.
> 
> It's actually normal practice in C/C++ world to have platform specific code, especially for sockets, because Windows sockets API is different from sockets API in UNIX (and behavior is different too). In UNIX-like systems - sockets are files (actually in UNIX everything is file), but in Windows files are only files and nothing more.
> 
> Projects like cURL, nginx, boost.asio (async I/O library for C++), libuv (event loop and non-blocking I/O library, used in Node.js) have different network layer implementation for Windows.
> 
> UNIX-like systems have advanced socket polling mechanisms like kqueue, epoll, but Windows doesn't have kqueue nor epoll, it has [WaitForMultipleObjects](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects). UNIX-like systems have `fcntl`/`ioctl` syscalls to set socket to non-blocking mode, Windows have `ioctlsocket` syscall. In general, there are many differences between sockets in UNIX-like systems and Windows.

First of all I appreciate a lot your efforts.
Sure in C/C++ is normal, but not in Go. I wonder if we can just use goroutines so the code will work the same on every platforms. Anyway this issue seems fixed. The only remaining test failure should not matter now that non blocking I/O works also on Windows

codercms:
@drakkan for sure we can use just goroutines and simulate non-blocking I/O, but see this comment - https://github.com/jackc/pgx/pull/1525#discussion_r1141023089
>  I'm half tempted to try using goroutines and normal Go I/O again. I tried it a few years ago and the performance hit was significant, but maybe modern Go has higher performance...

internal/nbconn/nbconn_real_non_block_windows.go
codercms: @drakkan thanks for the tip, fixed in https://github.com/jackc/pgx/pull/1555/commits/fbe7692f66730900f3a2f9ca589daa66bcd57d6d

drakkan:
> I think this is in a mergeable state. But I don't plan on tagging a new release for a while, I think it would be good to let the more adventurous people who run master give us some more real world testing.
> 
> @drakkan
> 
> > To be honest, I'm not happy with all this platform-specific, complicated and hard-to-maintain code (in general, not only related to this PR), it can hide subtle and hard-to-diagnose bugs (e.g. the remaining failed test case)
> 
> You and me both. I wrote almost all of nbconn except for this new Windows support and I totally agree it is very complicated and tricky. Unfortunately, I have been unable to find a better way. (TLS support greatly complicates it as well. )
> 
> > I wonder if we can just use goroutines so the code will work the same on every platforms. pgx v4 and lib/pq does not have non-blocking I/O and seem to work fine.
> 
> There were a few reasons for introducing non-blocking IO. One was to be able to efficiently test if a connection was live before sending a query. If a connection has been closed on the server end it is sometimes still possible to write to the connection. The connection closed error is not received until the connection is read from. This causes a problem where is is impossible to say if a insert/update/delete statement actually was executed. By testing the connection first this occurrence can be minimized.
> 
> In addition, the CopyFrom logic is much simpler with non-blocking I/O.

@jackc thanks for the detailed explanation. I updated SFTPGo to use the latest version and I did several tests on a Windows VM. It works fine for me but SFTPGo is not a database intesive application. @codercms thanks for your work!

PR #1558, opened 2023-03-24
#1515 add pgx.RowToStructByNameLax

### LINKED ISSUE #1515: pgx.RowToStructByName: skip unknown struct fields
Func `RowToStructByName[T any](row CollectableRow)` requires that: "T must have the same number of named public fields as row has fields".
That means that I must define different Go types for different requests to the same DB table. I might have table `books` with columns: `id, name and synopsis`. Sometimes I want to query synopsis and sometimes I don't. Hence I would like to define type as:
```
type book struct {
	Id        int
	Name string `db:"name"`
	Synopsis     string `db:"synopsis"`
}

```

And my query could be `SELECT id, name FROM books` or `SELECT id, name, synopsis FROM books` depends on place in application.
So I would like to have an option or a wrapper function that allows to scan row to struct and silently skip unknown fields.

ronaudinho:
closes #1515 

tests are pretty much a copy of `TestRowToStructByName` with slight modifications (changing a test case allowing struct field to exist without corresponding query result), but I have a hunch that I am missing a test case here.

I am also not sure if the approach is correct, let me know if you have different implementation in mind.


rows.go
```diff
@@ -638,7 +657,13 @@ func (rs *namedStructRowScanner) appendScanTargets(dstElemValue reflect.Value, s
 				colName = sf.Name
 			}
 			fpos := fieldPosByName(fldDescs, colName)
-			if fpos == -1 || fpos >= len(scanTargets) {
+			if fpos == -1 {
```
ronaudinho, line 660: pretty much the only difference is adding this lax check, allowing struct field to exist without corresponding query result

rows.go
```diff
@@ -638,7 +657,13 @@ func (rs *namedStructRowScanner) appendScanTargets(dstElemValue reflect.Value, s
...
+				if rs.lax {
+					continue
+				}
+				return nil, fmt.Errorf("cannot find field %s in returned row", colName)
+			}
+			if fpos >= len(scanTargets) && !rs.lax {
```
ronaudinho, line 666: ah perhaps I am missing case where `fpos >= len(scanTargets) && rs.lax`, but I don't see how `fpos >= len(scanTargets)` is possible in the first place.

jackc:
I think this should work.

PR #1565, opened 2023-04-03
add extra tests for `RowToStructByNameLax`

danicc097:
NOTE: currently, these tests fail for some reason (`user.user_id` is zero valued, which can be reproduced with `int`, `string`, changing column name to a unique one instead of `user_id`...)

I don't know if it's a known problem, let me know if you want me to open accompanying issue

### DIFF
--- a/rows_test.go
+++ b/rows_test.go
@@ -736,6 +736,42 @@ func TestRowToStructByNameLaxEmbeddedStruct(t *testing.T) {
 	})
 }
 
+func TestRowToStructByNameLaxJSON(t *testing.T) {
+	type AnotherTable struct{}
+	type User struct {
+		UserID int    `json:"userId" db:"user_id"`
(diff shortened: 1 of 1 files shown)


danicc097:
I was incorrectly using the previous `row_to_json` instead of plain `row` now, which obviously would need the json tag to be the same as the db tag. Regardless, this extra test won't hurt (which is still failing CI somehow)

danicc097:
cockroachdb messes up the [`row`](https://www.cockroachlabs.com/docs/v22.2/scalar-expressions#tuple-constructors) call for some reason I'm unfamiliar with and yields `("(1,""John Doe"")")`, but works with `(users.*)`. It does look like a bug in cockroachdb to me although I don't use it, so I'm just skipping it

jackc:
LGTM

PR #1579, opened 2023-04-22
Support different bool string representations

lzakharov:
Add support for different bool string representations, mentioned in the [documentation](https://www.postgresql.org/docs/11/datatype-boolean.html).

### DIFF
--- a/pgtype/bool.go
+++ b/pgtype/bool.go
@@ -1,10 +1,12 @@
 package pgtype
 
 import (
+	"bytes"
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 type BoolScanner interface {
@@ -264,8 +266,8 @@ func (scanPlanTextAnyToBool) Scan(src []byte, dst any) error {
 		return fmt.Errorf("cannot scan NULL into %T", dst)
 	}
 
-	if len(src) != 1 {
-		return fmt.Errorf("invalid length for bool: %v", len(src))
+	if len(src) == 0 {
+		return fmt.Errorf("cannot scan empty string into %T", dst)
 	}
 
 	p, ok := (dst).(*bool)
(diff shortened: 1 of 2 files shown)


jackc:
PostgreSQL does allow multiple formats for bools on *input*, but AFAIK it only ever uses `t`/`f` on output. I don't see how this code would be called. Do you have an example of how it would be used?

lzakharov:
It's used in binary encoding, where the value is first scanned in text format: https://github.com/jackc/pgx/blob/6defa2a607fd64cb5f792dfef0cf17476bfe7d57/pgtype/pgtype.go#L1453-L1467

Original problem described in the linked issue.

pgtype/bool.go
```diff
@@ -309,9 +316,28 @@ func (scanPlanTextAnyToBoolScanner) Scan(src []byte, dst any) error {
...
+
+	return s.ScanBool(Bool{Bool: v, Valid: true})
+}
+
+// https://www.postgresql.org/docs/11/datatype-boolean.html
+func planTextToBool(src []byte) (bool, error) {
+	s := string(bytes.ToLower(bytes.TrimSpace(src)))
+
+	switch {
+	case strings.HasPrefix("true", s), strings.HasPrefix("yes", s), s == "on", s == "1":
```
drakkan, line 336: This will match `t`, `tr`, `tru`, `true`, is this by purpose?

lzakharov: From docs:
> Unique prefixes of these strings are also accepted

pgtype/bool.go
```diff
@@ -309,9 +316,28 @@ func (scanPlanTextAnyToBoolScanner) Scan(src []byte, dst any) error {
...
+	if err != nil {
+		return err
+	}
+
+	return s.ScanBool(Bool{Bool: v, Valid: true})
+}
+
+// https://www.postgresql.org/docs/11/datatype-boolean.html
+func planTextToBool(src []byte) (bool, error) {
+	s := string(bytes.ToLower(bytes.TrimSpace(src)))
```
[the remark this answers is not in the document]
lzakharov: Agree, but this conversion helps to avoid `[]byte` around the string constants.

Issue #1097, opened 2021-10-07
Per-Connection Metadata?


jakedt:
I am trying to associate some metadata with each connection in a connection pool. Specifically, I would like to store the results of a query that should always be stable across a particular connection. Ideally, this would execute one query when the connection was established, and then the results of that query would always be associated with the connection, and available to the `BeforeAcquire` function when the connection is going to be leased from the pool.

I've tried doing this with `AfterConnect` but I can't find anywhere on the connection to stash metadata:

```go
	poolConfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
		var someValue int

		err := pgconn.QueryRow(ctx, someQuery).Scan(&someValue)
		if err != nil {
			return err
		}

		// What do I do with someValue now?
		
		return nil
	}
```

I tried using the `conn` pointer as a map key to save my `someValue`, but without an `AfterDestroy` hook, I can't ensure that this map would stay valid as connections cycled through the pool.

I also tried doing the whole thing with closures by using the `BeforeConnect` function to install a specialized `AfterConnect` function that is already specialized for each connection:

```go
	poolConfig.BeforeConnect = func(ctx context.Context, config *pgx.ConnConfig) error {
		var someValue int

		config.AfterConnect = func(ctx context.Context, pgconn *pgconn.PgConn) error {
			// Why is this a pgconn.PgConn and not a pgx.Conn?
			return nil
		}

		return nil
	}
```

With this method I have to drop down to the pgconn interface, which seems like an unnecessary loss of convenience having to replicate the work of `Query()` and `Scan()`.

Do you have any ideas for how I could accomplish this? I desperately need to avoid querying the database every time I check-out a connection.

PR #1587, opened 2023-05-01
Add a BeforeClose hook to pgxpool.Pool

ecordell:
I needed to track per-connection metadata (similar to what is described in https://github.com/jackc/pgx/issues/1097)

In order to clean up unneeded metadata after a connection has been closed, I added a `BeforeClose` hook, similar to the existing hooks. In the linked issue @jackc suggested that an `AfterClose` hook would be accepted - I went with a `BeforeClose` hook to ensure that there was no risk of connection data being cleared out during the lifetime of the hook.

I also un-skipped tests for cockroach that rely on backend PIDs, since cockroach returns PIDs now (as of v22.1).

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -85,6 +85,7 @@ type Pool struct {
 	afterConnect          func(context.Context, *pgx.Conn) error
 	beforeAcquire         func(context.Context, *pgx.Conn) bool
 	afterRelease          func(*pgx.Conn) bool
+	beforeClose           func(*pgx.Conn)
 	minConns              int32
 	maxConns              int32
 	maxConnLifetime       time.Duration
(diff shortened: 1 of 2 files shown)


ecordell:
@jackc have you had a chance to take a look at this? thanks in advance!

ecordell:
Here's an example of how it's being used: https://github.com/authzed/spicedb/blob/37555d522869621b09e359fc6f9259319f55386a/internal/datastore/crdb/pool/pool.go#L106-L114

With the `BeforeClose` hook, we can track which connections are to which nodes in a cockroach cluster behind a load balancer.

jackc:
I think this is a reasonable change. The only thing that gives me pause is the signature of the hook function.

First, should it accept a context? This would allow it to use the overall destructor timeout. It also could allow smuggling additional values into the hook function without changing the signature should that ever be necessary.

Second, should it return an error? I suppose the pool can't do anything with the error so maybe not, but it still feels funny to have a function that could run arbitrary, important logic not to return an error. 

---

I'm actually fine with merging it as is, but I'd like you to consider whether either of these changes make sense and give me your conclusion before I do.

ecordell:
> First, should it accept a context? This would allow it to use the overall destructor timeout. It also could allow smuggling additional values into the hook function without changing the signature should that ever be necessary.

I played with a version that had context, but the only context it could have access to is the `Destructor`s timeout context, which didn't seem very useful to me:

```go
func(value *connResource) {
    ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
    conn := value.conn
    beforeClose(ctx)
```

>Second, should it return an error? I suppose the pool can't do anything with the error so maybe not, but it still feels funny to have a function that could run arbitrary, important logic not to return an error.

Same thing here - I played with a version that could return an error, but wasn't sure what pgxpool should do with it if it returned one other than log it (which the hook could also do before returning).

Happy to make either change if you think they're valuable, I just couldn't convince myself they were.

jackc:
I'll merge it as is then. I wasn't sure either way, and just wanted to make sure someone else had thought about it as well.

PR #1592, opened 2023-05-11
Fix hstore NULL versus empty

evanj:
When running queries with the hstore type registered, and with simple mode queries, the scan implementation does not correctly distinguish between NULL and empty. Fix the implementation and add a test to verify this.


jackc:
Thanks!

PR #1595, opened 2023-05-12
Correctly handle bool type aliases

### LINKED ISSUE #1593: Bool opaque type cannot be encoded/decoded
**Describe the bug**

Encoding and decoding doesn't work for bool opaque types (type definition). It works for strings and integers though.
```golang
type TrafficDirection bool

const (
	RightHandTraffic TrafficDirection = false
	LeftHandTraffic  TrafficDirection = true
)


db.Exec(ctx, "INSERT INTO table (bool_col) VALUES ($1)", LeftHandTraffic)

var dir TrafficDirection
db.QueryRow(ctx, "SELECT bool_col from table").Scan(&dir)
```

Error:
```
// encode error
unable to encode false into binary format for bool (OID 16): cannot find encode plan

// decode error
can't scan into dest[0]: cannot scan bool (OID 16) in binary format into *main.TrafficDirection false
```

**To Reproduce**

Create a table (used only for testing insertion)
```sql
create table bool_test (direction boolean);
```

Run the following code:

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
)

type TrafficDirection bool

const (
	RightHandTraffic TrafficDirection = false
	LeftHandTraffic  TrafficDirection = true
)

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
        // Encode
	ct, err := conn.Exec(context.Background(), "INSERT INTO bool_test (direction) VALUES ($1)", LeftHandTraffic)
	if err != nil {
		log.Fatal(err)
	}
	log.Println(ct)

        // Decode
	var dir TrafficDirection

	err = conn.QueryRow(context.Background(), "SELECT 2 > 1").Scan(&dir)
	if err != nil {
		log.Fatal(err)
	}
	log.Println("Direction", dir)
}
```

**Expected behavior**
Bools are encoded and decoded. In pgx v4 bool opaque types are handled by default.

**Actual behavior**
Error is thrown when scanning to an opaque type and encoding it.

**Version**
 - Go:  `go1.20.4 darwin/amd64`
 - PostgreSQL: ` PostgreSQL 13.1 on x86_64-pc-linux-musl, compiled by gcc (Alpine 9.3.0) 9.3.0, 64-bit`
 - pgx: `v5.3.1`

wichert:
The bool type was missing in the `kindToTypes` map, which breaks handling of type aliases for bool.

fixes #1593

### DIFF
--- a/pgtype/convert.go
+++ b/pgtype/convert.go
@@ -64,6 +64,9 @@ func underlyingNumberType(val any) (any, bool) {
 	case reflect.String:
 		convVal := refVal.String()
 		return convVal, reflect.TypeOf(convVal) != refVal.Type()
+	case reflect.Bool:
+		convVal := refVal.Bool()
+		return convVal, reflect.TypeOf(convVal) != refVal.Type()
 	}
 
 	return nil, false
--- a/pgtype/pgtype.go
+++ b/pgtype/pgtype.go
@@ -666,6 +666,7 @@ var elemKindToPointerTypes map[reflect.Kind]reflect.Type = map[reflect.Kind]refl
 	reflect.Float32: reflect.TypeOf(new(float32)),
 	reflect.Float64: reflect.TypeOf(new(float64)),
 	reflect.String:  reflect.TypeOf(new(string)),
+	reflect.Bool:    reflect.TypeOf(new(bool)),
 }
 
 type underlyingTypeScanPlan struct {
--- a/pgtype/pgtype.go
+++ b/pgtype/pgtype.go
@@ -1519,6 +1519,7 @@ var kindToTypes map[reflect.Kind]reflect.Type = map[reflect.Kind]reflect.Type{
 	reflect.Float32: reflect.TypeOf(float32(0)),
 	reflect.Float64: reflect.TypeOf(float64(0)),
 	reflect.String:  reflect.TypeOf(""),
+	reflect.Bool:    reflect.TypeOf(false),
 }
 
 type underlyingTypeEncodePlan struct {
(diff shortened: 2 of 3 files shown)


jackc:
I would have expected this to work, but I am getting test locally on PG 15 as well as in CI on CockroachDB.

jackc:
Yes, I meant to say test errors. 

```
--- FAIL: TestEncodeTypeRename (0.04s)
    --- FAIL: TestEncodeTypeRename/cache_statement (0.01s)
        values_test.go:889: Failed with type rename: can't scan into dest[11]: cannot scan bool (OID 16) in binary format into *pgx_test._bool
    --- FAIL: TestEncodeTypeRename/cache_describe (0.01s)
        values_test.go:889: Failed with type rename: can't scan into dest[11]: cannot scan bool (OID 16) in binary format into *pgx_test._bool
    --- FAIL: TestEncodeTypeRename/describe_exec (0.01s)
        values_test.go:889: Failed with type rename: can't scan into dest[11]: cannot scan bool (OID 16) in binary format into *pgx_test._bool
    --- FAIL: TestEncodeTypeRename/exec (0.01s)
        values_test.go:889: Failed with type rename: cannot use unregistered type pgx_test._bool as query argument in QueryExecModeExec
    --- FAIL: TestEncodeTypeRename/simple_protocol (0.01s)
        values_test.go:889: Failed with type rename: can't scan into dest[11]: cannot scan bool (OID 16) in text format into *pgx_test._bool
```

The code looks correct at an initial glance, but the failures do seem to be related to the new functionality.

wichert:
@sanohin thanks for fixing the test failures!

PR #1597, opened 2023-05-12
pgtype.TestTextCodeACLItem: Use built-in role instead of postgres

evanj:
The role "postgres" is not guaranteed to exist. When creating a new database with initdb, the default is the current user's name. Most Linux distributions and many "production" Postgres installs will use the name "postgres", but it is not the default. The "pg_database_owner" role on the other hand is predefined and is guaranteed to exist:
https://www.postgresql.org/docs/current/predefined-roles.html

This should make the tests easier to run for some users (like me!)

See initdb documentation for the --username flag for details: https://www.postgresql.org/docs/current/app-initdb.html

### DIFF
--- a/pgtype/text_test.go
+++ b/pgtype/text_test.go
@@ -83,16 +83,18 @@ func TestTextCodecBPChar(t *testing.T) {
 // ACLItem is used for PostgreSQL's aclitem data type. A sample aclitem
 // might look like this:
 //
-//	postgres=arwdDxt/postgres
+//	pg_database_owner=arwdDxt/pg_database_owner
 //
 // Note, however, that because the user/role name part of an aclitem is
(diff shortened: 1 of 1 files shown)


jackc:
It looks like this role was only added in PG 14. It might be possible to create a connection before the test is run to see what current user is and use that in the test.

evanj:
Ah oops I should have tested this, sorry. I'll take another look a a better fix for this in a bit. It obviously isn't a huge deal, it just took me a bit to understand that I needed to create the role before I could run the pgx tests ... Thanks for the fast response!

evanj:
I'm going to close this: this is clearly not a big deal, and I don't have time to investigate a better fix now. Thanks!

PR #1598, opened 2023-05-12
pgtype: Change hstore tests to execute "create extension hstore"

evanj:
The hstore extension can be loaded by any user with "create" priviledges. The test already queries for the hstore OID, so we might as well load the extension if it does not exist. This removes some required setup for running the tests, so should make it a bit easier for others to run them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,6 @@ Create and setup a test database:
 ```
 export PGDATABASE=pgx_test
 createdb
-psql -c 'create extension hstore;'
 psql -c 'create domain uint64 as numeric(20,0);'
 ```
 
--- a/pgtype/hstore_test.go
+++ b/pgtype/hstore_test.go
@@ -51,13 +51,34 @@ func isExpectedEqMapStringPointerString(a any) func(any) bool {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


jackc:
What happens if the user doesn't have create extension privileges?

I've tried to keep the tests runnable by as low privileged users as possible, and anything that requires more privileges to be part of initial setup.

evanj:
Great point. I don't understand postgres permissions very well, so let me test it in a bit and report back. The tests already use "create schema", and my reading of the documentation is that if those are permitted, "create extension hstore" should be also:

* Hstore is a "trusted" extension: https://www.postgresql.org/docs/current/hstore.html
* `CREATE EXTENSION` for trusted extensions are permitted by users who have "create" permission on the database: https://www.postgresql.org/docs/current/sql-createextension.html
* I think "create" on the database is required to use "create schema": https://www.postgresql.org/docs/current/ddl-priv.html
* The tests execute `create schema`: https://github.com/jackc/pgx/blob/master/conn_test.go#L873

evanj:
Never mind: I think this doesn't make sense: if the test is going to require some "setup" SQL to be executed, we should just include this there, and that makes this unnecessary. If we were able to get rid of the "setup" commands entirely, this might make sense.

PR #1602, opened 2023-05-15
feat(error): add known error codes constants

aerialls:
This PR adds constants for known Postgres error codes. This allows more easily to perform the following checks without the need to look at the documentation for the associated codes.

```go
	if err != nil {
		var pgErr *pgconn.PgError
		if errors.As(err, &pgErr) {
			if pgErr.Code == pgconn.PgErrorUniqueViolationCode {
				// ...
			}
			
		}

		return nil, err
	}
```

The list was extracted from the [official postgres repository](https://github.com/postgres/postgres/blob/master/src/backend/utils/errcodes.txt).


jackc:
I think this repo will do what you want: https://github.com/jackc/pgerrcode

rubiagatra:
can we close it @aerialls?

cc: @jackc

PR #1603, opened 2023-05-15
pgtype/hstore_test.go: Add coverage for text protocol

evanj:
The existing test registers pgtype.Hstore in the text map, then uses the query modes that use the binary protocol. The existing test did not use the text parsing code. Add a version of the test that uses pgtype.Hstore as the input and output argument in all query modes, and tests it without registering the codec.


jackc:
LGTM

PR #1604, opened 2023-05-15
pgtype.parseHstore: Reject invalid input; Fix error messages

evanj:
The parseHstore function did not check the return value from p.Consume() after a ', ' sequence. It expects a doublequote '"' that starts the next key, but would accept any character. This means it accepted invalid input such as:

    "key1"=>"b", ,key2"=>"value"

Add a unit test that covers this case
Fix a couple of the nearby error strings while looking at this.

Found by looking at staticcheck warnings:

    pgtype/hstore.go:434:6: this value of end is never used (SA4006)
    pgtype/hstore.go:434:6: this value of r is never used (SA4006)

### DIFF
--- a/pgtype/hstore.go
+++ b/pgtype/hstore.go
@@ -428,12 +428,21 @@ func parseHstore(s string) (k []string, v []Text, err error) {
 				r, end = p.Consume()
 				switch {
 				case end:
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #1606, opened 2023-05-16
pgtype.Hstore: Fix quoting of whitespace; Add test

evanj:
Before this change, the Hstore text protocol did not quote keys or values containing non-space whitespace ("\r\n\v\t"). This causes inserts with these values to fail with errors like:

    ERROR: Syntax error near "r" at position 17 (SQLSTATE XX000)

The previous version also quoted curly braces ("{}"), but they don't seem to require quoting.

It is possible that it would be easier to just always quote the values, which is what Postgres does when encoding its text protocol, but this is a smaller change.


jackc:
Thanks!

PR #1612, opened 2023-05-17
pgconn.CancelRequest: Fix unix sockets: don't use RemoteAddr()

evanj:
The tests for cancelling requests were failing when using unix sockets. The reason is that net.Conn.RemoteAddr() calls getpeername() to get the address. For Unix sockets, this returns the address that was passed to bind() by the *server* process, not the address that was passed to connect() by the *client*. For postgres, this is always relative to the server's directory, so is a path like:

    ./.s.PGSQL.5432

Since it does not return the full absolute path, this function cannot connect, so it cannot cancel requests. To fix it, use the connection's config for Unix sockets. I think this should be okay, since a system using unix sockets should not have "fallbacks". If that is incorrect, we will need to save the address on PgConn.

Fixes the following failed tests when using Unix sockets:
```
--- FAIL: TestConnCancelRequest (2.00s)
    pgconn_test.go:2056:
          Error Trace:  /Users/evan.jones/pgx/pgconn/pgconn_test.go:2056
                              /Users/evan.jones/pgx/pgconn/asm_arm64.s:1172
          Error:        Received unexpected error:
                        dial unix ./.s.PGSQL.5432: connect: no such file or directory
          Test:         TestConnCancelRequest
    pgconn_test.go:2063:
          Error Trace:  /Users/evan.jones/pgx/pgconn/pgconn_test.go:2063
          Error:        Object expected to be of type *pgconn.PgError, but was <nil>
          Test:         TestConnCancelRequest
--- FAIL: TestConnContextCanceledCancelsRunningQueryOnServer (5.10s)
    pgconn_test.go:2109:
          Error Trace:  /Users/evan.jones/pgx/pgconn/pgconn_test.go:2109
          Error:        Received unexpected error:
                        timeout: context already done: context deadline exceeded
          Test:         TestConnContextCanceledCancelsRunningQueryOnServer
```

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -857,9 +857,28 @@ func (pgConn *PgConn) CancelRequest(ctx context.Context) error {
 	// the connection config. This is important in high availability configurations where fallback connections may be
 	// specified or DNS may be used to load balance.
 	serverAddr := pgConn.conn.RemoteAddr()
-	cancelConn, err := pgConn.config.DialFunc(ctx, serverAddr.Network(), serverAddr.String())
+	var serverNetwork string
+	var serverAddress string
+	if serverAddr.Network() == "unix" {
+		// for unix sockets, RemoteAddr() calls getpeername() which returns the name the
+		// server passed to bind(). For Postgres, this is always a relative path "./.s.PGSQL.5432"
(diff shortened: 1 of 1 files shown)


jackc:
LGTM

PR #1613, opened 2023-05-17
ParseConfig: default_query_exec_mode: Return arg in error

evanj:
If the default_query_exec_mode is unknown, the returned error previously was:

    invalid default_query_exec_mode: <nil>

This changes it to return the argument. Add a test that unknown modes fail to parse and include this string.


jackc:
LGTM

Issue #1456, opened 2022-12-29
Possible memory leak?


mohsalsaleem:
pgx version: v5.2.0
PostgreSQL version: 13.9

I have an application where I repeatedly query from a table in PostgreSQL. I noticed the amount of RAM being consumed by the application increasing linearly overtime.

Here is the pprof's result for top:

```bash
(pprof) top10
Showing nodes accounting for 39287.54kB, 100% of 39287.54kB total
Showing top 10 nodes out of 46
      flat  flat%   sum%        cum   cum%
32352.25kB 82.35% 82.35% 32352.25kB 82.35%  github.com/jackc/pgx/v5/pgconn.(*PgConn).convertRowDescription
 1537.50kB  3.91% 86.26%  1537.50kB  3.91%  runtime.allocm
 1280.36kB  3.26% 89.52% 33632.60kB 85.61%  github.com/jackc/pgx/v5.(*Conn).Prepare
 1024.41kB  2.61% 92.13%  1024.41kB  2.61%  runtime.malg
  532.26kB  1.35% 93.48%   532.26kB  1.35%  github.com/jackc/pgx/v5/pgtype.(*Map).RegisterDefaultPgType
  512.62kB  1.30% 94.79%   512.62kB  1.30%  regexp/syntax.(*compiler).inst (inline)
  512.07kB  1.30% 96.09%  1024.70kB  2.61%  github.com/aws/aws-sdk-go/aws/endpoints.init
  512.05kB  1.30% 97.39%   512.05kB  1.30%  context.(*cancelCtx).Done
  512.02kB  1.30% 98.70%   512.02kB  1.30%  runtime.gcBgMarkWorker
  512.01kB  1.30%   100%   512.01kB  1.30%  github.com/jackc/pgx/v5/internal/stmtcache.NextStatementName
```

Here is the png of the heap
![profile003](https://user-images.githubusercontent.com/6555290/209921733-89680c12-996e-4988-a24c-d33f55dd93c0.png)

The number of records queried every iteration is around 20-25 and a record contains around 100 columns.

Here is the pprof's result of top after 30 minutes from the previous result:

```
Showing nodes accounting for 56.92MB, 98.27% of 57.92MB total
Showing top 10 nodes out of 54
      flat  flat%   sum%        cum   cum%
   50.65MB 87.44% 87.44%    50.65MB 87.44%  github.com/jackc/pgx/v5/pgconn.(*PgConn).convertRowDescription
[...]

PR #1614, opened 2023-05-17
pgx.Conn: Fix memory leak: Delete items from preparedStatements

evanj:
Previously, items were never removed from the preparedStatements map. This means workloads that send a large number of unique queries could run out of memory. Delete items from the map when sending the deallocate command to Postgres. Add a test to verify this works.

Fixes https://github.com/jackc/pgx/issues/1456

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1326,6 +1326,7 @@ func (c *Conn) deallocateInvalidatedCachedStatements(ctx context.Context) error 
 
 	for _, sd := range invalidatedStatements {
 		pipeline.SendDeallocate(sd.Name)
+		delete(c.preparedStatements, sd.Name)
 	}
 
 	err := pipeline.Sync()
--- a/conn_internal_test.go
+++ b/conn_internal_test.go
@@ -1,0 +1,55 @@
+package pgx
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func mustParseConfig(t testing.TB, connString string) *ConnConfig {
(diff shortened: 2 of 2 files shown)


evanj:
I could not find a way to test this using the public interface, so I wrote a test that directly inspects the internals of this type. I guess the alternative might be to check the amount of memory being used? This seemed like the easiest way to do this to me, but I would be happy to take better ideas!

jackc:
LGTM - good catch and fix!

conn.go
```diff
@@ -1326,6 +1326,7 @@ func (c *Conn) deallocateInvalidatedCachedStatements(ctx context.Context) error
 
 	for _, sd := range invalidatedStatements {
 		pipeline.SendDeallocate(sd.Name)
+		delete(c.preparedStatements, sd.Name)
```
teror4uks, line 1329: this actually might not solve the "leak" issue, since Go won't deallocate memory after you delete a key from a map https://github.com/golang/go/issues/20135

evanj: The map *itself* won't shrink, but `delete()` does zero the keys and values in the map, so the `*pgconn.StatementDescription` value will be garbage collected. Additionally, `delete()` decrements the size of the map, which means its capacity will now be limited to the max number of prepared statements (with extra space for the map "fill factor"). This means this change should mean the amount of memory used by the c.preparedStatements is now fixed, rather than increasing without bound.

Additionally, my reproduction program on the original issue can run forever without running out of memory: https://github.com/jackc/pgx/issues/1456#issuecomment-1550373854

So: I'm pretty confident that this fix will resolve this specific bug at least. There could be others though! I would be happy to try to figure them out if we have a sample program that reproduces the problem.

Issue #1574, opened 2023-04-12
database/sql: storing driver.Value type string into type *time.Time


daxzhuo:
**Describe the bug**
database/sql: unsupported Scan, storing driver.Value type string into type *time.Time

**To Reproduce**
```
 CREATE TABLE vc (   
     ts timestamp
 );
INSERT INTO vc (ts) VALUES ('infinity');
```
```go
package main

import (
	"context"
	"log"
	"os"
        "time"
        "database/sql"

	_ "github.com/jackc/pgx/v5/stdlib"
)

func main() {
	db, err := sql.Open("pgx", os.Getenv("DATABASE_URL"))
	if err != nil {
		fmt.Fprintf(os.Stderr, "Unable to connect to database: %v\n", err)
		os.Exit(1)
	}
	defer db.Close()

        rows, err := db.Query("select * from vc")
	if err != nil {
		t.Error(err)
		return
	}

	for rows.Next() {
		columnTypes, _ := rows.ColumnTypes()
		for _, ct := range columnTypes {
			fmt.Println(ct.ScanType().Name())
		}
		ts := time.Time{}
		err = rows.Scan(&ts)
		if err != nil {
			t.Error(err)
		}
		fmt.Println(ts)
	}
}
```

**Expected behavior**
Return without error.

**Actual behavior**
Error in rows.Scan: unsupported Scan, storing driver.Value type string into type *time.Time

This is actually a simplify demo when using gorm library and pgx as driver, the default column type in go for timestamp is time.Time, but when the timestamp value set to infinity, the actual driver.Value is string type, hence the scan error happened.
I think this is a problem about how to represent infinity value in time.Time.
For code reference: https://github.com/jackc/pgx/blob/master/pgtype/timestamp.go#L58
**Version**
 - Go: go version go1.19.6 darwin/amd64
 - PostgreSQL: PostgreSQL 14.5 (Debian 14.5-2.pgdg110+2) on x86_64-pc-linux-gnu, compiled by gc
c (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
 - pgx: v5.3.0

**Additional context**

jackc:
I don't think there is any way around this. The Go `time.Time` type can't represent the PostgreSQL `infinity` value. However, `pgtype.Timestamp` and `pgtype.Timestamptz` can.

daxzhuo:
[...]

PR #1615, opened 2023-05-19
EnableInfinityTs for encoding / decoding infinity to got time.Time type

bastianrob:
This PR resolves https://github.com/jackc/pgx/issues/1574

Implemented only to TimestampCodec, TimestamptzCodec already handles infinity value

Changes:
- `pgtype.Map.EnableInfinityTs`

### DIFF
--- a/pgtype/builtin_wrappers.go
+++ b/pgtype/builtin_wrappers.go
@@ -419,12 +419,17 @@ func (w timeWrapper) DateValue() (Date, error) {
 	return Date{Time: time.Time(w), Valid: true}, nil
 }
 
-func (w *timeWrapper) ScanTimestamp(v Timestamp) error {
+func (w *timeWrapper) ScanTimestamp(v Timestamp, infinityTsEnabled bool) error {
 	if !v.Valid {
 		return fmt.Errorf("cannot scan NULL into *time.Time")
 	}
 
-	switch v.InfinityModifier {
+	infinityModifier := v.InfinityModifier
+	if infinityTsEnabled {
+		infinityModifier = Finite
+	}
+
+	switch infinityModifier {
 	case Finite:
 		*w = timeWrapper(v.Time)
 		return nil
--- a/pgtype/pgtype.go
+++ b/pgtype/pgtype.go
@@ -276,7 +276,7 @@ func NewMap() *Map {
 	m.RegisterType(&Type{Name: "text", OID: TextOID, Codec: TextCodec{}})
 	m.RegisterType(&Type{Name: "tid", OID: TIDOID, Codec: TIDCodec{}})
 	m.RegisterType(&Type{Name: "time", OID: TimeOID, Codec: TimeCodec{}})
-	m.RegisterType(&Type{Name: "timestamp", OID: TimestampOID, Codec: TimestampCodec{}})
+	m.RegisterType(&Type{Name: "timestamp", OID: TimestampOID, Codec: &TimestampCodec{}})
 	m.RegisterType(&Type{Name: "timestamptz", OID: TimestamptzOID, Codec: TimestamptzCodec{}})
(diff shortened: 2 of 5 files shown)


pgtype/builtin_wrappers.go
```diff
@@ -419,12 +419,17 @@ func (w timeWrapper) DateValue() (Date, error) {
 	return Date{Time: time.Time(w), Valid: true}, nil
 }
 
-func (w *timeWrapper) ScanTimestamp(v Timestamp) error {
+func (w *timeWrapper) ScanTimestamp(v Timestamp, infinityTsEnabled bool) error {
```
bastianrob, line 422: pass down infinityTsEnabled from codec

pgtype/timestamp.go
```diff
@@ -13,7 +13,7 @@ import (
 const pgTimestampFormat = "2006-01-02 15:04:05.999999999"
 
 type TimestampScanner interface {
-	ScanTimestamp(v Timestamp) error
+	ScanTimestamp(v Timestamp, infinityTsEnabled bool) error
```
jackc, line 16: This would be a breaking change. We can't do that in a without a major release.

pgtype/timestamp.go
```diff
@@ -170,31 +191,35 @@ func discardTimeZone(t time.Time) time.Time {
...
 
 	return nil
 }
 
-type scanPlanBinaryTimestampToTimestampScanner struct{}
+type scanPlanBinaryTimestampToTimestampScanner struct {
+	infinityTsEnabled bool
+	min               time.Time
+	max               time.Time
+}
```
jackc, line 216: In might be more efficient to store a pointer to the `Codec` rather than duplicate these fields.

pgtype/pgtype.go
```diff
@@ -418,6 +418,24 @@ func NewMap() *Map {
...
 }
 
+// EnableInfinityTs controls the handling of Postgres' "-infinity" and "infinity" "timestamp
+func (m *Map) EnableInfinityTs(negativeInfinity, positiveInfinity time.Time) error {
+	if negativeInfinity.Unix() >= positiveInfinity.Unix() {
+		return errors.New("invalid timerange between negative and positive infinity")
+	}
+
+	ts := m.nameToType["timestamp"]
+	tsc, _ := ts.Codec.(*TimestampCodec)
```
jackc, line 428: This would panic if a different `Codec` was registered.

jackc (changes requested):
The are a few issues that need to be resolved before this can be merged. A few are inline with the code and the broader ones are mentioned here.

First, I don't understand:

> Implemented only to TimestampCodec, TimestamptzCodec already handles infinity value

I don't see any code in `TimestamptzCodec` that handles this mapping of infinite values to normal time.Time.

Whatever is done should be done for both.

Second, I don't think the `Map` should have the `EnableInfinityTs` method. The `Map` type does not know anything about any `Codec` implementations. It only works in terms of the `Codec` interface. This violates that boundary. Instead, I would suggest the user of this feature call `Map.RegisterType` with the properly configured `Timestamp(tz)Codec`.

 Next, do the values that `Infinity` and `-Infinity` are mapped to need to be configurable? I would have expected them to be hard-coded to the largest and smallest possible `time.Time`s.

The `TimestampScanner` interface can't be changed without breaking backwards compatibility. These conversions will actually have to occur in the `Codec`s and/or `ScanPlan`s and `EncodePlan`s.

Lastly, I think we should find a better name for the whole feature than `EnableInfinityTs`. That name makes sense in lib/pq where `Infinity` wasn't supported at all. But `Infinity` is already supported in pgx because it defines its own types instead of using `time.Time`. This feature is clamping / converting values to fit in `time.Time`. I would expect the feature / flag to be named something like `(Clamp|Convert|Map)Infinity...`

rubiagatra:
Hi, @bastianrob do you still want to continue the PR or want to close it?

PR #1621, opened 2023-05-27
Ci windows

drakkan:
https://github.com/jackc/pgx/commit/2c1973de4634a6a83d3ba09bdcde392aaf7cfb71 introduced a build error on Windows.

Try adding Windows to CI, to prevent this from happening again in the future (some tests fail on Windows locally ...)

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -896,17 +896,11 @@ func (pgConn *PgConn) CancelRequest(ctx context.Context) error {
 	binary.BigEndian.PutUint32(buf[4:8], 80877102)
 	binary.BigEndian.PutUint32(buf[8:12], uint32(pgConn.pid))
 	binary.BigEndian.PutUint32(buf[12:16], uint32(pgConn.secretKey))
+	// Postgres will process the request and close the connection
+	// so when don't need to read the reply
+	// https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.6.7.10
 	_, err = cancelConn.Write(buf)
-	if err != nil {
-		return err
-	}
-
-	_, err = cancelConn.Read(buf)
-	if err != io.EOF {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 // WaitForNotification waits for a LISTON/NOTIFY message to be received. It returns an error if a notification was not
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
     steps:
 
     - name: Set up Go ${{ matrix.go-version }}
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
         go-version: ${{ matrix.go-version }}
 
@@ -123,3 +123,30 @@ jobs:
         # PGX_TEST_TLS_CONN_STRING: ${{ matrix.pgx-test-tls-conn-string }}
         PGX_SSL_PASSWORD: ${{ matrix.pgx-ssl-password }}
         PGX_TEST_TLS_CLIENT_CONN_STRING: ${{ matrix.pgx-test-tls-client-conn-string }}
+
+  test-windows:
+    name: Test Windows
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        go-version: ["1.19", "1.20"]
+
+    steps:
+      - name: Setup PostgreSQL
+        id: postgres
+        uses: ikalnytskyi/action-setup-postgres@v4
+        with:
+          database: pgx_test
+
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+
+      - name: Test
+        run: go test -v -race -timeout 5m  ./...
+        env:
+          PGX_TEST_DATABASE: ${{ steps.postgres.outputs.connection-uri }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,7 +149,6 @@ jobs:
       - name: Initialize test database
         run: |
           psql -f testsetup/postgresql_setup.sql pgx_test
-
         env:
           PGSERVICE: ${{ steps.postgres.outputs.service-name }}
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,6 +146,14 @@ jobs:
       - name: Check out code into the Go module directory
         uses: actions/checkout@v3
 
+      - name: Initialize test database
+        run: |
+          psql -f testsetup/postgresql_setup.sql pgx_test
+
+        env:
+          PGSERVICE: ${{ steps.postgres.outputs.service-name }}
+        shell: bash
+
       - name: Test
         run: go test -v -race -timeout 5m  ./...
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,6 @@ jobs:
         shell: bash
 
       - name: Test
-        run: go test -v -race -timeout 5m  ./...
+        run: go test -v -race  ./...
         env:
           PGX_TEST_DATABASE: ${{ steps.postgres.outputs.connection-uri }}
--- a/copy_from_test.go
+++ b/copy_from_test.go
@@ -640,7 +640,7 @@ func (fs *slowFailRaceSource) Err() error {
 func TestConnCopyFromSlowFailRace(t *testing.T) {
 	t.Parallel()
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
 	defer cancel()
 
 	conn := mustConnectString(t, os.Getenv("PGX_TEST_DATABASE"))
--- a/internal/nbconn/nbconn.go
+++ b/internal/nbconn/nbconn.go
@@ -96,6 +96,11 @@ type NetConn struct {
 
 	writeDeadlineLock sync.Mutex
 	writeDeadline     time.Time
+
+	// nbOperCnt Tracks how many operations performing simultaneously
+	nbOperCnt int
+	// nbOperMu Used to prevent concurrent SetBlockingMode calls
+	nbOperMu sync.Mutex
 }
 
 func NewNetConn(conn net.Conn, fakeNonBlockingIO bool) *NetConn {
--- a/internal/nbconn/nbconn.go
+++ b/internal/nbconn/nbconn.go
@@ -96,11 +96,12 @@ type NetConn struct {
 
 	writeDeadlineLock sync.Mutex
 	writeDeadline     time.Time
+	// The following fields are used in nbconn_real_non_block_windows
 
-	// nbOperCnt Tracks how many operations performing simultaneously
-	nbOperCnt int
 	// nbOperMu Used to prevent concurrent SetBlockingMode calls
 	nbOperMu sync.Mutex
+	// nbOperCnt Tracks how many operations performing simultaneously
+	nbOperCnt int
 }
 
 func NewNetConn(conn net.Conn, fakeNonBlockingIO bool) *NetConn {
--- a/internal/nbconn/nbconn_test.go
+++ b/internal/nbconn/nbconn_test.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"io"
 	"net"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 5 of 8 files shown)


drakkan:
@jackc `TLS_over_TCP_with_Fake_Non-blocking_IO` is expected to fail on Windows. Do you have ideas for other failures? Thank you

jackc:
Sorry about removing those fields needed for Windows. CI for Windows would be great!

The domain / uint64 tests are because testsetup/postgresql_setup.sql is not being run. It is normally run from ci/setup_test.bash.

I'm not sure about the other ones.

drakkan:
@jackc thank you domain/uilt64 tests are now fixed.
As for the parse config tests I'm unable to reproduce locally

Schermata del 2023-05-28 08-51-30

I think they are affected by some settings in the CI environment. I'll skip them for now.

drakkan:
still no luck, after skipping the parse config tests we now have a panic in TestConnCancelRequest

```
panic: Fail in goroutine after TestConnCancelRequest has completed

goroutine 10231 [running]:
testing.(*common).Fail(0xc0004c81a0)
	C:/hostedtoolcache/windows/go/1.19.9/x64/src/testing/testing.go:824 +0x1a8
testing.(*common).Errorf(0xc0004c81a0, {0xc566fe, 0x3}, {0xc00039c8f0, 0x1, 0x1})
	C:/hostedtoolcache/windows/go/1.19.9/x64/src/testing/testing.go:941 +0xaf
github.com/stretchr/testify/assert.Fail({0xd94b20, 0xc0004c81a0}, {0xc000d3c360, 0x84}, {0x0, 0x0, 0x0})
	C:/Users/runneradmin/go/pkg/mod/github.com/stretchr/testify@v1.8.1/assert/assertions.go:264 +0x438
github.com/stretchr/testify/assert.NoError({0xd94b20, 0xc0004c81a0}, {0xd948e0, 0xc000080780}, {0x0, 0x0, 0x0})
	C:/Users/runneradmin/go/pkg/mod/github.com/stretchr/testify@v1.8.1/assert/assertions.go:1392 +0x127
github.com/stretchr/testify/require.NoError({0xd98ac0, 0xc0004c81a0}, {0xd948e0, 0xc000080780}, {0x0, 0x0, 0x0})
	C:/Users/runneradmin/go/pkg/mod/github.com/stretchr/testify@v1.8.1/require/require.go:1261 +0xa5
github.com/jackc/pgx/v5/pgconn_test.TestConnCancelRequest.func1()
	D:/a/pgx/pgx/pgconn/pgconn_test.go:2056 +0x85
created by github.com/jackc/pgx/v5/pgconn_test.TestConnCancelRequest
	D:/a/pgx/pgx/pgconn/pgconn_test.go:2050 +0x2ea
```

ContextWatcherStress passes but is very slow
```
--- PASS: TestContextWatcherStress (396.25s)
```

drakkan:
a little better after the last change but the tests keep failing

```
=== CONT  TestConnCancelRequest
    pgconn_test.go:2061: 
        	Error Trace:	D:\a\pgx\pgx\pgconn\pgconn_test.go:2061
        	            				D:\a\pgx\pgx\pgconn\asm_amd64.s:1594
        	            	read tcp [::1]:50692->[::1]:5432: wsarecv: An existing connection was forcibly closed by the remote host.
        	Test:       	TestConnCancelRequest
--- FAIL: TestConnCancelRequest (0.79s)
```

I've seen this error randomly in previous runs as well. I don't have more time to better understand the problem for today, sorry

drakkan:
@codercms you may be interested in this

codercms:
@drakkan good news, I can reproduce this error on my localhost

image

codercms:
@jackc @drakkan I think I know the root of problem, an error occurs here - https://github.com/jackc/pgx/blob/master/pgconn/pgconn.go#L904

image

According to Postgres spec here - https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.6.7.10
> To issue a cancel request, the frontend opens a new connection to the server and sends a CancelRequest message, rather than the StartupMessage message that would ordinarily be sent across a new connection. The server will process this request and then close the connection. For security reasons, no direct reply is made to the cancel request message.

It means - when a read attempt is made, it of course fails because Postgres closes the connection.
And WSAECONNRESET error doesn't converted to io.EOF error that is why test actually fails.

codercms:
@drakkan @jackc It looks like a bug in GoLang

from `net.go`:
```go
package net

// Read implements the Conn Read method.
func (c *conn) Read(b []byte) (int, error) {
	if !c.ok() {
		return 0, syscall.EINVAL
	}
	n, err := c.fd.Read(b)
	if err != nil && err != io.EOF {
		err = &OpError{Op: "read", Net: c.fd.net, Source: c.fd.laddr, Addr: c.fd.raddr, Err: err}
	}
	return n, err
}
```

Which invokes func from `fd_posix.go`:
```go
func (fd *netFD) Read(p []byte) (n int, err error) {
	n, err = fd.pfd.Read(p)
	runtime.KeepAlive(fd)
	return n, wrapSyscallError(readSyscallName, err)
}
```

Which invokes func from `fd_windows.go`:
```go
// Read implements io.Reader.
func (fd *FD) Read(buf []byte) (int, error) {
	if err := fd.readLock(); err != nil {
		return 0, err
	}
	defer fd.readUnlock()

	if len(buf) > maxRW {
		buf = buf[:maxRW]
	}

	var n int
	var err error
	if fd.isFile {
		fd.l.Lock()
		defer fd.l.Unlock()
		switch fd.kind {
		case kindConsole:
			n, err = fd.readConsole(buf)
		default:
			n, err = syscall.Read(fd.Sysfd, buf)
			if fd.kind == kindPipe && err == syscall.ERROR_OPERATION_ABORTED {
				// Close uses CancelIoEx to interrupt concurrent I/O for pipes.
				// If the fd is a pipe and the Read was interrupted by CancelIoEx,
				// we assume it is interrupted by Close.
				err = ErrFileClosing
			}
		}
		if err != nil {
			n = 0
		}
	} else {
		o := &fd.rop
		o.InitBuf(buf)
		n, err = execIO(o, func(o *operation) error {
			return syscall.WSARecv(o.fd.Sysfd, &o.buf, 1, &o.qty, &o.flags, &o.o, nil)
		})
		if race.Enabled {
			race.Acquire(unsafe.Pointer(&ioSync))
		}
	}
	if len(buf) != 0 {
		err = fd.eofError(n, err)
	}
	return n, err
}
```

image

And here is `eofError` func call:
image


`error_posix.go`:
```go
// wrapSyscallError takes an error and a syscall name. If the error is
// a syscall.Errno, it wraps it in a os.SyscallError using the syscall name.
func wrapSyscallError(name string, err error) error {
	if _, ok := err.(syscall.Errno); ok {
		err = os.NewSyscallError(name, err)
	}
	return err
}

```

codercms:
@jackc @drakkan perhaps this case should be handled within pgx, like this:
```go
_, err = cancelConn.Read(buf)
if err != io.EOF {
	if errors.Is(err, windows.WSAECONNRESET) {
		return nil
	}
	return err
}
```

With this patch test passed

What do you think?

drakkan:
> @jackc @drakkan perhaps this case should be handled within pgx, like this:
> 
> ```go
> _, err = cancelConn.Read(buf)
> if err != io.EOF {
> 	if errors.Is(err, windows.WSAECONNRESET) {
> 		return nil
> 	}
> 	return err
> }
> ```
> 
> With this patch test passed
> 
> What do you think?

Thank you @codercms. I think we don't need to read the response at all, once the server accepts the cancel request, we're done

drakkan:
Tests finally passed! I think the last things to investigate is why `TestContextWatcherStress` is so slow on Windows.
@codercms can you reproduce on a real Windows host? However, this can also be fixed in a separate PR

codercms:
@drakkan well I'm not sure why this is happens, but here is some data from blocking profiler from modified test case (I've reduced cycle count from 100000 to 100):
image

image

I commented out this line of code:
```go
// Without time.Sleep, cw.Unwatch will almost always run before the cancel func which means cancel will never happen. This gives us a better mix.
if i%3 == 0 {
	//time.Sleep(time.Nanosecond)
}
```

And after that test with 100000 cycles passed in 0.1s.

drakkan:
@codercms yes this is related to the high resolution timer issue on windows. I'm executing the test with this change

```
diff --git a/pgconn/internal/ctxwatch/context_watcher_test.go b/pgconn/internal/ctxwatch/context_watcher_test.go
index 583199c7..6b0f50f9 100644
--- a/pgconn/internal/ctxwatch/context_watcher_test.go
+++ b/pgconn/internal/ctxwatch/context_watcher_test.go
@@ -105,7 +105,7 @@ func TestContextWatcherStress(t *testing.T) {
 
                // Without time.Sleep, cw.Unwatch will almost always run before the cancel func which means cancel will never happen. This gives us a better mix.
                if i%3 == 0 {
-                       time.Sleep(time.Nanosecond)
+                       time.Sleep(15 * time.Millisecond)
                }
 
                cw.Unwatch()
```

and it is taking a long time also on Linux. This is a know Windows [issue](https://github.com/golang/go/issues/44343)

codercms:
@drakkan and also sleep in Windows isn't accurate, here is example:
```go
start := time.Now()
time.Sleep(time.Nanosecond)
fmt.Printf("Slept for %s\n", time.Since(start))
os.Exit(1)
```

It outputs on my system: `Slept for 6.756ms`

drakkan:
so this does not seem a real issue, maybe we should only change the test case or just remove the sleep on Windows until the know issue is fixed. Thank you to all

codercms:
> @codercms yes this is related to the high resolution timer issue on windows. I'm executing the test with this change
> 
> ```
> diff --git a/pgconn/internal/ctxwatch/context_watcher_test.go b/pgconn/internal/ctxwatch/context_watcher_test.go
> index 583199c7..6b0f50f9 100644
> --- a/pgconn/internal/ctxwatch/context_watcher_test.go
> +++ b/pgconn/internal/ctxwatch/context_watcher_test.go
> @@ -105,7 +105,7 @@ func TestContextWatcherStress(t *testing.T) {
>  
>                 // Without time.Sleep, cw.Unwatch will almost always run before the cancel func which means cancel will never happen. This gives us a better mix.
>                 if i%3 == 0 {
> -                       time.Sleep(time.Nanosecond)
> +                       time.Sleep(15 * time.Millisecond)
>                 }
>  
>                 cw.Unwatch()
> ```
> 
> and it is taking a long time also on Linux.
> 
> ```
> go test -v -run ^TestContextWatcherStress$ github.com/jackc/pgx/v5/pgconn/internal/ctxwatch
> === RUN   TestContextWatcherStress
> --- PASS: TestContextWatcherStress (512.41s)
> PASS
> ok  	github.com/jackc/pgx/v5/pgconn/internal/ctxwatch	512.418s
> ```
> 
> This is a know Windows [issue](https://github.com/golang/go/issues/44343)

@drakkan It's actually very hard to make Windows to sleep for microseconds, I think we could just disable sleep for Windows, or just reduce sleeps count like this:
```
// Without time.Sleep, cw.Unwatch will almost always run before the cancel func which means cancel will never happen. This gives us a better mix.
if i%333 == 0 {
	time.Sleep(time.Nanosecond)
}
```

codercms:
@drakkan thank you for your contribution, It's really great to have Windows CI for pgx

drakkan:
> @drakkan thank you for your contribution, It's really great to have Windows CI for pgx

@codercms  just to be sure can you please confirm that the following tests pass on your Windows host?

- TestParseConfig
- TestParseConfigReadsPgServiceFile
- TestParseConfigReadsPgPassfile

they work for me but in my Windows VM I haven't Postgres installed, I connect to the Postgres instance running on my Linux host. I think they fail for some [CI setting](https://github.com/ikalnytskyi/action-setup-postgres/blob/master/action.yml#L45)

codercms:
@drakkan well, this methods actually doesn't make any connection to Postgres, they're just testing that the connection could be parsed into config struct.

I commented out `skipOnWindows(t)` from these tests and then all tests passed.

<details>

image

image

image
</details>

drakkan:
> @drakkan well, this methods actually doesn't make any connection to Postgres, they're just testing that the connection string could be parsed into config struct.
> 
> I commented out `skipOnWindows(t)` from these tests and then all tests passed.

Thanks for confirming. Since I don't have Postgres installed locally, differently from the CI env, I haven't a pg service file or other configuration files that can influence this tests.

pgconn/pgconn.go
```diff
@@ -896,17 +896,11 @@ func (pgConn *PgConn) CancelRequest(ctx context.Context) error {
...
-		return err
-	}
-
-	_, err = cancelConn.Read(buf)
-	if err != io.EOF {
-		return err
-	}
-
-	return nil
+	return err
```
jackc, line 903: This will change behavior somewhat. Read until `EOF` means that `CancelRequest` will only return after the server has definitely received the cancel message or an error occurs. With this change it will return without waiting for the acknowledgement of the server closing the connection.

But I'm not sure if the change is meaningful as the server side cancel request is asynchronous anyway. 🤔

drakkan: > This will change behavior somewhat. Read until `EOF` means that `CancelRequest` will only return after the server has definitely received the cancel message or an error occurs. With this change it will return without waiting for the acknowledgement of the server closing the connection.
> 
> But I'm not sure if the change is meaningful as the server side cancel request is asynchronous anyway. thinking

once write returns we have sent the request to Postgres. It should be safe to stop here. 
If you prefer I catch the `windows.WSAECONNRESET` as @codercms suggested above

jackc:
I decided to go with the `CancelRequest` changes as is.

Thanks for all your work on this!

drakkan:
> I decided to go with the `CancelRequest` changes as is.
> 
> Thanks for all your work on this!

Thank you!

Issue #1618, opened 2023-05-22
Timeout if connection is lost (OpenVPN)


JamesInform:
I am connected to a PostgreSQL server via OpenVPN.
I have an open connection that sends a select statement every 10 seconds in a loop. 

If the OpenVPN connection closes due to some network issue, the next select statement takes 3min+ until a timeout appears.

Question:
Is there any elegant way of handling such a scenario and get an appropriate quick timeout response?
I have tested pgconn.checkConn(), but checkConn() tells, that everything is fine altough it isn't because OpenVPN connection has been gone.

jackc:
If `*PgConn.CheckConn` is not detecting the dropped connection then I would guess that the connection isn't lost as far as Go or the OS is concerned. My guess is OpenVPN is keeping the local side of the connection open while it tries to reconnect transparently (or something like that).

I don't know of any way to detect a situation like that reliably, but what you could do is execute your queries with a context with a short timeout. That would at least keep it from blocking (and setting a context timeout is good practice on any operation that can hang anyway).

JamesInform:
Indeed OpenVPN seems to do nasty tricks on the connection. Even if I kill my go application during the OpenVPN connection is not available, the connection keeps established on the PostgreSQL server. side Very strange.

A context with a timeout is a good idea, but unfortunately I cannot use it, because the executed sql statement 's duration can vary from some seconds to some minutes. So a timeout context could also break a correctly running statement.

Nevertheless zhanks for your quick response and greate support..

ns-gzhang:
@jackc Hi Jack, my scenario is related to `pgconn.CheckConn()` in our environments, not OpenVPN, but PG container on my Mac, or on K8s stacks. The symptom is that `pgconn.Exec()` returns error `conn closed`, but `pgconn.CheckConn()` says no error, it repeated over 2 dozens of times, and finally `pgconn.CheckConn()` detects it. Seems if I use `pgconn.IsClosed()` after `pgconn.Exec()`, it detects the status right away. Is it expected? Or I didn't use it the right way?

jackc:
@ns-gzhang 

[...]

Issue #1605, opened 2023-05-15
Problem with nbconn.NetConn


mhkarimi1383:
I know that this problem is not fully related to this project, but I need help with my problem

```go
func GetRawConnection() (net.Conn, error) {
	conn, err := writePools[rand.Intn(len(writePools))].Acquire(context.Background())
	if err != nil {
		return nil, err
	}
	return conn.Conn().PgConn().Conn(), nil
}
```

and use it as a normal `TCPConn` but I'm getting following error

```
panic: interface conversion: net.Conn is *nbconn.NetConn, not *net.TCPConn
```

Also I cannot use `nbconn.NetConn` since it's in internal module also conn property is not public

here is my code that is code that uses `GetRawConnection`

```go
func handleClient(cliConn *net.TCPConn) {
	defer cliConn.Close()

	serverConn, err := connection.GetRawConnection()
	if err != nil {
		log.Printf("Failed to connect to server: %s", err)
		return
	}
	defer serverConn.Close()
	fmt.Printf("Type of cliConn: %T\n", cliConn)
	fmt.Printf("Type of serverConn: %T\n", serverConn)
	// Proxy(serverConn.(*net.TCPConn), cliConn)
	Proxy(serverConn.(*net.TCPConn), cliConn)
}

func Proxy(srvConn, cliConn *net.TCPConn) {
	serverClosed := make(chan struct{}, 1)
	clientClosed := make(chan struct{}, 1)
	// backend := pgproto3.NewBackend(cliConn, cliConn)

	// startupMsg, err := backend.ReceiveStartupMessage()
	// if err != nil {
	// 	log.Fatal(err)
	// }
	// backend.SetAuthType(pgproto3.AuthTypeMD5Password)

	// switch startupMsg.(type) {
	// case *pgproto3.StartupMessage:
	//	_, err := cliConn.Write((&pgproto3.AuthenticationMD5Password{
	//		Salt: types.MD5AuthSalt,
	//	}).Encode(nil))
	//	if err != nil {
	//		log.Fatal(err)
	//	}
	//}

	go broker(srvConn, cliConn, clientClosed)
	go broker(cliConn, srvConn, serverClosed)

	var waitFor chan struct{}
	select {
	case <-clientClosed:
		srvConn.SetLinger(0)
		srvConn.CloseRead()
		waitFor = serverClosed
	case <-serverClosed:
		cliConn.CloseRead()
		waitFor = clientClosed
	}

	<-waitFor
}

[...]

Issue #1600, opened 2023-05-13
Connection over SSH hung with DialFunc in pgx v5 (v4 works)


d-bytebase:
**Describe the bug**
I'm writing postgres connection over SSH with pgx v5, see code below.

1) The first "SELECT 1" returns, but "SELECT 2" hangs.
2) When I use conn from  db.Conn() instead of db to make the queries, everything works as expected.
3) If I switch to v4, everything works as expected.

**To Reproduce**
Sample code: https://go.dev/play/p/vdy7W739QAl

**Expected behavior**
See above

**Actual behavior**
See above

**Version**
 - Go: `$ go version` -> [go version go1.20.4 darwin/arm64]
 - PostgreSQL: `$ psql --no-psqlrc --tuples-only -c 'select version()'` -> [PostgreSQL 14.7 (Ubuntu 14.7-0ubuntu0.22.04.1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0, 64-bit]
 - pgx: `$ grep 'github.com/jackc/pgx/v[0-9]' go.mod` -> [v5.3.1]

**Additional context**

d-bytebase:
We have switched to pgx/v4 to work around the problem at Bytebase. However, we are hoping v5 to work in the future.

jackc:
Unfortunately, this is a rather messy issue. pgx v5 relies on non-blocking I/O. Go only exposes non-blocking I/O via syscalls. This is implemented on for unix and windows platforms, but a fallback mode that uses `SetWriteDeadline` is also provided. 

Obviously, we can't use OS level syscalls on the conn created by the SSH library so the fallback mode is enabled. But the SSH conn also doesn't actually implement the deadline functions so the fallback mode doesn't work either.

Presumably, the hang is occurring because it's reaching a part of the code that actually needs the deadline to work -- and the stubbed version in your example can't actually abort the read or write.

I'm not really sure what a solution would be here.

d-bytebase:
Thanks! Is there a way to disable the non-blocking IO in v5?

jackc:
Unfortunately, not in this case. Disabling non-blocking I/O was intended to be done with the `SetWriteDeadline` fallback. But that fallback doesn't work here either.

brandur:
[...]

Issue #1481, opened 2023-01-27
CopyFrom is ~80x slower in v5.2.0 versus v4.17.2


jeremy-dupre:
**Describe the bug**
CopyFrom in v5.2.0 is ~80x slower than CopyFrom in v4.17.2
Likely because of the new logic implemented in: func (pgConn *PgConn) CopyFrom(ctx context.Context, r io.Reader, sql string) (CommandTag, error) 

Here is the bench result:

goos: windows
goarch: amd64
pkg: nph/bench
cpu: Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
BenchmarkPgx4-6   	      10	 106087930 ns/op	  363762 B/op	      66 allocs/op
BenchmarkPgx5-6   	       1	8003570000 ns/op	 2422688 B/op	  100559 allocs/op
PASS
ok  	nph/bench	11.587s

**To Reproduce**
Run the bench provided here below

```go
package bench

import (
	"context"
	"testing"

	pgx4 "github.com/jackc/pgx/v4"
	pgx5 "github.com/jackc/pgx/v5"
)

var connString = "postgres://xxx"

func BenchmarkPgx4(b *testing.B) {

	ctx := context.Background()
	conn, err := pgx4.Connect(ctx, connString)
	if err != nil {
		panic(err)
	}
	_, err = conn.Exec(ctx, "CREATE TABLE IF NOT EXISTS test_bench(int_col BIGINT, text_col TEXT)")
	if err != nil {
		panic(err)
	}
	_, err = conn.Exec(ctx, "TRUNCATE TABLE test_bench")
	if err != nil {
		panic(err)
	}
	rows := make([][]any, 100000)
	for i := range rows {
		rows[i] = []any{123, "456789"}
	}
	b.ResetTimer()
	var count int64
	for n := 0; n < b.N; n++ {
		count, err = conn.CopyFrom(ctx, pgx4.Identifier{"test_bench"}, []string{"int_col", "text_col"}, pgx4.CopyFromRows(rows))
		if err != nil {
			panic(err)
		}
	}
	if int(count) != len(rows) {
		b.Fatal("bad result")
	}
}

func BenchmarkPgx5(b *testing.B) {

	ctx := context.Background()
	conn, err := pgx5.Connect(ctx, connString)
	if err != nil {
		panic(err)
	}
	_, err = conn.Exec(ctx, "CREATE TABLE IF NOT EXISTS test_bench(int_col BIGINT, text_col TEXT)")
[...]

PR #1629, opened 2023-06-04
Replace raw syscalls for non-blocking reads

jackc:
The non-blocking IO system was designed to solve three problems:

1. Deadlock that can occur when both sides of a connection are blocked writing because all buffers between are full.
2. The inability to use a write deadline with a TLS.Conn without killing the connection.
3. Efficiently check if a connection is broken before writing. Due to fun characteristics of TCP it is possible a connection broken but writes still to succeed. This can be detected by doing a read. This reduces the cases where the application doesn't know if a query that does a INSERT/UPDATE/DELETE was actually sent to the server or not.

However, the nbconn package is extraordinarily complex, has been a source of very tricky bugs, and has OS specific code paths. It also does not work at all with underlying `net.Conn` implementations that do not have platform specific non-blocking IO syscall support and do not properly implement deadlines. In particular, this is the case with golang.org/x/crypto/ssh. See #1600, #1605, and #1618. In addition, there appears to be a Windows platform bug that is impossible to work around in pgx. See https://github.com/golang/go/issues/58764.

The deadlock problem is now solved with a combination of a goroutine for `CopyFrom` like v4 used and a watchdog timer for other writes that starts a background reader. Moving back to the v4 style `CopyFrom` implementation should also resolve any potential remaining Windows performance issues. See #1481.

The write deadline problem is actually moot. We check for context cancellation before sending a query and the actual Write should be almost instant as long as the underlying connection is not blocked.  (We should only have to wait until it is accepted by the OS, not until it is fully sent.) So the write deadline is almost never encountered.

Efficiently checking if a connection has been closed is probably the hardest to solve without non-blocking reads. However, the existing code only solves part of the problem. It can detect a closed or broken connection the OS knows about, but it won't actually detect other types of broken connections such as a network interruption. This is currently implemented in `CheckConn` and called automatically when checking a connection out of the pool that has been idle for over one second. This is resolved by changing `CheckConn` to a very short deadline read, deprecating it, and replacing internal calls to `CheckConn` with `Ping`.


jackc:
This PR is an overall reduction of over 1,000 lines. The new system is simpler, but still complex in places, and it is a substantial change. I would appreciate some review and testing.

drakkan:
@jackc thank you!

I did a quick test

```
cat go.mod | grep pgx
	github.com/jackc/pgx/v5 v5.3.2-0.20230604013921-d9633608d0fe
```

sftpgo test cases pass on both Linux and Windows. I just use the database/sql driver so nothing complex. 
Honestly this no raw syscalls approach makes me feel much more relaxed, this way pgx should work the same on any platform :smile: 

cc @codercms

pgconn/pgconn.go
```diff
@@ -1633,34 +1675,66 @@ func (pgConn *PgConn) EscapeString(s string) (string, error) {
...
+
 // makeCommandTag makes a CommandTag. It does not retain a reference to buf or buf's underlying memory.
 func (pgConn *PgConn) makeCommandTag(buf []byte) CommandTag {
 	return CommandTag{s: string(buf)}
 }
 
+// enterPotentialWriteReadDeadlock must be called before a write that could deadlock if the server is simultaneously
+// blocked writing to us.
+func (pgConn *PgConn) enterPotentialWriteReadDeadlock() {
+	pgConn.slowWriteTimer.Reset(10 * time.Millisecond)
```
drakkan, line 1715: 10 milliseconds seems like a pretty arbitrary value. Since in most cases this timer shouldn't be needed I was wondering if we could use a higher value, e.g. 100 milliseconds. Does using a higher value create performance issues with batch queries? In any case I would use at least 15 milliseconds as this is the current minimum resolution of timers on Windows so that pgx behaves the same on all supported operating systems

codercms:
@jackc It looks good for me, I'll test it on real Windows machine

pgconn/pgconn.go
jackc: 10 milliseconds is basically an educated guess. A Write should only take as long as the syscall and memcpy to the OS outbound network buffer unless the buffer is full (which potentially is a block). It seemed like long enough for the normal case, but short enough not to kill performance if a block occurs.

And yes, large batch queries are the primary place this would be an issue.

I don't mind increasing it to 15ms, but I wonder Windows would still behave differently. Is that resolution per timer or is it some sort of system wide tick? i.e. Could we end up waiting almost 30ms in the worst case?

drakkan: > 10 milliseconds is basically an educated guess. A Write should only take as long as the syscall and memcpy to the OS outbound network buffer unless the buffer is full (which potentially is a block). It seemed like long enough for the normal case, but short enough not to kill performance if a block occurs.
> 
> And yes, large batch queries are the primary place this would be an issue.
> 
> I don't mind increasing it to 15ms, but I wonder Windows would still behave differently. Is that resolution per timer or is it some sort of system wide tick? i.e. Could we end up waiting almost 30ms in the worst case?

I think this is the resolution per timer. Windows supports high resolution timers but the Go implementation don't use them. There is a WIP patch [here](https://go-review.googlesource.com/c/go/+/488675)

codercms: @jackc Default timer resolution in Windows is 15.6 ms, so in theory the worst case would be awaiting for 15.6 ms.  
Ref: https://randomascii.wordpress.com/2013/07/08/windows-timer-resolution-megawatts-wasted/

**But** it's possible to decrease timer resolution to 1ms or even to 0.5.  
Ref: https://stackoverflow.com/a/22862989/2585154 , but this settings are system-wide, it will change timer resolution for entire system and affect all processes.

P.S. Some process may set timer resolution e.g. to 50ms, then the worst case would be awaiting for 50ms.

codercms: @jackc here is more detailed information about system timers on Windows - http://windowstimestamp.com/description

brandur:
I put the branch into our staging environment and tested it over a number of SSH connections. So far I haven't noticed any problems, which is good, because previously we were able to observe locking issues ~immediately.

So I'm hopeful it's working as expected! I'll follow up if we notice anything amiss. Thanks for the quick work!

pgconn/pgconn.go
jackc: Just pushed a change to 15ms and added docs explaining how 15ms was chosen.

danmcgee-soda:
This change looks great!

We are using a custom DialFunc to connect to Google Cloud SQL databases using the dialer from the `cloudsqlconn` library, and we were seeing a 100ms delay on the first query issued, each time an idle connection was pulled from the pool. This is because the custom dialer forced the code down the fake non-blocking path, instead of having direct access to the underlying net.Conn object. We had to work around it with this bit of code as an afterConnect function, to shorten the hardcoded 100ms timeout value that could only be reduced if you triggered certain code paths:

```golang
func afterConnect(ctx context.Context, c *pgx.Conn) error {
	/*
		Because of our use of the Google Cloud SQL proxy library to talk to the database in cloud environments,
		the switch to non-blocking I/O in pgx/v5 started causing us a 100ms delay on acquiring a connection
		from the connection pool. This doesn't happen locally because the connection is more "direct", and pgx
		can use real non-blocking I/O. However, with the proxy tunnel involved, it falls back to a fake non-blocking
		implementation, which by default waits 100ms to prove the connection hasn't been closed. See `CheckConn`
		and chase it all the way down to `nbconn.go`.

		We can work around this by doing the only operation that reduces this 100ms delay - a COPY FROM. The current
		fake non-blocking implementation will time and measure a read operation in this code path, and reduce the
		delay to 2x the measured delay. In a cloud environment, this should be 1-5 ms at most. While it would be
		great to get this back to 0ms, it is orders of magnitude better than 100ms.

		This one line of code took us about two days to figure out, and many other ideas were tried and ruled out
		along the way. Please don't change this unless you really know what you're doing.
	*/
	_, err := c.CopyFrom(ctx, pgx.Identifier{"schema_migrations"}, []string{"version"}, pgx.CopyFromRows([][]any{}))
	return err
}
```

I'm going to try and find some time to try a build with the changes from this branch, and see if it lets us remove our workaround. I'll report back with the results.

danmcgee-soda:
> I'm going to try and find some time to try a build with the changes from this branch, and see if it lets us remove our workaround. I'll report back with the results.

I was able to run this branch in a Google Cloud K8S deployment today with our workaround removed, connecting via a custom dialer, and it worked great- no slowdown on first usage of the connection from the pool. Let me know if there is any additional testing we can do!

codercms:
@jackc @drakkan I ran tests of PR on a real Windows machine, pgconn_test fails on my machine with these errors (I use PostgreSQL 15 for tests):
```
=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult
--- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult (0.24s)
=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_statement
    conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
    --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_statement (0.03s)

=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_describe
    conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
    --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_describe (0.02s)

=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/describe_exec
    conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
    --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/describe_exec (0.14s)

=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/exec
    conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
    --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/exec (0.02s)

=== RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/simple_protocol
    conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
    --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/simple_protocol (0.02s)
```

But it doesn't seem to be affected by this pull request, so I guess it doesn't matter.
All other tests passed successfully:
```
go test ./...

...

FAIL    github.com/jackc/pgx/v5 10.766s
ok      github.com/jackc/pgx/v5/internal/iobufpool      0.226s
ok      github.com/jackc/pgx/v5/internal/pgio   0.042s
ok      github.com/jackc/pgx/v5/internal/pgmock 0.070s
ok      github.com/jackc/pgx/v5/internal/sanitize       0.041s
ok      github.com/jackc/pgx/v5/pgconn  9.859s
ok      github.com/jackc/pgx/v5/pgconn/internal/bgreader        1.585s
ok      github.com/jackc/pgx/v5/pgconn/internal/ctxwatch        4.008s
ok      github.com/jackc/pgx/v5/pgproto3        0.471s
ok      github.com/jackc/pgx/v5/pgtype  10.166s
ok      github.com/jackc/pgx/v5/pgtype/zeronull 0.713s
ok      github.com/jackc/pgx/v5/pgxpool 7.709s
ok      github.com/jackc/pgx/v5/stdlib  10.982s
ok      github.com/jackc/pgx/v5/tracelog        0.682s
FAIL
```

drakkan:
> @jackc @drakkan I ran tests of PR on a real Windows machine, pgconn_test fails on my machine with these errors (I use PostgreSQL 15 for tests):
> 
> ```
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult
> --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult (0.24s)
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_statement
>     conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
>     --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_statement (0.03s)
> 
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_describe
>     conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
>     --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/cache_describe (0.02s)
> 
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/describe_exec
>     conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
>     --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/describe_exec (0.14s)
> 
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/exec
>     conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
>     --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/exec (0.02s)
> 
> === RUN   TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/simple_protocol
>     conn_test.go:870: ERROR: type "uint64" does not exist (SQLSTATE 42704)
>     --- FAIL: TestUnregisteredTypeUsableAsStringArgumentAndBaseResult/simple_protocol (0.02s)
> ```
> 
> But it doesn't seem to be affected by this pull request, so I guess it doesn't matter. All other tests passed successfully:
> 

@codercms you need to create to uint64 domain like [this](https://github.com/jackc/pgx/blob/master/.github/workflows/ci.yml#L151). After executing the initialization sql all tests should pass

> ```
> go test ./...
> 
> ...
> 
> FAIL    github.com/jackc/pgx/v5 10.766s
> ok      github.com/jackc/pgx/v5/internal/iobufpool      0.226s
> ok      github.com/jackc/pgx/v5/internal/pgio   0.042s
> ok      github.com/jackc/pgx/v5/internal/pgmock 0.070s
> ok      github.com/jackc/pgx/v5/internal/sanitize       0.041s
> ok      github.com/jackc/pgx/v5/pgconn  9.859s
> ok      github.com/jackc/pgx/v5/pgconn/internal/bgreader        1.585s
> ok      github.com/jackc/pgx/v5/pgconn/internal/ctxwatch        4.008s
> ok      github.com/jackc/pgx/v5/pgproto3        0.471s
> ok      github.com/jackc/pgx/v5/pgtype  10.166s
> ok      github.com/jackc/pgx/v5/pgtype/zeronull 0.713s
> ok      github.com/jackc/pgx/v5/pgxpool 7.709s
> ok      github.com/jackc/pgx/v5/stdlib  10.982s
> ok      github.com/jackc/pgx/v5/tracelog        0.682s
> FAIL
> ```

codercms:
@drakkan thx, all tests passed:
```
go test ./...
?       github.com/jackc/pgx/v5/examples/chat   [no test files]
?       github.com/jackc/pgx/v5/examples/todo   [no test files]
?       github.com/jackc/pgx/v5/examples/url_shortener  [no test files]
?       github.com/jackc/pgx/v5/internal/anynil [no test files]
?       github.com/jackc/pgx/v5/internal/stmtcache      [no test files]
?       github.com/jackc/pgx/v5/pgproto3/example/pgfortune      [no test files]
?       github.com/jackc/pgx/v5/log/testingadapter      [no test files]
?       github.com/jackc/pgx/v5/pgxtest [no test files]
ok      github.com/jackc/pgx/v5 10.184s
ok      github.com/jackc/pgx/v5/internal/iobufpool      0.210s
ok      github.com/jackc/pgx/v5/internal/pgio   0.043s
ok      github.com/jackc/pgx/v5/internal/pgmock 0.067s
ok      github.com/jackc/pgx/v5/internal/sanitize       0.042s
ok      github.com/jackc/pgx/v5/pgconn  10.277s
ok      github.com/jackc/pgx/v5/pgconn/internal/bgreader        1.473s
ok      github.com/jackc/pgx/v5/pgconn/internal/ctxwatch        4.332s
ok      github.com/jackc/pgx/v5/pgproto3        0.610s
ok      github.com/jackc/pgx/v5/pgtype  10.857s
ok      github.com/jackc/pgx/v5/pgtype/zeronull 0.605s
ok      github.com/jackc/pgx/v5/pgxpool 7.655s
ok      github.com/jackc/pgx/v5/stdlib  10.346s
ok      github.com/jackc/pgx/v5/tracelog        0.519s
```

codercms:
@jackc @drakkan I also installed the pgx version from that PR into my real app. It seems to work fine, and sometimes I see even better query times.

jackc:
All the feedback has been positive so I'm going to merge this. Thanks for testing it everyone!

pgconn/pgconn.go
```diff
@@ -1713,6 +1793,8 @@ func Construct(hc *HijackedConn) (*PgConn, error) {
 	}
 
 	pgConn.contextWatcher = newContextWatcher(pgConn.conn)
+	pgConn.bgReader = bgreader.New(pgConn.conn)
+	pgConn.slowWriteTimer = time.AfterFunc(time.Duration(math.MaxInt64), pgConn.bgReader.Start)
```
ujihisa, line 1797: Should this also need something like this?

```
pgConn.frontend = config.BuildFrontend(pgConn.bgReader, pgConn.conn)
```

jackc: Yes, you are correct. Fixed in cd46cdd450a2c447d50bf30d6a5be513a0631fdf.

ujihisa: :+1:

hothanhloc68:
> @jackc thank you!
> 
> I did a quick test
> 
> ```
> cat go.mod | grep pgx
> 	github.com/jackc/pgx/v5 v5.3.2-0.20230604013921-d9633608d0fe
> ```
> 
> sftpgo test cases pass on both Linux and Windows. I just use the database/sql driver so nothing complex. 
> Honestly this no raw syscalls approach makes me feel much more relaxed, this way pgx should work the same on any platform :smile: 
> 
> cc @codercms

PR #1631, opened 2023-06-06
Add singleton pgtype.Map for default type mappings

lzakharov:
This PR decreases the memory consumption occurred during type mapping initialization. In the case of using pgx as a driver, a new map is created with default mappings for each database connection, resulting in a bunch of identical allocations that require a lot of memory.

### DIFF
--- a/pgtype/pgtype.go
+++ b/pgtype/pgtype.go
@@ -211,7 +211,9 @@ type Map struct {
 }
 
 func NewMap() *Map {
-	m := &Map{
+	defaultMapInitOnce.Do(initDefaultMap)
+
+	return &Map{
 		oidToType:         make(map[uint32]*Type),
 		nameToType:        make(map[string]*Type),
 		reflectTypeToName: make(map[reflect.Type]string),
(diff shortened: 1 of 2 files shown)


jackc:
At first glance this change appears safe, but I am reluctant to have anything shared between connections. Two areas in particular I wonder about are the scan plan caches and Codecs being concurrency safe. I think the cache is still safe. And AFAIK all the Codecs happen to be safe for concurrent usage, but they never were required to be safe before.

How much improvement does this change make?

lzakharov:
If my calculations are correct, each map requires almost 100kb of memory. Then for 100 connections it takes almost 10mb. I observed an increase in allocations and an increase in GC time, when, under a high load, the proxy in front of the database began to break connections from time to time.

```go
func BenchmarkMap100(b *testing.B) {
	m := make([]*Map, 100)

	for i := 0; i < b.N; i++ {
		for j := 0; j < 100; j++ {
			m[j] = NewMap()
		}
	}
}
```

Current version:

```
BenchmarkMap100-10    	     135	   8917356 ns/op	12734258 B/op	   84265 allocs/op
```

With default map:

```
BenchmarkMap100-10    	   55068	     20623 ns/op	   52800 B/op	     900 allocs/op
```

jackc:
That's much bigger than I would have expected, but it seems to be correct.

It appears that the majority of the the savings are from `reflectTypeToType` / `registerDefaultPgTypeVariants` / `RegisterDefaultPgType`.

Would it be reasonable to only share that `reflectTypeToType` map? Both its keys and values are immutable so I don't worry about sharing data.

But `Type` and `Codec` values are another matter. While I don't see why someone _would_ mutate them, it is not forbidden in the docs and there is nothing to prevent it. If someone did have some reason to do that, and was using `TypeForName` to get the `Codec` and changing something in an `AfterConnect` hook then this would introduce a race condition.

I suppose we _could_ just document that you can't mutate `Type` or `Codec` values after they are registered, but I'm not sure if that is sufficient. Or we could only share `Codec` values that are of size 0 (most non-array values qualify). But if it doesn't make a huge difference I'd rather not share them at all.

lzakharov:
I checked the amount of memory required for types and codecs - they require around 10kb per connection. In general, this does not look like a big problem, even in the case of a fairly large connection pool. On the other hand, I also do not see good examples where their mutation would be necessary (I would even say that it draws on an antipattern). Thus, I tend to share all data and add documentation about `Type` and `Codec` immutability.

PR #1633, opened 2023-06-07
pgtype/hstore: Save 2 allocs in database/sql Scan implementation

evanj:
Remove unneeded string to []byte to string conversion, which saves 2 allocs and should make Hstore text scanning slightly faster.

The Hstore.Scan() function takes a string as input, converts it to []byte, and calls scanPlanTextAnyToHstoreScanner.Scan(). That function converts []byte back to string and calls parseHstore. This refactors scanPlanTextAnyToHstoreScanner.Scan into scanPlanTextAnyToHstoreScanner.scanString so the database/sql Scan function can call it directly, bypassing this conversion.

The added Benchmark shows this saves 2 allocs for longer strings, and saves about 5% CPU overall on my M1 Pro. benchstat output:

```
goos: darwin
goarch: arm64
pkg: github.com/jackc/pgx/v5/pgtype
              │  orig.txt   │              new.txt               │
              │   sec/op    │   sec/op     vs base               │
HstoreScan-10   1.334µ ± 2%   1.257µ ± 2%  -5.77% (p=0.000 n=10)

              │   orig.txt   │               new.txt               │
              │     B/op     │     B/op      vs base               │
HstoreScan-10   2.094Ki ± 0%   1.969Ki ± 0%  -5.97% (p=0.000 n=10)

              │  orig.txt  │              new.txt              │
              │ allocs/op  │ allocs/op   vs base               │
HstoreScan-10   36.00 ± 0%   34.00 ± 0%  -5.56% (p=0.000 n=10)
```

### DIFF
--- a/pgtype/hstore.go
+++ b/pgtype/hstore.go
@@ -43,7 +43,7 @@ func (h *Hstore) Scan(src any) error {
 
 	switch src := src.(type) {
 	case string:
-		return scanPlanTextAnyToHstoreScanner{}.Scan([]byte(src), h)
+		return scanPlanTextAnyToHstoreScanner{}.scanString(src, h)
 	}
 
 	return fmt.Errorf("cannot scan %T", src)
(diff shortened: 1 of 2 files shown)


jackc:
I wasn't able to resolve the merge conflict properly here. So I pulled it down and fixed it locally via rebase. This is merged now even though the PR doesn't reflect it.

evanj:
Great thank you!

PR #1636, opened 2023-06-09
pgtype/hstore_test.go: Extend coverage of scan benchmark

evanj:
I am working on an application that reads a lot of hstore values, and have discovered that scanning it is fairly slow. I'm working on some improvements, but first I wanted a better benchmark. This adds more realistic data, and extends it to cover the three APIs: database/sql, and pgconn.Rows.Scan with both text and binary protocols.

### DIFF
--- a/pgtype/hstore_test.go
+++ b/pgtype/hstore_test.go
@@ -247,7 +247,9 @@ func TestHstoreCodec(t *testing.T) {
 func TestParseInvalidInputs(t *testing.T) {
 	// these inputs should be invalid, but previously were considered correct
 	invalidInputs := []string{
+		// extra comma between values
 		`"a"=>"1", ,b"=>"2"`,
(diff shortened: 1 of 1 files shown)


evanj:
Thanks! Yes, I have noticed that, but unfortunately this application is using database/sql. It is going to be a bit complicated to switch it to using pgx, and also a bit complicated to use `AcquireConn` to use the underlying `pgx.Conn` directly, due to how the queries are being generated. I haven't figured out a way to get `database/sql` + pgx to use the binary format for Hstore. I *think* my optimizations are going to make this "fast enough" for my application for now. We plan on eventually using pgx directly, so we should see the benefits then.

It also turns out that after experimenting with a few optimizations for text parsing, I think I can apply some of the same optimizations to the binary parsing, so hopefully that can get a bit faster also.

Issue #1641, opened 2023-06-13
Pool triggerHealthCheck leaves stranded goroutines for 500ms


bgentry:
**Describe the bug**
After releasing all conns back to a `pgxpool.Pool` and closing the pool, there can still be goroutines left running. That's because `triggerHealthCheck` gets called upon `Release()`, and [it spawns a goroutine](https://github.com/jackc/pgx/blob/c542df4fb4cfaa8126143de5dd8c69f84a9c28cb/pgxpool/pool.go#L383C16-L393) which begins by sleeping 500ms no matter what.

This means there is no way to guarantee that a pgxpool is fully shut down with no remaining remnants, at least not without sleeping a non-trivial amount of time.

**To Reproduce**

I can try to put together a reproduction if it helps. I believe this is triggered by:

1. Running `goleak` in your program
2. Starting up a pgxpool
3. Acquiring a conn
4. Releasing the conn
5. Closing the pool
6. Observing that `goleak` sometimes detects open/leaked goroutines when the program exits.

**Expected behavior**
Closing a `pgxpool.Pool` should leave behind no remnants and no open goroutines.

**Actual behavior**
Because there is a hardcoded 500ms sleep that doesn't terminate when i.e. the pool closes sooner, goroutines are sometimes left running after the pool is closed.

**Version**
 - Go: `$ go version` -> `go version go1.20.4 darwin/amd64`
 - PostgreSQL: `$ psql --no-psqlrc --tuples-only -c 'select version()'` -> ` PostgreSQL 14.8 (Homebrew) on x86_64-apple-darwin22.4.0, compiled by Apple clang version 14.0.3 (clang-1403.0.22.14.1), 64-bit`
 - pgx: `$ grep 'github.com/jackc/pgx/v[0-9]' go.mod` -> `github.com/jackc/pgx/v5 v5.3.1`

**Additional context**
I did not see any past issues discussing this, so I wanted to at least confirm that it was an intentional design choice and not merely an oversight or bug.

jackc:
> Because there is a hardcoded 500ms sleep that doesn't terminate when i.e. the pool closes sooner, goroutines are sometimes left running after the pool is closed.

But the goroutine terminates after that 500ms, right? Does this behavior cause a problem?

jackc:
🤷‍♂️ 

[...]

PR #1642, opened 2023-06-14
Faster shutdown of pgxpool.Pool background goroutines

bgentry:
When a pool is closed, some background goroutines may be left open, particularly for health checks as detailed in #1641. Two specific examples have been refactored here to avoid a blocking sleep and instead also select on the pool being closed to potentially return/continue sooner.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -93,6 +93,9 @@ type Pool struct {
 	maxConnIdleTime       time.Duration
 	healthCheckPeriod     time.Duration
 
+	healthCheckMu    sync.Mutex
+	healthCheckTimer *time.Timer
+
 	healthCheckChan chan struct{}
 
 	closeOnce sync.Once
@@ -381,15 +384,25 @@ func (p *Pool) isExpired(res *puddle.Resource[*connResource]) bool {
 }
 
 func (p *Pool) triggerHealthCheck() {
-	go func() {
+	const healthCheckDelay = 500 * time.Millisecond
+
+	p.healthCheckMu.Lock()
+	defer p.healthCheckMu.Unlock()
+
+	if p.healthCheckTimer == nil {
 		// Destroy is asynchronous so we give it time to actually remove itself from
 		// the pool otherwise we might try to check the pool size too soon
-		time.Sleep(500 * time.Millisecond)
-		select {
-		case p.healthCheckChan <- struct{}{}:
-		default:
-		}
-	}()
+		p.healthCheckTimer = time.AfterFunc(healthCheckDelay, func() {
+			select {
+			case <-p.closeChan:
+			case p.healthCheckChan <- struct{}{}:
+			default:
+			}
+		})
+		return
+	}
+
+	p.healthCheckTimer.Reset(healthCheckDelay)
 }
 
 func (p *Pool) backgroundHealthCheck() {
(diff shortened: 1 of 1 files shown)


pgxpool/pool.go
```diff
@@ -408,6 +417,9 @@ func (p *Pool) backgroundHealthCheck() {
 }
 
 func (p *Pool) checkHealth() {
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
```
bgentry, line 421: I opted for a `time.Ticker` on this one for simplicity, but this does have the side effect of changing this behavior from a "sleep 500ms between attempts" to a "make an attempt every 500ms" type behavior. If the health checks take awhile then there could be minimal sleeping time.

I can refactor to use a timer, it would just be messier so I wanted to propose this option first and see what you think.

bgentry:
@jackc Just pushed an updated version, is this roughly what you had in mind?

Note that I needed to introduce a mutex for synchronization over the new timer field. I considered instead just setting this field up as part of pool init so it never needs to be directly written to, but if I did that I would need to specify an initial duration for the timer. Always triggering a health check 500ms after startup seems like a behavioral change AFAICT so I didn't want to go down that route without discussing.

jackc:
> Note that I needed to introduce a mutex for synchronization over the new timer field. I considered instead just setting this field up as part of pool init so it never needs to be directly written to, but if I did that I would need to specify an initial duration for the timer. Always triggering a health check 500ms after startup seems like a behavioral change AFAICT so I didn't want to go down that route without discussing.

I think it would be fine to initially set the timer to 100 years or something like that. Then `triggerHealthCheck` would reset it.

jameshartig:
I realized a cleaner way to do this is to call `triggerHealthCheck` in the destructor callback sent to puddle. Unfortunately the destructor is called before the actual resource is removed but it's called directly after it so we might be able to relax the sleep time to be something even smaller.

One "hack" would be in the destructor passed to puddle to do:
```go
go func() {
  p.p.Stat()
  select {
  case p.healthCheckChan <- struct{}{}:
  default:
  }
}()
```
The reason this works is that `destroyAcquiredResource` has a lock on the puddle's mux and is not unlocked until after the resource is removed. `Stat` tries to lock the same mux so it would lock until the resource has been removed.

That's obviously relying on something that puddle isn't guaranteed to maintain but maybe it'll inspire some better ideas.

nicois:
I know Puddle is in maintenance mode (ie: avoiding changing the API) but wouldn't it be helpful in general to have something like `onRelease(func ())` available on the `Puddle.Pool`? This is a case where it's not possible to safely schedule an activity for such a situation.

batazor:
any update?

rubiagatra:
Any update for this PR @bgentry?

pgxpool/pool.go
bgentry: I'm a bit fuzzy since this was 2 years ago, but maybe this was an attempt to avoid timer leaks on shutdown? But post Go 1.23 [this is no longer a concern](https://pkg.go.dev/time#After) so I think I can revert this part.

bgentry:
@jackc I rebased this and trimmed out the unnecessary diff. The flaky test failure I got appears to be unrelated—at least I can't see a reason why changes would cause an error to occur when terminating the backend, or cause the connection to terminate prematurely. I wonder if you've seen other failures related to 1516fb81255edebc7023cee586eb8906980f130f since that was merged a month ago.

jackc:
Thanks!

PR #1645, opened 2023-06-15
pgtype/hstore: Make text parsing about 6X faster

evanj:
I am working on an application that uses hstore types, and we found that returning the values is slow, particularly when using the text protocol, such as when using database/sql. This improves parsing to be about 6X faster (currently faster than binary). The changes are:

* referencing the original string instead of copying into new strings (very large win)
* using string.IndexByte to scan double quoted strings: it has architecture-specific assembly implementations, and most of the time is spent in key/value strings.
* estimating the number of key/value pairs to allocate the correct size of the slice and map up front. This reduces the number of allocations and bytes allocated by a factor of 2, and was a small CPU win.
* parsing directly into the Hstore, rather than copying into it.

This parser is stricter than the old one. It only accepts hstore strings serialized by Postgres. The old one was already stricter than Postgres's own parser, but previously accepted any whitespace character after a comma. This one only accepts space. Example:
```
  "k1"=>"v1",\t"k2"=>"v2"
```
Postgres only ever uses ", " as the separator. See hstore_out: https://github.com/postgres/postgres/blob/master/contrib/hstore/hstore_io.c

The result of using benchstat to compare the benchmark on my M1 Pro with the following command line in below. The new text parser is now faster than the binary parser. I will improve the binary parser in a separate change.
```
for i in $(seq 10); do go test ./pgtype -run=none -bench=BenchmarkHstoreScan -benchtime=1s >> new.txt; done

goos: darwin
goarch: arm64
pkg: github.com/jackc/pgx/v5/pgtype
                               │  orig.txt   │               new.txt               │
                               │   sec/op    │   sec/op     vs base                │
HstoreScan/databasesql.Scan-10   82.11µ ± 1%   10.51µ ± 0%  -87.20% (p=0.000 n=10)
HstoreScan/text-10               83.30µ ± 1%   11.49µ ± 1%  -86.20% (p=0.000 n=10)
HstoreScan/binary-10             15.99µ ± 2%   15.77µ ± 1%   -1.35% (p=0.007 n=10)
geomean                          47.82µ        12.40µ       -74.08%

                               │   orig.txt   │               new.txt                │
                               │     B/op     │     B/op      vs base                │
HstoreScan/databasesql.Scan-10   56.23Ki ± 0%   11.68Ki ± 0%  -79.23% (p=0.000 n=10)
HstoreScan/text-10               65.12Ki ± 0%   20.58Ki ± 0%  -68.40% (p=0.000 n=10)
HstoreScan/binary-10             21.09Ki ± 0%   21.09Ki ± 0%        ~ (p=0.378 n=10)
geomean                          42.58Ki        17.18Ki       -59.66%

                               │  orig.txt   │               new.txt                │
                               │  allocs/op  │ allocs/op   vs base                  │
HstoreScan/databasesql.Scan-10   744.00 ± 0%   44.00 ± 0%  -94.09% (p=0.000 n=10)
HstoreScan/text-10               743.00 ± 0%   44.00 ± 0%  -94.08% (p=0.000 n=10)
HstoreScan/binary-10              464.0 ± 0%   464.0 ± 0%        ~ (p=1.000 n=10) ¹
geomean                           635.4        96.49       -84.81%
¹ all samples are equal
```

### DIFF
--- a/pgtype/hstore.go
+++ b/pgtype/hstore.go
@@ -1,13 +1,11 @@
 package pgtype
 
 import (
-	"bytes"
 	"database/sql/driver"
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"unicode"
-	"unicode/utf8"
+	"strings"
 
 	"github.com/jackc/pgx/v5/internal/pgio"
 )
@@ -247,21 +245,11 @@ func (s scanPlanTextAnyToHstoreScanner) Scan(src []byte, dst any) error {
 
 // scanString does not return nil hstore values because string cannot be nil.
 func (scanPlanTextAnyToHstoreScanner) scanString(src string, scanner HstoreScanner) error {
-	keys, values, err := parseHstore(src)
+	hstore, err := parseHstore(src)
 	if err != nil {
 		return err
 	}
-
-	m := make(Hstore, len(keys))
-	for i := range keys {
-		if values[i].Valid {
-			m[keys[i]] = &values[i].String
-		} else {
-			m[keys[i]] = nil
-		}
-	}
-
-	return scanner.ScanHstore(m)
+	return scanner.ScanHstore(hstore)
 }
 
 func (c HstoreCodec) DecodeDatabaseSQLValue(m *Map, oid uint32, format int16, src []byte) (driver.Value, error) {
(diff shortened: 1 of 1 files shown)


evanj:
Feel free to reject this: This is a big change. I would be happy to attempt to make this change in multiple smaller pieces if that makes sense. I also tested this with a really gross fuzz test that round-trips values to postgres. I would be happy to try to contribute that if it would be helpful, but I will need to clean it up.

jackc:
> using string.IndexByte to scan double quoted strings: it has architecture-specific assembly implementations, and most of the time is spent in key/value strings.

and

> parsing directly into the Hstore, rather than copying into it.

seem like pretty obvious wins. Also, sizing the map appropriately upfront.

I'm not so sure about the single `string` and `[]string` to reduce allocations. I've used that technique multiple times in pgx and it yielded impressive results. But I've also reverted it multiple times because on rare occasions people ended up pinning large amounts of memory unexpectedly. And that type issue can be difficult to diagnose.

If it would make sense to separate that out, then the rest should be mergeable. But it may or may not be worth the work without the big win from the reduced allocations.

I guess the same issue would apply to #1646.

---

Also, if you want to make a "fasterhstore" package with the memory optimizations I'd be happy to include a link to it in the README.md.

evanj:
Totally reasonable! Let me remove those parts of the changes.

Good idea about the hstore alternative. I was already considering it, but I think it would be even better if it was `map[string]pgtype.Text` since that should remove the need for the `[]string` since it gets rid of the pointers.

evanj:
Now that I looked at the code again, I am going to undo the single string part, but I don't think this is actually a change for text parsing. The text parsing code *already* allocates a single backing slice. I learned this because my first attempt to parse directly into the map was slower, and I was very confused about why.

In the current master branch:

1. `parseHstore` allocates a slice of `[]Text` by building it with `append()`: https://github.com/jackc/pgx/blob/master/pgtype/hstore.go#L395
2. `scanString` converts each one of these elements to *string by taking a pointer to it, and storing that in the Hstore map: https://github.com/jackc/pgx/blob/master/pgtype/hstore.go#L258

I'm pretty sure this means the backing `[]Text` won't be deallocated until all the `*string` pointers in the Hstore go away, which is the *same* as this current text parser.

Since this is *not* a change, I would argue this part of the text parser is okay.

Note that this *is* a change for binary parsing, but if this part is okay for text parsing, I would argue it should also be okay for binary parsing, (or I should remove it from both).

... change coming shortly.

evanj:
Benchstat output from the commit message, showing that this doesn't slow things down too much:

benchstat difference of original versus this version:
```
                               │  orig.txt   │     new-do-not-share-string.txt     │
                               │   sec/op    │   sec/op     vs base                │
HstoreScan/databasesql.Scan-10   82.11µ ± 1%   14.24µ ± 2%  -82.66% (p=0.000 n=10)
HstoreScan/text-10               83.30µ ± 1%   14.97µ ± 1%  -82.03% (p=0.000 n=10)
HstoreScan/binary-10             15.99µ ± 2%   15.80µ ± 0%   -1.16% (p=0.024 n=10)
geomean                          47.82µ        14.99µ       -68.66%

                               │   orig.txt   │     new-do-not-share-string.txt      │
                               │     B/op     │     B/op      vs base                │
HstoreScan/databasesql.Scan-10   56.23Ki ± 0%   20.11Ki ± 0%  -64.24% (p=0.000 n=10)
HstoreScan/text-10               65.12Ki ± 0%   29.00Ki ± 0%  -55.47% (p=0.000 n=10)
HstoreScan/binary-10             21.09Ki ± 0%   21.09Ki ± 0%        ~ (p=0.722 n=10)
geomean                          42.58Ki        23.08Ki       -45.80%

                               │  orig.txt  │     new-do-not-share-string.txt      │
                               │ allocs/op  │ allocs/op   vs base                  │
HstoreScan/databasesql.Scan-10   744.0 ± 0%   340.0 ± 0%  -54.30% (p=0.000 n=10)
HstoreScan/text-10               743.0 ± 0%   339.0 ± 0%  -54.37% (p=0.000 n=10)
HstoreScan/binary-10             464.0 ± 0%   464.0 ± 0%        ~ (p=1.000 n=10) ¹
geomean                          635.4        376.8       -40.70%
¹ all samples are equal
```
benchstat difference of the shared string versus not:
```
                               │ new-share-string.txt │     new-do-not-share-string.txt     │
                               │        sec/op        │   sec/op     vs base                │
HstoreScan/databasesql.Scan-10            10.57µ ± 2%   14.24µ ± 2%  +34.69% (p=0.000 n=10)
HstoreScan/text-10                        11.60µ ± 2%   14.97µ ± 1%  +29.03% (p=0.000 n=10)
HstoreScan/binary-10                      15.87µ ± 2%   15.80µ ± 0%        ~ (p=0.280 n=10)
geomean                                   12.48µ        14.99µ       +20.07%

                               │ new-share-string.txt │     new-do-not-share-string.txt      │
                               │         B/op         │     B/op      vs base                │
HstoreScan/databasesql.Scan-10           11.68Ki ± 0%   20.11Ki ± 0%  +72.17% (p=0.000 n=10)
HstoreScan/text-10                       20.58Ki ± 0%   29.00Ki ± 0%  +40.93% (p=0.000 n=10)
HstoreScan/binary-10                     21.08Ki ± 0%   21.09Ki ± 0%        ~ (p=0.427 n=10)
geomean                                  17.17Ki        23.08Ki       +34.39%

                               │ new-share-string.txt │      new-do-not-share-string.txt       │
                               │      allocs/op       │  allocs/op   vs base                   │
HstoreScan/databasesql.Scan-10             44.00 ± 0%   340.00 ± 0%  +672.73% (p=0.000 n=10)
HstoreScan/text-10                         44.00 ± 0%   339.00 ± 0%  +670.45% (p=0.000 n=10)
HstoreScan/binary-10                       464.0 ± 0%    464.0 ± 0%         ~ (p=1.000 n=10) ¹
geomean                                    96.49         376.8       +290.47%
```

PR #1648, opened 2023-06-17
pgtype array: Fix encoding of vtab \v

evanj:
Arrays with values that start or end with vtab ("\v") must be quoted. Postgres's array parser skips leading and trailing whitespace with the array_isspace() function, which is slightly different from the scanner_isspace() function that was previously linked. Add a test that reproduces this failure, and fix the definition of isSpace.

This also includes a change to use strings.EqualFold which should really not matter, but does not require copying the string.


jackc:
LGTM

PR #1664, opened 2023-06-29
pgtype.Hstore: add a round-trip test for binary and text codecs

evanj:
This ensures the output of Encode can pass through Scan and produce the same input. This found two two minor problems with the text codec. These are not bugs: These situations do not happen when using pgx with Postgres. However, I think it is worth fixing to ensure the code is internally consistent.

The problems with the text codec are:

* It did not correctly distinguish between nil and empty. This is not a problem with Postgres, since NULL values are marked separately, but the binary codec distinguishes between them, so it seems like the text codec should as well.
* It did not output spaces between keys. Postgres produces output in this format, and the parser now only strictly parses the Postgres format. This is not a bug, but seems like a good idea.

### DIFF
--- a/pgtype/hstore.go
+++ b/pgtype/hstore.go
@@ -121,8 +121,15 @@ func (encodePlanHstoreCodecText) Encode(value any, buf []byte) (newBuf []byte, e
 		return nil, err
 	}
 
-	if hstore == nil {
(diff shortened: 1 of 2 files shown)


jackc:
LGTM

PR #1678, opened 2023-07-07
Use bytes.Equal rather than bytes.Compare ==/!= 0

danmcgee-soda:
This is a no-op change that I noticed when poking around the codebase to see how something else worked.

This change is recommended by go-staticcheck, but also might be a bit more efficient for the compiler to implement, since we don't care about which slice of bytes is greater than the other one.

https://www.sobyte.net/post/2022-01/go-bytes-compare-equal/


jackc:
LGTM

Issue #1673, opened 2023-07-05
Session reset callbacks cannot be used with the background reader


adriansmares:
**Describe the bug**

The background reader introduced in https://github.com/jackc/pgx/pull/1629 maintains the FD lock over the underlying `net.Conn` / file descriptor. As such, because the background reader keeps the file descriptor lock, the session reset callback cannot do a low level file descriptor `Read` in order to check if the connection has been gracefully ended or not.

**To Reproduce**

We use [this](https://github.com/TheThingsNetwork/lorawan-stack/blob/a19daeeea885c4985d91b0695c67fdf068de4010/pkg/util/store/conncheck_unix.go#L26-L65) implementation of a session reset fallback.

Because the background reader is still operating in the background after a connection has been used for a query at least once, the session reset fallback basically deadlocks while waiting for the background read to end (which it never does, since the background reader has no deadline).

```
goroutine 84 [semacquire, 4 minutes]:
internal/poll.runtime_Semacquire(0x4058cf?)
	/opt/hostedtoolcache/go/1.20.5/x64/src/runtime/sema.go:67 +0x27
internal/poll.(*fdMutex).rwlock(0xc0001ba780, 0xd0?)
	/opt/hostedtoolcache/go/1.20.5/x64/src/internal/poll/fd_mutex.go:154 +0xd2
internal/poll.(*FD).readLock(...)
	/opt/hostedtoolcache/go/1.20.5/x64/src/internal/poll/fd_mutex.go:221
internal/poll.(*FD).RawRead(0xc0001ba780, 0xc0018342d0)
	/opt/hostedtoolcache/go/1.20.5/x64/src/internal/poll/fd_unix.go:755 +0x45
net.(*rawConn).Read(0xc001814cc0, 0x2e22328?)
	/opt/hostedtoolcache/go/1.20.5/x64/src/net/rawconn.go:43 +0x45
go.thethings.network/lorawan-stack/v3/pkg/util/store.checkConn(***0x2e22328, 0xc000d50a20***)
	/home/runner/work/lorawan-stack/lorawan-stack/pkg/util/store/conncheck_unix.go:45 +0xba

...

 goroutine 1010 [IO wait, 4 minutes]:
internal/poll.runtime_pollWait(0x7ff1457afe78, 0x72)
	/opt/hostedtoolcache/go/1.20.5/x64/src/runtime/netpoll.go:306 +0x89
[...]

PR #1679, opened 2023-07-08
Add PgConn.SyncConn

jackc:
This provides a way to ensure it is safe to directly read or write to the underlying net.Conn.

https://github.com/jackc/pgx/issues/1673

---

Unfortunately, I'm not sure of a good way to test this as I don't know of a sure way to cause the conditions that `SyncConn` removes.

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -1732,6 +1732,7 @@ func (pgConn *PgConn) exitPotentialWriteReadDeadlock() {
 	// The state of the timer is not relevant upon exiting the potential slow write. It may both
 	// fire (due to a slow write), or not fire (due to a fast write).
 	_ = pgConn.slowWriteTimer.Stop()
+	pgConn.bgReader.Stop()
 }
 
 func (pgConn *PgConn) flushWithPotentialWriteReadDeadlock() error {
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -556,7 +556,8 @@ func (pgConn *PgConn) receiveMessage() (pgproto3.BackendMessage, error) {
 	return msg, nil
 }
 
-// Conn returns the underlying net.Conn. This rarely necessary.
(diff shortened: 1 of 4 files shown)


jackc:
Huh. Not sure that test is failing.

pgconn/pgconn.go
```diff
@@ -1740,6 +1741,30 @@ func (pgConn *PgConn) flushWithPotentialWriteReadDeadlock() error {
...
 }
 
+// SyncConn prepares the underlying net.Conn for direct use. PgConn may internally buffer reads or use goroutines for
+// background IO. This means that any direct use of the underlying net.Conn may be corrupted if a read is already
+// buffered or a read is in progress. SyncConn drains read buffers and stops background IO. In some cases this may
+// require sending a ping to the server. ctx can be used to cancel this operation. This should be called before any
+// operation that will use the underlying net.Conn directly. e.g. Before Conn() or Hijack().
+//
+// This should not be confused with the PostgreSQL protocol Sync message.
+func (pgConn *PgConn) SyncConn(ctx context.Context) error {
```
adriansmares, line 1751: I am not sure if the description is correct here - the background reader continues as long as there was no error, or the state is still running, so without calling `Close()` first, how would the background reader actually close ?

This relates to the test as well - since you never call close on the background reader, `SyncConn` will never finish since there is no signal to the background reader to stop after one read.

jackc: 🤦‍♂️ I missed stopping the background reader. It was never supposed to run continuously it was supposed to stop as soon as possible after `exitPostentialWriteReadDeadlock` is called (i.e. after the next read).

That's what I was relying on with the design of `SyncConn`. If `bgReader` is running then a `Ping` will trigger a read which will terminate the background read. It loops several times in case a LISTEN/NOTIFY or other server initiated message is received.

I think it should work now with c8656db84c492e4400f72aa6cf240d373e4822d1.

PR #1680, opened 2023-07-10
Remove some now unused pgtype code

danmcgee-soda:
Most of this is in conversion, and I assume it became unused with some of the v5 changes and refactors to a codec-based approach.

There are likely a few more cleanups to be made, but these ones seemed easy and safe to start with.


jackc:
LGTM

PR #1688, opened 2023-07-14
go.mod: run go mod tidy; removes golang.org/x/sys

evanj:
I'm not sure exactly what commit removed the usage of this module, but it seems worth simplifying the dependencies.


jackc:
👍

PR #1693, opened 2023-07-20
fix TestPoolBackgroundChecksMinConns and NewConnsCount

### LINKED ISSUE #1690: Windows CI failure with TestPoolBackgroundChecksMinConns
CI consistently fails for Windows on the test TestPoolBackgroundChecksMinConns. I do not have a working Windows development available so I am unable to reproduce locally.

Also, the test is failing in a way that should be impossible. The test runner is aborting after 10 minutes. However, there is a context timeout for 2 minutes in the loop that the stack trace reveals the execution was at.

jameshartig:
Previously it was checking TotalConns but that includes ConstructingConns. Instead it should directly check IdleConns so the next Acquire takes one of those and doesn't make a 3rd connection. The check against the context was also wrong which prevented this from timing out after 2 minutes.

This also fixes a bug where NewConnsCount was not correctly counting connections created by Acquire directly.

Fixes #1690

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -199,6 +199,7 @@ func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
 	p.p, err = puddle.NewPool(
 		&puddle.Config[*connResource]{
 			Constructor: func(ctx context.Context) (*connResource, error) {
+				atomic.AddInt64(&p.newConnsCount, 1)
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

jameshartig:
@jackc could you include this in a release when you have a chance?

jackc:
@jameshartig Released in `v5.4.3`.

PR #1698, opened 2023-07-26
feat: Support injecting tls.Config and x509.CertPool for sql.Open

siketyan:
pgx fully supports TLS connection between Go and PostgreSQL server. However, we cannot inject TLS configurations or CA certificates through the database/sql abstraction, as the abstraction is not aware of TLS connections for drivers.

go-mysql-driver, a MySQL connection driver that also implements database/sql, supports injecting TLS configurations via a global variable [^1]. To support TLS connections through database/sql, we need to add the same implementation as go-mysql-driver does.

This pull request proposes to add injection of tls.Config and x509.CertPool for usages through database/sql. These will be injected on parsing the DSN. After we registered a tls.Config with `custom` key, we can use it by adding `sslmode=custom` in DSN. For use cases that need injection of root CA certificates only, we can also register x509.CertPool. The certificate pool will be injected to TLS configuration in conjuction with `sslmode`.

[^1]: https://pkg.go.dev/github.com/go-sql-driver/mysql#RegisterTLSConfig

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -679,20 +679,26 @@ func configTLS(settings map[string]string, thisHost string, parseConfigOptions P
 	case "verify-full":
 		tlsConfig.ServerName = host
 	default:
-		return nil, errors.New("sslmode is invalid")
+		if tlsConfig = findTLSConfig(sslmode); tlsConfig == nil {
+			return nil, errors.New("sslmode is invalid")
+		}
 	}
 
 	if sslrootcert != "" {
 		caCertPool := x509.NewCertPool()
 
-		caPath := sslrootcert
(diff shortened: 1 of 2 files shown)


jackc:
> pgx fully supports TLS connection between Go and PostgreSQL server. However, we cannot inject TLS configurations or CA certificates through the database/sql abstraction, as the abstraction is not aware of TLS connections for drivers.

Any configuration you can set through pgx directly should be available through database/sql via https://pkg.go.dev/github.com/jackc/pgx/v5@v5.4.2/stdlib#RegisterConnConfig.

siketyan:
Oh, I didn't notice that. Thank you for the suggestion.
Closing this pull request.

Issue #1411, opened 2022-12-04
Proposal: CollectOneRow and QueryRow should return an error if multiple rows are returned


jackc:
**Is your feature request related to a problem? Please describe.**

`CollectOneRow()` and `QueryRow()` return an error if no rows are returned by the query. However, if multiple rows are returned the first row is used and all other rows are ignored. These functions imply that the caller expects exactly one row in response. Multiple rows being returned is (almost?) always a bug.

**Describe the solution you'd like**

A new error (e.g. `pgx.ErrTooManyRows`) should be returned if more than one row is found. The first row would still be scanned or returned.

**Describe alternatives you've considered**

* New functions could be introduced with this behavior. e.g. `CollectExactlyOneRow()` or `CollectOneRowStrict()`. `QueryRow` would be left alone in this case.
* A strict mode flag could be added to connection config that enabled / disabled this behavior. (I think this is probably a bad idea.) 

**Additional context**

This is a behavior change so it's arguable whether this change should be allowed in a minor version change. But I can't think of any legitimate use for the existing behavior. I think it is always a bug and any breakage would actually be revealing that bug.

Also, it's not clear to me whether, how, or if this behavior should cascade to database/sql (stdlib).

mwchambers:
@jackc 

> A new error (e.g. pgx.ErrTooManyRows) should be returned if more than one row is found. The first row would still be scanned or returned.

:thumbsup:

> New functions could be introduced with this behavior. e.g. CollectExactlyOneRow() or CollectOneRowStrict(). QueryRow would be left alone in this case. A strict mode flag could be added to connection config that enabled / disabled this behavior.

:thumbsdown: I vote to fix CollectOneRow (even with behaviour change)

> Multiple rows being returned is (almost?) always a bug.

Certainly true for my uses of CollectOneRow and QueryRow.

> This is a behavior change so it's arguable whether this change should be allowed in a minor version change. But I can't think of any legitimate use for the existing behavior. I think it is always a bug and any breakage would actually be revealing that bug.

You could argue that it's not _always_ a bug. Some people might be relying on the behaviour of getting any row of many. But how many are _unintentionally_ doing this? 

If the intent is to take the first of many we can be explicit and `LIMIT 1`

eandre:
[...]

PR #1720, opened 2023-08-19
Add CollectExactlyOneRow function

MartyHub:
Hello,

this is a PR for #1411

Thanks and regards

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -99,8 +99,12 @@ func (ident Identifier) Sanitize() string {
 	return strings.Join(parts, ".")
 }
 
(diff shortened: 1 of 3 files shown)


jackc:
`TestCollectExactlyOneRowExtraRows` is failing because it is not expecting the first row when there is an error because of multiple rows found.

I think the behavior the test is expecting is actually preferable -- no data result on error.

MartyHub:
OK, sorry, I'll correct my last commit!

jackc:
👍

PR #1749, opened 2023-09-27
allow replace xid datatypes

nurlantulemisov:
PR from https://github.com/jackc/pgtype/issues/209

By default, we get a 32-bit XID with the data `type XID pguint32`. If we replace `pgtype.XIDOID`, we can use a custom data type instead.


jackc:
I don't see exactly how this would work. You can't scan to a `pgtype.DataType`.

PR #1757, opened 2023-10-04
Make use of strings.Cut

scop:


jackc:
Merged with typo correction.

PR #1765, opened 2023-10-10
Fix pgconn cancel query through the PGBouncer

### LINKED ISSUE #1764: pgx doesn't cancel query correctly through pgbouncer
**Describe the bug**
https://github.com/jackc/pgx/pull/1621/commits/12b177f8a34d2c666304f3cc84fd8ca4b460da7a causes queries not being cancelled corretly through the PGBouncer. It's not sufficient to only write in the connection.

**To Reproduce**
Test is updated also to check cancellation of queries through the PGBouncer.

1. Run long query and cancel it
```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
)

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("PGX_TEST_PGBOUNCER_CONN_STRING"))
	if err != nil {
		log.Fatal(err)
	}
        
        ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
	defer cancel()
	
	err = conn.Query(ctx, "select pg_sleep(30)")
}
```
2. Check pg_stat_activity
```
select * from pg_stat_activity where state='active';
```

**Expected behavior**
Queries should be cancelled when using PGbouncer as it works with PostgreSQL directly.

**Actual behavior**
Queries remain active in pg_stat_activity

**Version**
 - Go: go1.21.2 darwin/amd64
 - PGBouncer: 1.20.1
 - PostgreSQL: PostgreSQL 15.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 12.2.1_git20220924-r4) 12.2.1 20220924, 64-bit
 - pgx: v5.4.3

levakin:
Closes #1764

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -937,11 +937,17 @@ func (pgConn *PgConn) CancelRequest(ctx context.Context) error {
 	binary.BigEndian.PutUint32(buf[4:8], 80877102)
 	binary.BigEndian.PutUint32(buf[8:12], uint32(pgConn.pid))
 	binary.BigEndian.PutUint32(buf[12:16], uint32(pgConn.secretKey))
-	// Postgres will process the request and close the connection
-	// so when don't need to read the reply
-	// https://www.postgresql.org/docs/current/protocol-flow.html#id-1.10.6.7.10
 	_, err = cancelConn.Write(buf)
-	return err
+	if err != nil {
+		return err
+	}
+
+	_, err = cancelConn.Read(buf)
+	if err != io.EOF {
+		return err
+	}
+
+	return nil
 }
 
 // WaitForNotification waits for a LISTON/NOTIFY message to be received. It returns an error if a notification was not
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -935,17 +935,16 @@ func (pgConn *PgConn) CancelRequest(ctx context.Context) error {
 	buf := make([]byte, 16)
 	binary.BigEndian.PutUint32(buf[0:4], 16)
 	binary.BigEndian.PutUint32(buf[4:8], 80877102)
-	binary.BigEndian.PutUint32(buf[8:12], uint32(pgConn.pid))
-	binary.BigEndian.PutUint32(buf[12:16], uint32(pgConn.secretKey))
-	_, err = cancelConn.Write(buf)
-	if err != nil {
-		return err
+	binary.BigEndian.PutUint32(buf[8:12], pgConn.pid)
+	binary.BigEndian.PutUint32(buf[12:16], pgConn.secretKey)
+
+	if _, err := cancelConn.Write(buf); err != nil {
+		return fmt.Errorf("write to connection for cancellation: %w", err)
 	}
 
-	_, err = cancelConn.Read(buf)
-	if err != io.EOF {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


jackc:
As far as I can tell, the current behavior is still within acceptable behavior for PostgreSQL. The change this reverts fails on Windows. I'm not sure exactly why. @drakkan worked on this change. Perhaps they could elaborate.

But it is surprising that not reading could be a problem -- there are no bytes to read. The only reason the code originally read was to wait for the server to close the connection.

levakin:
```
pgconn_test.go:2253: 
        	Error Trace:	D:\a\pgx\pgx\pgconn\pgconn_test.go:2253
        	            	read tcp [::1]:56732->[::1]:5432: wsarecv: An existing connection was forcibly closed by the remote host.
        	Test:       	TestConnCancelRequest
```
@drakkan network seems to work differently on windows. Any ideas why?

drakkan:
Hi,

This change was made to fix test cases on Windows. I thought the error was related to the raw system calls used at the time, but I see this happening even now that we no longer have platform-specific code

levakin:
hi, @jackc I've added handling of of this specific error. What do you think of it? Can you please run workflow again?

levakin:
@jackc 

> The only reason the code originally read was to wait for the server to close the connection.

That's exactly the reason I want to revert this commit. 
My current understanding is that due to the usage of PGbouncer, the connection is not fast enough to deliver the instruction and `defer cancelConn.Close()` closes the connection than the instruction is delivered. And the query keeps running as a result.

Here's the pgbouncer log:
```
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:51.238 UTC [1] LOG S-0x5631ef4891d0: shard1/postgres@172.22.0.2:5432 new connection to server (from 172.22.0.6:33088)
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:54.005 UTC [1] LOG C-0x5631ef481e80: shard1/postgres@172.22.0.1:57294 closing because: client close request (age=345s)
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:54.005 UTC [1] LOG S-0x5631ef4891d0: shard1/postgres@172.22.0.2:5432 closing because: client disconnect while server was not ready (age=2s)
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:54.005 UTC [1] LOG C-0x5631ef481e80: (nodb)/(nouser)@172.22.0.1:38600 closing because: failed cancel request (age=0s)
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:54.506 UTC [1] LOG C-0x5631ef481e80: shard1/postgres@172.22.0.1:38606 login attempt: db=shard1 user=postgres tls=no
kpis-pgbouncer-shard1-1  | 2023-10-09 14:42:54.508 UTC [1] LOG S-0x5631ef4891d0: shard1/postgres@172.22.0.2:5432 new connection to server (from 172.22.0.6:33094)
```

za-arthur:
FYI here is how Postgres libpq sends cancel request:
https://github.com/postgres/postgres/blob/REL_16_0/src/interfaces/libpq/fe-connect.c#L4946-L4960

Looks like it waits for a response from a server and ignores all errors. AFAIK EINTR can be ignored by pgx since go runtime handles it.

timyakjuju:
not a bad idea

jackc:
LGTM.

PR #1769, opened 2023-10-13
Enable SSH

jackc:
Trying to debug the flickering TestConnConcurrency on CI


jackc:
Was able to duplicate error locally.

PR #1776, opened 2023-10-20
Proposal: CopyFromCh

robfordww:
Summary
This pull request proposes a new helper function to extend the functionality of CopyFromSlice() and CopyFromRow(). The new function aims to facilitate streaming data to the database via channels, which is particularly useful in concurrent programming scenarios.

### DIFF
--- a/copy_from.go
+++ b/copy_from.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"reflect"
 
 	"github.com/jackc/pgx/v5/internal/pgio"
 	"github.com/jackc/pgx/v5/pgconn"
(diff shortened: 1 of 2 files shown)


jackc:
I don't this is quite the right abstraction. Usually, its best for libraries to expose non-channel interfaces and let the application add channels on top if desired.

In this case there are a couple concerning points.

1. The `CopyFromSource` interface isn't expected to block. But the channel could block indefinitely. The context passed to `CopyFrom` couldn't cancel this and then `CopyFrom` could block.
2. Someone may want to pass `[]any` instead of `T` (similar to `CopyFromRows`).

If anything should be added I think a better interface might be something like:

```go
func CopyFromFunc(next func() ([]any, error)) CopyFromSource
```

The `next` function could read from a channel. `next` could use a closure to gain access to a cancellation channel (like `ctx.Done()`) to avoid blocking. `next` could also handle `[]any`, a struct with reflect, or a struct manually converted to `[]any`.

`CopyFromFunc` could also be used to read from any other source.

robfordww:
Yeah, I support your thinking. Makes sense.  
The one thing I was thinking of is how to properly handle closing of channel, and if such an event should be registered as an "error".  With a return signature of` func() ([]any, err)`, one would have to catch a special type of error for the closing of channels to make the` Next()` method return false, and zero that particular error type unless one would want that to show up from the `Err() `method?!

jackc:
AFAICT, the code is fine, but the comments still refer to the previous version. It also should really be tested.

> The one thing I was thinking of is how to properly handle closing of channel, and if such an event should be registered as an "error". With a return signature of func() ([]any, err), one would have to catch a special type of error for the closing of channels to make the Next() method return false, and zero that particular error value unless one would want that to show up from the Err() method?!

Well, that's a great example of why the `CopyFromFunc` is a good abstraction. In most cases I would expect closing a channel not to be an error. It's simply the signal that the copy is finished. But there could be rare cases where closing the channel would signal an error. Each caller can choose the appropriate behavior for itself.

robfordww:
```

      closeChanErr := errors.New("closed channel")
	copyCount, err := conn.CopyFrom(context.Background(), pgx.Identifier{"foo"}, []string{"a"},
		pgx.CopyFromFunc(func() ([]any, error) {
			v, ok := <-dataCh
			if !ok {
				return nil, closeChanErr
			}
			return []any{v}, nil
		}))
```
This won't work, since when any error occurs the CopyFrom() seems to abort when an error is detected, and no data is written to database. The approach does seem to work if 

`func (g *copyFromFunc) Err() error {
	return nil
}`

is hardcoded to nil.

```

jackc:
> This won't work, since when any error occurs the CopyFrom() seems to abort when an error is detected, and no data is written to database.

Correct. Any returned error aborts the copy. If a closed channel is not an error then it should `return nil, nil`.

What I think is missing in the current design is a non-error termination signal. I would expect the function returning `nil, nil` would be the appropriate "end of data stream" signal.

jackc:
LGTM

Issue #1779, opened 2023-10-24
How to use arrays with v5/stdlib


regeda:
The documentation declares the following:

> ArrayCodec implements support for arrays. If pgtype supports type T then it can easily support []T by registering an 
> ArrayCodec for the appropriate PostgreSQL OID. In addition, Array[T] type can support multi-dimensional arrays. 

I'm trying to use the following:
```go
type X struct {
  ids []pgtype.UUID
}

func smth(x X) error {
  // "conn" is established using `v5/stdlib`
  _, err := conn.ExecContext("....$1....", x.ids)
  return err
}
```

However, I'm getting the following:
```
sql: converting argument $1 type: unsupported type []pgtype.UUID, a slice of struct
```

How to use arrays in both ways `Valuer` and `Scanner`?

jackc:
You can pass Go arrays directly as query arguments. For scanning you need to use `pgtype.Map.SQLScanner`.

```go
package main

import (
	"database/sql"
	"log"
	"os"

	"github.com/jackc/pgx/v5/pgtype"
	_ "github.com/jackc/pgx/v5/stdlib"
)

func main() {
	db, err := sql.Open("pgx", os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()
	m := pgtype.NewMap()

	uuidsIn := []pgtype.UUID{
		{Bytes: [16]byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}, Valid: true},
		{Bytes: [16]byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15}, Valid: true},
	}

	var uuidsOut []pgtype.UUID

	err = db.QueryRow("select $1::uuid[]", uuidsIn).Scan(m.SQLScanner(&uuidsOut))
	if err != nil {
		log.Fatal(err)
	}

	log.Println(uuidsOut)
}
```

Output:
```
2023/10/26 20:38:15 [{[1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 0] true} {[1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 0] true}]
``

regeda:
Got it!

@jackc Do you have any plans to avoid using SQLScanner in future? It makes the data model layer not interoperable if different SQL drivers are in place.

jackc:
[...]

PR #1786, opened 2023-10-31
Add Scan to pgtype.FlatArray

regeda:
Make `pgtype.FlatArray` respectful to `sql.Scanner` interface which makes sense when the standard `sql` package in use.

See https://github.com/jackc/pgx/issues/1779

### DIFF
--- a/pgtype/array.go
+++ b/pgtype/array.go
@@ -458,3 +458,8 @@ func (a FlatArray[T]) ScanIndex(i int) any {
 func (a FlatArray[T]) ScanIndexType() any {
 	return new(T)
 }
+
+func (a *FlatArray[T]) Scan(src any) error {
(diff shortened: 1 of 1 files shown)


jackc:
This seems to have broken a couple tests. Not immediately obvious why.

But even without that I'm not sure if this is the right approach. I would expect it to work with most of the built-in types, but not all. And if the user has customized their pgtype.Map it would mysteriously fail.

There's a fundamental issue that the with database/sql we don't have access to the type information from the database. And with a generic array type we don't have access to a specific Go type that we know corresponds to a specific PostgreSQL type.

I suspect that the solution will have to be in the stdlib package either as concrete types for each array type or a generic array type that requires the underlying type to have an additional method.

rubiagatra:
I think we can close it @jackc,  see #2020 and we are waiting https://github.com/golang/go/issues/67546

PR #1804, opened 2023-11-21
Fix usage of logger in stdlib docs

SamWhited:
The documentation previously showed the old way of logging and not the newer tracer adapter. This is a minor patch updates the example to build correctly with pgx/v5.


jackc:
Thanks!

Issue #1758, opened 2023-10-05
can't Scan amount string of 10,000,000,000 pgytpe.Numeric


walvisk:
I have this below code. I want to convert from big.Float value to pgtype.Numeric but from it's string format
```
package main

import (
	"errors"
	"fmt"
	"math/big"
	"strings"

	"github.com/jackc/pgx/v5/pgtype"
)

func parseAmount(amountStr string) (big.Float, error) {
	flt := new(big.Float)
	newStr := strings.ReplaceAll(amountStr, ",", "")
	amount, isSuccess := flt.SetString(newStr)
	if !isSuccess {
		return *flt, errors.New("parse error")
	}
	return *amount, nil
}

func parseToNumeric(amountStr string) (pgtype.Numeric, error) {
	var amount pgtype.Numeric

	if err := amount.Scan(amountStr); err != nil {
		return amount, err
	}

	return amount, nil
}

type Transaction struct {
	Amount *big.Float
}

func main() {
	dump := []string{"10,000,000,000", "trx1"}
	amount, err := parseAmount(dump[0])
	if err != nil {
		fmt.Println(err)
	}

	trx := Transaction{Amount: &amount}
	numericAmount, err := parseToNumeric(trx.Amount.String())
	if err != nil {
		fmt.Printf("An error %v", err)
	}
	fmt.Println(numericAmount)
}
```
It produce errors `An error 1e+1 is not a number`. 

But when i change the string amount to 9999999999 (10000000000 - 1), it does not produce error.
Is Scan for pgtype only can scan up to 9999999999? 

could you guys please help me on this? i really appreciate it. thanks

yungtrizzle:
Did you solve this?

PR #1809, opened 2023-11-23
support scientific notation big floats

yungtrizzle:
Adds some support following discovery of https://github.com/jackc/pgx/issues/1758 by
1. detecting scientific notation
2. scan into big/Float
3. replace original string with a true numeric string
4. ensure using minimal significant figures after the decimal

### DIFF
--- a/pgtype/numeric.go
+++ b/pgtype/numeric.go
@@ -762,7 +762,7 @@ func (scanPlanTextAnyToNumericScanner) Scan(src []byte, dst any) error {
 	if strings.ContainsAny(string(src), "eE") {
 		if bigF, ok := new(big.Float).SetString(string(src)); ok {
 			smallF, _ := bigF.Float64()
-			src = []byte(strconv.FormatFloat(smallF, 'f', -1, int(bigF.Prec())))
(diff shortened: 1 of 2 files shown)


jackc:
Does PostgreSQL ever return numerics in scientific notation? If not, I'm not sure that Scan is the right place for that logic.

yungtrizzle:
> Does PostgreSQL ever return numerics in scientific notation? If not, I'm not sure that Scan is the right place for that logic.

Not necessarily but Scan is the only method to instantiate a numeric when it's not a returned value from a query. This is a common case for sending numerics to PostgreSQL

jackc:
Would it make sense to instead add a `ScanScientific` or something like that? It seems slightly off to me to make a function whose purpose is to scan values PostgreSQL produces also do something else. `Numeric` already has a couple `Scan*` methods so it wouldn't be unprecedented.

Also, I'm a little surprised that `Numeric` is being used directly. It is included so pgx can represent numeric values without an external dependency, but I expected anyone who was actually using PostgreSQL numerics would use something like https://github.com/shopspring/decimal.

PR #1811, opened 2023-11-25
for performance: change JSONBCodec format to BinaryFormatCode.

piyongcai:
just for performance. because json.Unmarshal need []byte parameter


jackc:
In general, I don't want to merge any performance optimizations without a benchmark that shows the improvement.

That's especially true in this case because the text format was chosen on purpose. PostgreSQL (at least historically) has a significant performance penalty when returning text-like types as binary. See https://github.com/jackc/pg_text_binary_bench.

rubiagatra:
Do you still want to continue @piyongcai? or we can close the PR?

Issue #1572, opened 2023-04-10
Wrap Scan errors


injeniero:
### Discussed in https://github.com/jackc/pgx/discussions/1567

<div type='discussions-op-text'>

<sup>Originally posted by **injeniero** April  4, 2023</sup>
Here the returned error from the scanner is converted to string, I would like it to be wrapped so you can search for it later using `errors.Is`
https://github.com/jackc/pgx/blob/master/pgtype/pgtype.go#L2048
```Go
- return fmt.Errorf("unable to encode %#v into %s format for %s (OID %d): %s", value, format, dataTypeName, oid, err)
+ return fmt.Errorf("unable to encode %#v into %s format for %s (OID %d): %w", value, format, dataTypeName, oid, err)
```

This would be used like: 
```Go
func (s *MyType) Scan(value interface{}) error {
	if value == nil {
		return nil
	}
	...

	return fmt.Errorf("unknown type: %s, %w", value, ErrInvalidMyType)
}
```
then somewhere else in the code:
```Go


db.Insert(ctx, myType)
if errors.Is(err, ErrInvalidProductSupport) {
   //do something for this special case
} else {
   // unknown error type
}

```</div>

PR #1813, opened 2023-11-27
Fix/error wrapping

injeniero:
Error wrapping instead of just formatting the errors.

Also, fixed two cases of bad error messages.


injeniero:
@jackc this is a follow up PR to fix all places where errors should be wrapped. Original issue: https://github.com/jackc/pgx/issues/1572

jackc:
Thanks

Issue #1820, opened 2023-12-02
Missing `QueryTracer` `TraceQueryEnd()` call in conn `Query()` func


renanferr:
**Describe the bug**
The Query Tracer is not working properly with the `Query()` connection func. The current unit tests are using `QueryRow()` and actually there are missing calls for the `TraceQueryEnd()` func for several exec modes.

**To Reproduce**
Just run `conn.Query()` with an active query tracer and check if `TraceQueryEnd()` was called.

I wrote a unit test for this case in #1821 to better demonstrate.

Runnable example to reproduce:

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
)

type testTracer struct {
	traceQueryStart    func(ctx context.Context, conn *pgx.Conn, data pgx.TraceQueryStartData) context.Context
	traceQueryEnd      func(ctx context.Context, conn *pgx.Conn, data pgx.TraceQueryEndData)
	traceBatchStart    func(ctx context.Context, conn *pgx.Conn, data pgx.TraceBatchStartData) context.Context
	traceBatchQuery    func(ctx context.Context, conn *pgx.Conn, data pgx.TraceBatchQueryData)
	traceBatchEnd      func(ctx context.Context, conn *pgx.Conn, data pgx.TraceBatchEndData)
	traceCopyFromStart func(ctx context.Context, conn *pgx.Conn, data pgx.TraceCopyFromStartData) context.Context
	traceCopyFromEnd   func(ctx context.Context, conn *pgx.Conn, data pgx.TraceCopyFromEndData)
	tracePrepareStart  func(ctx context.Context, conn *pgx.Conn, data pgx.TracePrepareStartData) context.Context
	tracePrepareEnd    func(ctx context.Context, conn *pgx.Conn, data pgx.TracePrepareEndData)
	traceConnectStart  func(ctx context.Context, data pgx.TraceConnectStartData) context.Context
	traceConnectEnd    func(ctx context.Context, data pgx.TraceConnectEndData)
}

func (tt *testTracer) TraceQueryStart(ctx context.Context, conn *pgx.Conn, data pgx.TraceQueryStartData) context.Context {
	if tt.traceQueryStart != nil {
		return tt.traceQueryStart(ctx, conn, data)
	}
	return ctx
}

func (tt *testTracer) TraceQueryEnd(ctx context.Context, conn *pgx.Conn, data pgx.TraceQueryEndData) {
	if tt.traceQueryEnd != nil {
[...]

PR #1821, opened 2023-12-02
rename TestTraceQuery test to TestTraceQueryRow and add new TestTraceQuery

renanferr:
This PR demonstrates the bug reported in #1820.

The original TestTraceQuery test was actually using the QueryRow func and turns out that the Query func may not be calling trace hooks properly, with missing calls to QueryTracerEnd at the moment.

So I renamed it to `TestTraceQueryRow` and added a new `TestTraceQuery` test case which actually uses the `Query()` func.

As you can see, `TestTraceQueryRow`, which is the original `TestTraceQuery` works smoothly. However, the new `TestTraceQuery` fails.

### DIFF
--- a/tracer_test.go
+++ b/tracer_test.go
@@ -137,7 +137,7 @@ func TestTraceExec(t *testing.T) {
 	})
 }
 
-func TestTraceQuery(t *testing.T) {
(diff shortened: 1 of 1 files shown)


renanferr:
Closing due to #1820 being closed.

PR #1824, opened 2023-12-04
Batch.Queue: document always uses the conn's DefaultQueryExecMode

evanj:
The only way to change the query mode used by Batch.Queue and SendBatch is to use a connection with a different
DefaultQueryExecMode. Add this to the function documentation.

Conn.SendBatch: Move where mode is defined to make this clearer in the code. I spent time looking for the option that does not exist.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -61,6 +61,8 @@ type Batch struct {
 }
 
 // Queue queues a query to batch b. query can be an SQL query or the name of a prepared statement.
+// The only pgx option argument that is supported is QueryRewriter. Queries are executed using the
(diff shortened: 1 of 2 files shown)


jackc:
Good to have it documented more clearly.

If we did want to be able to control the execution mode per batch then that probably could be implemented via an additional field on the pgx.Batch struct fairly easily.

evanj:
Sounds great thank you. Right now I'm just experimenting with this, and haven't needed to change the execution mode. If I do, I'll submit a PR. Thanks as always for the fast response.

PR #1825, opened 2023-12-04
pgconn: add LastError()

### LINKED ISSUE #1803: Destroy connection on particular errors
There are certain types of errors that might cause a developer to want to destroy a connection. Unfortunately, when you're using a pool its not easy to destroy a particular connection when an error is received. This is because the pool does:

```go
	rows, err := c.Query(ctx, sql, args...)
	if err != nil {
		c.Release()
		return errRows{err: err}, err
	}
```

So it releases the connection before the caller has a chance to assess the error.

There are a few solutions that I see:

1. Adding a LastError to pgconn.PgConn function that returns the last `*PgError` generated from a received `*pgproto3.ErrorResponse` then a developer could setup a `AfterRelease` function that calls that function and returns false if the last error was error that should trigger a destroy. The odd part about this is that an ignored error might be returned from LastError indefinitely if another error never occurs. I'm not sure if this is a problem but it is a bit odd.

2. Adding yet another callback to a pool that gets called on errors and destroys a connection if the function returns true. This feels much more difficult since there are a lot of spots that a connection could error and we'd have to add something to all of those spots. This also feels a bit too bespoke versus the other solution.

I had made a map of connections -> errors and then the next time I see a connection in `AfterRelease` that had an error I wanted to destroy, I'd return false. But the way the releasing works, I'd only be able to observe the error after the connection was put back in the pool so I'd always be 1 error behind.

**Additional context**
The particular errors that we are trying to destroy based on is a shutdown error as well as an unknown session error that seems specific to yugabyte.

jameshartig:
LastError() returns the last error encountered by the underlying connection or received from postgres. It is cleared when a new request is initiated.

Fixes #1803

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -87,6 +87,8 @@ type PgConn struct {
 
 	peekedMsg pgproto3.BackendMessage
 
+	lastError error
+
 	// Reusable / preallocated resources
 	resultReader      ResultReader
 	multiResultReader MultiResultReader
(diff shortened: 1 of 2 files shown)


jameshartig:
Closing in favor of https://github.com/jackc/pgx/pull/1826. See https://github.com/jackc/pgx/issues/1803#issuecomment-1841283324

PR #1835, opened 2023-12-13
pgconn: normalize more connection errors

samuel:
Normalize the error that is returned by startTLS in pgconn.connect. This makes it possible to determine if the error was a context error. Before this change the returned error for a context that is canceled or hits the deadline would be `tls error (read tcp 127.0.0.1:55914->127.0.0.1:5432: i/o timeout)`.

This change also uses errors.As to unwrap errors in normalizeTimeoutError. This allows the function to handle the `writeError` type that shows up as `failed to write startup message (read tcp 127.0.0.1:55914->127.0.0.1:5432: i/o timeout)`.

### DIFF
--- a/pgconn/errors.go
+++ b/pgconn/errors.go
@@ -107,14 +107,15 @@ func (e *parseConfigError) Unwrap() error {
 }
 
 func normalizeTimeoutError(ctx context.Context, err error) error {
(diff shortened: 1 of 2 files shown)


jackc:
LGTM - thanks!

PR #1846, opened 2023-12-22
Add snake_case support to RowToStructByName

tikhonfedulovv:
RowToStructureByName doesn't support the conversion of snake_case to PascalCase, which is used in Go.

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -668,11 +668,11 @@ func fieldPosByName(fldDescs []pgconn.FieldDescription, field string) (i int) {
 	i = -1
 	for i, desc := range fldDescs {
 
-		// Support snake and spinal case.
-		desc.Name = strings.ReplaceAll(desc.Name, "_", "")
-		desc.Name = strings.ReplaceAll(desc.Name, "-", "")
+		// Snake case support.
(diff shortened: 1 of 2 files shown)


jackc:
* It looks like this broken existing tests / behavior.
* It would be good to have tests for this new behavior?
* Is the `kebab-case` support necessary? That is only possible in PostgreSQL if using double-quotes around the identifiers. I've never actually seen it in the wild.

tikhonfedulovv:
> * It looks like this broken existing tests / behavior.
> * It would be good to have tests for this new behavior?

Yeah, it would be great. I didn't think that such a change would break anything

> * Is the `kebab-case` support necessary? That is only possible in PostgreSQL if using double-quotes around the identifiers. I've never actually seen it in the wild.

No, I don't think so. I've never seen it too 😃

tikhonfedulovv:
I'll write TestFieldPosByName later

jackc:
I'll merge this as is.

I am slightly concerned about performance as this adds two string allocations for every assignment. But I think there is already some reflection in that path so it might not matter.

PR #1875, opened 2024-01-14
Support really large objects

### LINKED ISSUE #1865: Unable to write large objects
**Describe the bug**

I am unable to write 3 GB large object into the database.

**To Reproduce**

See [this repository](https://github.com/mitar/pgx-large-objects). Run:

```sh
$ docker run --name pgsql --rm -d --network=host -e LOG_TO_STDOUT=1 -e PGSQL_ROLE_1_USERNAME=test -e PGSQL_ROLE_1_PASSWORD=test -e PGSQL_DB_1_NAME=test -e PGSQL_DB_1_OWNER=test registry.gitlab.com/tozd/docker/postgresql:16
$ go run ./...
```

**Expected behavior**

But large objects are correctly stored.

**Actual behavior**

The first (small) value is stored. The second (3 GB large value) is not. I get the error:

```
2024/01/04 21:42:04 write failed: write tcp 127.0.0.1:45636->127.0.0.1:5432: write: connection reset by peer
```

In PostgreSQL logs I can also see:

```
{"database":"test","level":"LOG","msg":"invalid message length","pid":44,"time":"2024-01-04T20:42:04.533Z","user":"test","service":"postgresql","stage":"run","logged":"2024-01-04T20:42:04.533Z"}
```

**Version**
 - Go: go version go1.21.3 linux/amd64
 - PostgreSQL: PostgreSQL 16.1 (Ubuntu 16.1-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
 - pgx: v5.5.1

**Additional context**

Interestingly, there is another error:

```
{"database":"[unknown]","level":"LOG","msg":"PID 44 in cancel request did not match any process","pid":45,"time":"2024-01-04T20:42:04.537Z","user":"[unknown]","service":"postgresql","stage":"run","logged":"2024-01-04T20:42:04.537Z"}
```

This happens because code calls `e.dbpool.Close()` always, even on error. And I am not sure why pgx tries to close a connection which does not exist anymore, it seems? At least this is how I am reading this log message.

mitar:
Fixes #1865.

### DIFF
--- a/large_objects_test.go
+++ b/large_objects_test.go
@@ -13,7 +13,8 @@ import (
 )
 
 func TestLargeObjects(t *testing.T) {
-	t.Parallel()
+	// We use a very short limit to test chunking logic.
+	pgx.SetMaxLargeObjectMessageLength(t, 2)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
 	defer cancel()
@@ -34,7 +35,8 @@ func TestLargeObjects(t *testing.T) {
 }
 
 func TestLargeObjectsSimpleProtocol(t *testing.T) {
-	t.Parallel()
+	// We use a very short limit to test chunking logic.
+	pgx.SetMaxLargeObjectMessageLength(t, 2)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
 	defer cancel()
@@ -160,7 +162,8 @@ func testLargeObjects(t *testing.T, ctx context.Context, tx pgx.Tx) {
 }
 
 func TestLargeObjectsMultipleTransactions(t *testing.T) {
-	t.Parallel()
+	// We use a very short limit to test chunking logic.
+	pgx.SetMaxLargeObjectMessageLength(t, 2)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
 	defer cancel()
--- a/large_objects.go
+++ b/large_objects.go
@@ -6,6 +6,11 @@ import (
 	"io"
 )
 
+// The PostgreSQL wire protocol has a limit of 1 GB - 1 per message. See definition of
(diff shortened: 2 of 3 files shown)


large_objects_test.go
```diff
@@ -7,23 +7,24 @@ import (
...
 	defer cancel()
 
-	conn, err := pgx.Connect(ctx, os.Getenv("PGX_TEST_DATABASE"))
+	conn, err := Connect(ctx, os.Getenv("PGX_TEST_DATABASE"))
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	pgxtest.SkipCockroachDB(t, conn, "Server does support large objects")
+	// pgxtest.SkipCockroachDB(t, conn, "Server does support large objects")
```
mitar, line 27: Importing `pgxtest` creates an import loop. Should I move this function to some other package? Like `pgxtest_cockroach`?

jackc: The solution is to make a new file such as large_objects_private_test.go that is in the pgx package. This file should contain a public function that changes maxLargeObjectMethodLength.

jackc:
The large object tests should not be parallel anymore with the changes to the global variable `maxLargeObjectMessageLength`. In addition, they should use a `t.Cleanup()` to restore the original value.

jackc:
The implementation itself is fine as far as I can tell.

mitar:
> The large object tests should not be parallel anymore with the changes to the global variable maxLargeObjectMessageLength. In addition, they should use a t.Cleanup() to restore the original value.

I was thinking about that, but they are all set to the same value. So does it really matter for the purpose of tests?

mitar:
I updated PR and I think it is ready for review/merge.

PR #1879, opened 2024-01-18
feat: Add BatchN to create a batch with an initial capacity.

s-l-teichmann:
This PR adds a function BatchN which returns a batch with a pre-allocated capacity.
I ran more than once into the situation that I know beforehand how large my Batch is going to be.
As there is no way to access the slices holding  the queries I've added this kind of
constructor. Maybe you find this helpful.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -72,6 +72,13 @@ func (b *Batch) Queue(query string, arguments ...any) *QueuedQuery {
 	return qq
 }
 
+// BatchN returns a new Batch with an initial capacity of n.
+// This is useful to avoid allocations if you know beforehand how many queries
+// at least you want to enqueue.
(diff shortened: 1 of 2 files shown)


jackc:
Have you benchmarked any significant improvements with this change? Given all that goes on in queuing the queries, sending the batch, and reading the results of all the queries, it would surprise me if removing saving the allocation was noticeable.

jackc:
Also, please take a look at https://github.com/jackc/pgx/discussions/1878. If that proposal comes to fruition that would also allow the preallocation you are looking for.

pashagolub:
Can be solved by #1886

jackc:
Closed in favor of #1886.

PR #1885, opened 2024-01-23
add ltree pgtype support

luxifer:
This PR backports the `ltree` type from [pgtype](https://github.com/jackc/pgtype/blob/master/ltree.go).


jackc:
My guess is that anywhere `create extension hstore` occurs we would also need `create extension ltree`;

PR #1897, opened 2024-02-05
Add a new example usage that reminds the user to utilize the escape method for the connection string

karminski:
For certain Golang database driver libraries, such as [go-sql-driver](https://github.com/go-sql-driver/mysql?tab=readme-ov-file#password) the connection string is automatically escaped by the library itself. Therefore, I believe it is crucial to inform users whether the connection string requires escaping or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,14 +20,22 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 
 	"github.com/jackc/pgx/v5"
 )
 
 func main() {
-	// urlExample := "postgres://username:password@localhost:5432/database_name"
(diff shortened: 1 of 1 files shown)


jackc:
The example would no longer represent idiomatic usage. If the application is has the host, user name, password, etc. individually then it should call `pgx.ParseConfig("")` and manually set the fields on the returned `*pgx.ConnConfig`. In this case, the password would not need to be escaped.

jackc:
I think that adding a note to the `ParseConfig` function to this effect would be better. Possible the note would simply mention that URLs are parsed by the Go standard URL parser and must be properly escaped.

rubiagatra:
Do you still want to continue the PR @karminski ?

PR #1901, opened 2024-02-05
error returned by context watcher

zcolleen:


jackc:
As mentioned a few months ago in the other thread, I think it is fundamentally incorrect for a query method to return an error when we know the query has succeeded. Sorry, but even if it is not the default, I don't want to encourage bad behavior.

Issue #1924, opened 2024-03-03
5.5.3 changed the behavior of `CollectRows` on an empty result


its-felix:
**Describe the bug**
Since version 5.5.3 with the introduction of https://github.com/jackc/pgx/commit/a57bb8caeab831f71b25d77c9dde5d430226cf49 , the behavior of `CollectRows` changed. It now returns a `nil` slice if the passed `Rows` contains no rows.

**To Reproduce**
- Query something that doesn't return any rows
- Use `pgx.CollectRows`
- `nil` slice is returned

**Expected behavior**
An empty slice should be returned

**Actual behavior**
A `nil` slice is returned

**Version**
 - Go: 1.21
 - PostgreSQL: CockroachDB 23.1.15
 - pgx: 5.5.3

**Additional context**
N/A

ldemailly:
Curious why is a nil slice an issue? slices whether nil or empty have the exact same behavior wrt to len() or further append() and relying on one or the other representation is a bad idea

Ie if you want to json serialize to `null` for instance (instead of [], or vice versa), then check the len() first etc...

its-felix:
@ldemailly the issue here was that the behavior changed from 5.5.2 to 5.5.3 - where until 5.5.2 an empty slice was returned it changed to a nil slice with 5.5.3

I agree that in general a nil slice is mostly equal to a empty slice (append, len, for loops etc), but is different when serialized using `encoding/json`

In my case, I'm using the slice in a JSON HTTP Response where the update from 5.5.2 to 5.5.3 broke my API contract.

Such a change can of course be made and might actually make sense here, but it should ideally only be made in major releases (ie 5.x to 6.x)

ldemailly:
I know that every behavior/side effect on something widely used gets accidentally relied upon (Hyrum's Law) and also agree it's safer to keep the same across minor - but I would strongly suggest if you need an nil instead of [] to change the caller code to check len() and act accordingly, it'll make your code more correct and able to handle a different way to get a slice in the future

![obligatory xkcd](https://imgs.xkcd.com/comics/workflow.png)

PR #1925, opened 2024-03-03
Fix behavior of CollectRows to return empty slice if Rows are empty

its-felix:
https://github.com/jackc/pgx/issues/1924


erezrokah:
Thanks for the PR @its-felix and the quick merge @jackc. Will a new tag/release created with this fix?

The change in behavior is especially impactful when JSON serializing the response (like in a REST API backend server) since then instead of an empty array you'd get a null value which is unexpected (somehow released to https://github.com/golang/go/issues/37711)

its-felix:
That's also the exact reason why it caught my eye, leading to this PR and why I rolled back to 5.5.2 in my project.

jackc:
v5.5.4 was just released.

Issue #1928, opened 2024-03-05
SET LOCAL with parameters fails with syntax error


gregwoodfill:
**Describe the bug**
When calling conn.Exec to set a local postgres with parameters, a syntax error is returned.
This started happening in v5.5.4 and looks to be related to adding parentheses around parameter values from this commit: https://github.com/jackc/pgx/commit/c543134753a0c5d22881c12404025724cb05ffd8
Issuing SET LOCAL with a value wrapped in parentheses is not valid Postgres syntax.

**To Reproduce**

```go
package main

import (
	"context"
	"github.com/jackc/pgx/v5"
	"log"
	"os"
)

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(context.Background())
	connMetadata := "some metadata"
	_, err = conn.Exec(context.Background(), "set local someMeta to $1", pgx.QueryExecModeSimpleProtocol, connMetadata)
	if err != nil {
		log.Fatal(err)
	}
}
```

Please run your example with the race detector enabled. For example, `go run -race main.go` or `go test -race`.

```
2024/03/05 16:52:02 ERROR: syntax error at or near "(" (SQLSTATE 42601)
```

**Expected behavior**
conn.Exec executes the statement without error.

**Actual behavior**
conn.Exec returns a syntax error.
```
syntax error at or near "(" (SQLSTATE 42601)
```

**Version**
 - Go: v1.22.0
 - PostgreSQL: PostgreSQL 16.1 (Debian 16.1-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
 - pgx: v5.5.4

**Additional context**

Appears to have been introduced here: https://github.com/jackc/pgx/commit/c543134753a0c5d22881c12404025724cb05ffd8
PostgreSQL documentation on SET LOCAL: https://www.postgresql.org/docs/current/sql-set.html

michaelcmelton:
Same issue and error is returned with the `INTERVAL` syntax as well.

nickpalmer:
This change:
https://github.com/jackc/pgx/commit/c543134753a0c5d22881c12404025724cb05ffd8#diff-65a99c1d9a3b358d8aebc18eb1f6d77c6142f66cdb2f330b66577681b38365ffR69

Should probably be something like:
```
[...]

PR #1932, opened 2024-03-06
fix: issue 1928

nickpalmer:
This PR fixes the issues raised in #1928 and should prevent any other attacks using whitespace tricks as raised in the vulnerability report which initial prompted the breaking parenthesis addition.

It uses postgresql escape strings to prevent whitespace embedded in a value from being passed through to the SQL parser.

See: https://www.postgresql.org/docs/9.0/sql-syntax-lexical.html for information on this extension.

### DIFF
--- a/internal/sanitize/sanitize.go
+++ b/internal/sanitize/sanitize.go
@@ -66,7 +66,24 @@ func (q *Query) Sanitize(args ...any) (string, error) {
 
 			// Prevent SQL injection via Line Comment Creation
 			// https://github.com/jackc/pgx/security/advisories/GHSA-m7wr-2xf7-cm9p
-			str = "(" + str + ")"
+			// If the value starts with a dash then we need to put parens around it
+			if strings.HasPrefix(str, "-") {
+				str = "(" + str + ")"
+			}
+
+			// Convert problematic line breaks to escape form
+			// See: https://www.postgresql.org/docs/9.0/sql-syntax-lexical.html Table 4.1
+			if strings.ContainsAny(str, "\b\f\n\r\t") {
+				str = "E" + str
+				// Escape existing escapes
+				str = strings.ReplaceAll(str, `\`, `\\`)
+				// Convert existing line breaks to escaped
+				str = strings.ReplaceAll(str, "\b", `\b`)
+				str = strings.ReplaceAll(str, "\f", `\f`)
+				str = strings.ReplaceAll(str, "\n", `\n`)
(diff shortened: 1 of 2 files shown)


jackc:
The reason I went for parentheses around all values was to avoid any possibility of missing a dangerous value. It works properly for true query parameters. But as reported in #1928, it can break where the parameter is not actually a query parameter as PostgreSQL understands it.

With non-actual query parameters in mind, I don't think that parentheses are the right approach, even if only done for negative values. It is entirely possible for a `SET` statement as described in #1928 to be negative.

I think the actual approach is even simpler. Instead of wrapping with parentheses, wrap with spaces. That would prevent the accidental construction of a `--`. It would instead be a harmless `- -`. This would also prevent the accidental construction of a `''` (two single quotes) which could cause two literal strings to be parsed as a literal string with a single quote in the middle. (Space was also mentioned in the original vulnerability report as a potential solution - I went with parentheses to in preference to having an invisible character be meaningful - but in light of #1928 I think it is worth it)

As far as the escape strings go, I don't think that is actually necessary. A PostgreSQL literal string is allowed to have newlines or any other character aside from `\0` (`NUL` character). The only character that needs to be escaped is the `'` (single quote).

nickpalmer:
The space solution is an interesting one.

I went to escape strings because that seemed like it would prevent other attacks we haven't thought of involving new lines, and because it makes the log of those queries much cleaner since they then don't log with newlines.

The other solution I considered was using the join of strings to substitute newlines with a pasted escaped newline.

`str = strings.ReplaceAll(str, "\n", "'E'\\n''")`

internal/sanitize/sanitize.go
```diff
@@ -66,7 +66,24 @@ func (q *Query) Sanitize(args ...any) (string, error) {
...
 			// https://github.com/jackc/pgx/security/advisories/GHSA-m7wr-2xf7-cm9p
-			str = "(" + str + ")"
+			// If the value starts with a dash then we need to put parens around it
+			if strings.HasPrefix(str, "-") {
+				str = "(" + str + ")"
+			}
+
+			// Convert problematic line breaks to escape form
+			// See: https://www.postgresql.org/docs/9.0/sql-syntax-lexical.html Table 4.1
+			if strings.ContainsAny(str, "\b\f\n\r\t") {
```
dmsolutionsmn, line 76: In the scenario where the value has both a line break AND a - prefix, we get an invalid `E('value')` instead of `(E'value')`

jackc:
I ended up going for the space approach.

Issue #1929, opened 2024-03-06
Wrong host in ConnectError


ndtretyak:
**Describe the bug**
When a connection string contains multiple hosts, the `Config.Host` field is populated with the first host, while the remaining hosts are treated as `FallbackConfig`s. The `connect` function, which accepts both `Config` and `FallbackConfig` to establish a connection, uses the `Host` and `Port` from the FallbackConfig.

https://github.com/jackc/pgx/blob/da6f2c98f2664b215b40b1606551fdfcc7f3ea5c/pgconn/pgconn.go#L283

However, in the event of a connection error, the default `Config` is used, leading to a mismatch between the error message (which contains the IP address of the host we attempted to connect to) and the `Host` and `Port` from a different host.

https://github.com/jackc/pgx/blob/da6f2c98f2664b215b40b1606551fdfcc7f3ea5c/pgconn/pgconn.go#L286

**To Reproduce**

```go
package main

import (
	"context"
	"log"

	"github.com/jackc/pgx/v5"
)

const url = "postgresql://user:password@localhost:1234,127.0.0.1:5678/database"

func main() {
	conn, err := pgx.Connect(context.Background(), url)
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(context.Background())
}
```

**Expected behavior**
The error message should match the host we attempted to connect to. Possible outcome are
- `host=localhost ... dial tcp 127.0.0.1:1234`
- `host=127.0.0.1 ... dial tcp 127.0.0.1:5678`

**Actual behavior**
The error message contains a mismatched host and port
```
2024/03/06 11:13:07 failed to connect to `host=localhost user=user database=database`: dial error (dial tcp 127.0.0.1:5678: connect: connection refused)
exit status 1
```

**Additional context**
It would be more informative to return an error group containing connection errors for all attempted hosts. This would be particularly useful when `target_session_attrs=primary` is used and a connection to the primary cannot be established. In this case we can get a useless `server is in standby mode` while connecting to the other node.

jackc:
> It would be more informative to return an error group containing connection errors for all attempted hosts.

As of Go 1.20 we now have a better way of returning multiple errors. That's a reasonable change. 👍

jackc:
I've added multiple error returns for connection failures in 8db971660e43a944c777d5e4fcfcb7885acb7566.

For example, the following connection string:

`host=localhost,127.0.0.1,foo.invalid port=1,2,3`

Will now return an error like the following:

```
[...]

PR #1937, opened 2024-03-10
Correct host in ConnectError

ndtretyak:
Added `*fallbackConfig` to `ConnectError` to have the correct host in the error message.
See https://github.com/jackc/pgx/issues/1929

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -152,15 +152,16 @@ func ConnectConfig(octx context.Context, config *Config) (pgConn *PgConn, err er
 	ctx := octx
 	fallbackConfigs, err = expandWithIPs(ctx, config.LookupFunc, fallbackConfigs)
 	if err != nil {
-		return nil, &ConnectError{Config: config, msg: "hostname resolving error", err: err}
+		lookupErr := err.(*lookupError)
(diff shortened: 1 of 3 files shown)


pgconn/pgconn.go
```diff
@@ -152,15 +152,16 @@ func ConnectConfig(octx context.Context, config *Config) (pgConn *PgConn, err er
...
 	fallbackConfigs, err = expandWithIPs(ctx, config.LookupFunc, fallbackConfigs)
 	if err != nil {
-		return nil, &ConnectError{Config: config, msg: "hostname resolving error", err: err}
+		lookupErr := err.(*lookupError)
+		return nil, &ConnectError{Config: config, fallbackConfig: lookupErr.fallbackConfig, msg: "hostname resolving error", err: lookupErr.err}
 	}
 
 	if len(fallbackConfigs) == 0 {
-		return nil, &ConnectError{Config: config, msg: "hostname resolving error", err: errors.New("ip addr wasn't found")}
+		return nil, &ConnectError{Config: config, fallbackConfig: fallbackConfigs[0], msg: "hostname resolving error", err: errIPAddrNotFound}
```
jackc, line 160: It seems referencing `fallbackConfigs[0]` when `len(fallbackConfigs) == 0` would always panic.

jackc:
Also, it appears that the change as a whole would make all error messages display the resolved IP address in exclusion of the name the IP was resolved from. This seems undesirable.

Issue #1942, opened 2024-03-13
Timeout error not show which fallback was used


laskoviymishka:
**Is your feature request related to a problem? Please describe.**

Same as here: https://github.com/jackc/pgconn/issues/139
Given: Cluster with ipv4 and ipv6 address.
when there is no network connection we should see what exact address we failed to connect

**Describe the solution you'd like**

Show IP-address that failed as part of error message.
```
failed to connect to `host=chinook.chiyepnb01it.eu-west-2.rds.amazonaws.com user=tutorial database=chinook`: dial error (timeout: dial tcp [2a05:d01c:d05:6500:3318:e752:f216:3d9b]:5432: i/o timeout)
```

**Describe alternatives you've considered**

Leave it as is 
```
failed to connect to `host=chinook.chiyepnb01it.eu-west-2.rds.amazonaws.com user=tutorial database=chinook`: dial error (timeout: context Deadline exceeded)
```

**Additional context**

This error also fixed at v4 driver - https://github.com/jackc/pgconn/pull/140

jackc:
I think this is fundamentally the same issue as #1929. And I think it should be solved by the same approach. `Connect` should return some sort of multi-error that includes the errors from all the connection attempts.

laskoviymishka:
This error is an exception, so solving same way as #1929 - is incorrect, since we try host sequential, once we reach first timeout all other host will receive same errTimeout automatically.

laskoviymishka:
In the meantime using parallel conn checker would break dual-stack logic.

I tried to implement parallel connect. In this approach we will catch the ipv4 address and start working with it, so in theory it's okay, but the `psql` does not work like this, so I would say that the current approach with a sequential check is right, but the error message can be improved.

jackc:
Yeah, I don't think a parallel connection attempt is a good idea. And I don't think the other PR that is proposed for #1929 is right either. But if the error message was a multi error that included all the attempts, I think that would solve this and #1929.

laskoviymishka:
Nah, multiple errors would be equal to no error, you will see 2 hosts that errored but in fact only one is not reachable.
Maybe it's a good idea to stop iterations after nor errTimeout but after errTimeout with context deadline exceeded. Will try to debug such impl.

jackc:
FWIW, multi errors are implemented in 8db971660e43a944c777d5e4fcfcb7885acb7566.

This will produce errors like:

```
    failed to connect to `user=postgres database=pgx_test`:
[...]

PR #1943, opened 2024-03-13
Expose host address when timeout exceeded

laskoviymishka:
Fixes: #1942

Once we face a timeout we should expose address, it can be for example ipv6 or ipv4 and knowing what exact address we took for resolver extreamly helpful for triaging a networking issue.

This issue is also fixed in pgx/v4, see: https://github.com/jackc/pgconn/issues/139

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -177,6 +177,7 @@ func ConnectConfig(octx context.Context, config *Config) (pgConn *PgConn, err er
 		pgConn, fberr = connect(ctx, config, fc, false)
 		if fberr == nil {
 			foundBestServer = true
+			err = nil
 			break
 		} else if pgerr, ok := fberr.(*PgError); ok {
 			err = &ConnectError{Config: config, msg: "server error", err: pgerr}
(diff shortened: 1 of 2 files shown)


jackc:
I don't think this is right. errTimeout can some sort of TCP timeout. It is entirely possible that one connection attempt to one host will timeout without the entire connection attempt to all hosts having timed out due to context cancellation or some other global timeout.

laskoviymishka:
It's not possible with current sequential logic, once we receive timeout all other fallback will receive same timeout error

pgconn/pgconn.go
```diff
@@ -193,6 +193,10 @@ func ConnectConfig(octx context.Context, config *Config) (pgConn *PgConn, err er
 			if _, ok := cerr.err.(*NotPreferredError); ok {
 				fallbackConfig = fc
 			}
+			if _, ok := cerr.err.(*errTimeout); ok {
+				// once we reach timeout it's useless to check other fallbacks
+				break
+			}
```
jackc, line 199: What I'm saying is it's not always useless to check other fallbacks. `errTimeout` could come from a TCP timeout. If the context has not expired the next fallback should be tried.

laskoviymishka: Got it.
I made one more attempt :D
So now I do a bit different thing, I try to keep track on fallback error and override result error only if needed.
For a deadline errors we should keep only first deadline error in result, every next one is false-positive errors (since we not actually do anything).

laskoviymishka:
This one is addressed here: https://github.com/jackc/pgx/commit/8db971660e43a944c777d5e4fcfcb7885acb7566

Issue #1195, opened 2022-04-26
Timestamps round trip incorrectly on non UTC systems


wbl:
When I am on a system not set to UTC, and store a timestamp to a database via the database/sql compatible interface, the value in the database is stored with the same numeric value but the zone set to UTC. When reading back the timestamp stored in the DB is correctly reconstituted, but alas this is not the same as the value that went in.

wbl:
Setting the location to UTC on the timestamp does work, so this is not really a function of the system timezone but the way timestamps are serialized.

jackc:
Is the database column a `timestamp` or `timestamptz`? Can you make a small reproduction case for this?

wbl:
It's timestamp. I can make a small reproduction sometime: been a bit busy lately.

frederikhors:
The same problem here (from https://github.com/ent/ent/issues/2692):

The column created in Postgres is:

```sql
used_at timestamp with time zone NOT NULL,
```

The value in Postgres is saved without timezone:

```
2022-06-30 22:49:03.970913+00
```

I'm using PG in Docker and using the PG query:

```sql
show timezone
```

I get:

```
Etc/UTC
```

I get from Ent (using pgx stdlib) the value:

```
2022-07-01T00:49:03.970913+02:00
```

This is a mess on the clients (where an UTC value is expected).

Using pgdriver/pq I get the UTC value from DB.

Am I wrong somewhere?

frederikhors:
I tried using this connection with this code too:

```
postgres://postgres:postgres@localhost/project?sslmode=disable&timezone=UTC"
```

```go
import (
	"database/sql"
	_ "github.com/jackc/pgx/v4/stdlib"
)

conn, err := sql.Open("pgx", dbPgConnURL+"&timezone=UTC")
//handle err
```

The problem is here.

I need a way to get back from DB the UTC values (that are laready stored in the DB).

frederikhors:
@jackc can you please help us here? 🙏

frederikhors:
I asked on SO too: https://stackoverflow.com/questions/72771272/how-to-setup-pgx-to-get-utc-values-from-db.

They suggest a custom type. More code, more issues!

@wbl did you find a way?

jackc:
Okay, I see what's going on here. The SO thread basically has the answer, but here's a little more of the explanation of why.

[...]

PR #1948, opened 2024-03-16
Add ScanLocation to Timestamp and Timestamptz

jackc:
This setting allows configuring the `time.Location` of `time.Time` values scanned from PostgreSQL `timestamp` and `timestamptz`.

For `timestamptz` it changes the display location but the instant in time is unchanged. That is, regardless of the value of `ScanLocation`, the results of `time.Time.Equal()` will be the same.

For `timestamp` it changes the location the time is assumed to have been in. That is, changing the value of `ScanLocation` will change the results of `time.Time.Equal()`.

### DIFF
--- a/pgtype/timestamp.go
+++ b/pgtype/timestamp.go
@@ -46,7 +46,7 @@ func (ts *Timestamp) Scan(src any) error {
 
 	switch src := src.(type) {
 	case string:
-		return scanPlanTextTimestampToTimestampScanner{}.Scan([]byte(src), ts)
+		return (&scanPlanTextTimestampToTimestampScanner{}).Scan([]byte(src), ts)
 	case time.Time:
 		*ts = Timestamp{Time: src, Valid: true}
 		return nil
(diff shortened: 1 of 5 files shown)


pgtype/timestamp.go
```diff
@@ -116,17 +116,21 @@ func (ts *Timestamp) UnmarshalJSON(b []byte) error {
...
 
-type TimestampCodec struct{}
+type TimestampCodec struct {
+	// ScanLocation is the location that the time is assumed to be in for scanning. This is different from
+	// TimestamptzCodec.ScanLocation in that this setting does change the instant in time that the timestamp represents.
+	ScanLocation *time.Location
+}
 
-func (TimestampCodec) FormatSupported(format int16) bool {
+func (*TimestampCodec) FormatSupported(format int16) bool {
```
alexrjones, line 125: Question: Is it strictly necessary for the new codec implementations to use a pointer receiver? In both cases the only member is the `*time.Location` field which is itself a pointer. As you mentioned in your commit descriptions, this is technically a breaking change, albeit a very minor one.

alexrjones (commented):
Hi @jackc, I saw you requested feedback on https://github.com/jackc/pgx/issues/1195#issuecomment-2002079265. Though I'm not experienced with pgx's internals, FWIW this change did address the issues we were facing in our application.

pgtype/timestamp.go
jackc: Is it strictly necessary? No. But as I understand it, it is usually better for interfaces implemented by structs to use methods that receive the pointer rather than the value.

fantapop:
Hi!  Is there a way to use the ScanLocation with pgx/v4?

jackc:
@fantapop Only if someone were to backport it.

PR #1949, opened 2024-03-17
Add more performant CollectRows functions

zolstein:
Add the functions CollectRowsUsing/Scanner et al. as more performant alternatives to CollectRows et al.

The new functions have the same names as the old one, suffixed with "Using" or "UsingScanner". The "UsingScanner" functions take a RowScannerInto object as an argument that provides similar functionaly to a RowScanner, but with some extensions and a generic type parameter.

The "Using" functions wrap the "UsingScanner" functions. They take a function as an argument that generates a RowScannerInto object. These functions have the same names as the "RowTo" functions, but with "RowTo" replaced with "RowInto". These "Using" functions provide a more-similar interface to the old functions, and guarantees about the lifetime of the RowScannerInto object - this can enable further optimizations like object pooling.

Ideally, the already-existing methods could have been modified, rather than duplicated. I couldn't find a way to do this without changing the type of RowToFunc, which would technically represent a breaking API change. The next major-versioned release should probably consolidate these methods.

Because CollectRows uses a RowToFunc to describe how to scan the row (and because these functions are totally opaque to the caller and return a value constructed internally) they inherently induce unnecessary allocations and copies. Additionally, struct types require some work to compute the mapping from columns to fields - this could be computed once and reused between rows, but the interface provides no easy way to do this and instead requires the work to be redone for each row.

The new interface allows the following optimizations:

* By passing a pointer receiver into ScanRowInto rather than returning the scanned result, we can serialize directly into the returned slice and avoid an additional allocation and copy for each value.
* By using a scanner object with state rather than just a function, we can reuse allocated memory for scannerTargets rather than reallocating for each row.
* By providing an Initialize method, we can pre-compute the mapping from column -> field for structs just once per query, rather than for each row.

### DIFF
--- a/row_scanner_into.go
+++ b/row_scanner_into.go
@@ -1,0 +1,573 @@
+package pgx
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/jackc/pgx/v5/pgconn"
+)
+
+// RowScannerInto provides an interface for scanning a row into a receiver value,
+// while doing type-checking and initialization only once per query.
+type RowScannerInto[T any] interface {
+	// Initialize sets up the RowScannerInto and validates it against the rows.
+	// Initialize must be called once before ScanRowInto.
+	Initialize(rows Rows) error
+	// ScanRowInto scans the row into the receiver.
+	ScanRowInto(receiver *T, rows Rows) error
+	// TODO: This could include a "Release" method, which could be used to pool
+	// scanner objects, and any internal objects they store, between queries.
+}
+
+// RowIntoSpec defines a specification for scanning rows into a given type.
+//
+// Note on the weird type definitions:
+// RowIntoSpec returns a struct containing a private function pointer because:
+//  1. We want to be able to manage the lifecycle of the returned value inside the
+//     collection functions. (E.g. we may decide to pool scanners for reuse.)
+//     In order to do this safely, we need to ensure the RowScannerInto returned by
+//     the inner function isn't referenced outside of the collecting function.
+//  2. Returning a struct allows us to extend this value in the future if necessary.
+//     By comparison, returning a function would not, and would require a (technically)
+//     breaking change if the type needed to change in the future.
+//  3. Returning a non-exported type lets us hide as many details as possible from
+//     the public API and restrict the only valid usage to:
+//     pgx.CollectRowsUsing(rows, RowInto[Type])
+//  4. RowIntoSpec is itself a function to provide a place to put the generic type
+//     parameter. rowIntoSpecRes cannot be a constant, since then there would be no
+//     place to put the type parameter. Since rowIntoSpecRes cannot be constructed in
+//     client code (by desing) it can't be applied when creating a struct value.
+type RowIntoSpec[T any] func() rowIntoSpecRes[T]
+
+type rowIntoSpecRes[T any] struct {
+	fn func() RowScannerInto[T]
+}
+
+// AppendRowsUsing iterates through rows, scanning each row according to into,
+// and appending the results into a slice of T.
+func AppendRowsUsing[T any, S ~[]T](slice S, rows Rows, into RowIntoSpec[T]) (S, error) {
+	return AppendRowsUsingScanner(slice, rows, into().fn())
+}
+
+// AppendRowsUsingScanner iterates through rows, scanning each row with the scanner,
+// and appending the results into a slice of T.
+func AppendRowsUsingScanner[T any, S ~[]T](
+	slice S,
+	rows Rows,
+	scanner RowScannerInto[T],
+) (s S, err error) {
+	defer rows.Close()
+
+	if err := scanner.Initialize(rows); err != nil {
+		return nil, err
+	}
+
+	startingLen := len(slice)
+	var startingPtr *T
+	if cap(slice) > 0 {
+		startingPtr = &slice[:cap(slice)][0]
+	}
+
+	defer func() {
+		// Named return values guarantee this err is the err that's actually returned.
+		if err != nil && len(slice) > startingLen && &slice[0] == startingPtr {
+			// An error occurred AND slice still has the same backing array as the input slice.
+			// Therefore, some written values are visible in the input slice. This could cause
+			// problems, especially if T contains pointers which are kept alive.
+			// To mitigate this, zero out the slice beyond the starting length.
+			for i := range slice[startingLen:] {
+				var zero T
+				slice[startingLen+i] = zero
+			}
+		}
+	}()
+
+	for rows.Next() {
+		i := len(slice)
+		var zero T
+		slice = append(slice, zero)
+		err := scanner.ScanRowInto(&slice[i], rows)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	return slice, nil
+}
+
+// CollectRowsUsing iterates through rows, scanning each row according to into,
+// and collecting the results into a slice of T.
+func CollectRowsUsing[T any](rows Rows, into RowIntoSpec[T]) ([]T, error) {
+	return CollectRowsUsingScanner(rows, into().fn())
+}
+
+// CollectRowsUsingScanner iterates through rows, scanning each row with the scanner,
+// and collecting the results into a slice of T.
+func CollectRowsUsingScanner[T any](rows Rows, scanner RowScannerInto[T]) ([]T, error) {
+	return AppendRowsUsingScanner([]T{}, rows, scanner)
+}
+
+// CollectOneRowUsing scans the first row in rows and returns the result. If no rows are found returns an error where errors.Is(ErrNoRows) is true.
+// CollectOneRowUsing is to CollectRowsUsing as QueryRow is to Query.
+func CollectOneRowUsing[T any](rows Rows, into RowIntoSpec[T]) (T, error) {
+	return CollectOneRowUsingScannner(rows, into().fn())
+}
+
+// CollectOneRowUsingScanner scans the first row in rows and returns the result. If no rows are found returns an error where errors.Is(ErrNoRows) is true.
+// CollectOneRowUsingScanner is to CollectRowsUsingScanner as QueryRow is to Query.
+func CollectOneRowUsingScannner[T any](rows Rows, scanner RowScannerInto[T]) (T, error) {
+	defer rows.Close()
+
+	var (
+		err   error
+		value T
+		zero  T
+	)
+
+	err = scanner.Initialize(rows)
+	if err != nil {
+		return zero, err
+	}
+
+	if !rows.Next() {
+		if err = rows.Err(); err != nil {
+			return zero, err
+		}
+		return zero, ErrNoRows
+	}
+
+	err = scanner.ScanRowInto(&value, rows)
+	if err != nil {
+		return zero, err
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return zero, err
+	}
+
+	return value, nil
+}
+
+// CollectExactlyOneRowUsing scans the first row in rows and returns the result.
+//   - If no rows are found returns an error where errors.Is(ErrNoRows) is true.
+//   - If more than 1 row is found returns an error where errors.Is(ErrTooManyRows) is true.
+func CollectExactlyOneRowUsing[T any](rows Rows, into RowIntoSpec[T]) (T, error) {
+	return CollectExactlyOneRowUsingScanner(rows, into().fn())
+}
+
+// CollectExactlyOneRowUsingScanner scans the first row in rows and returns the result.
+//   - If no rows are found returns an error where errors.Is(ErrNoRows) is true.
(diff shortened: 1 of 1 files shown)


jackc:
I added some benchmarks to help me understand this change. It's definitely faster, significantly faster.

```sh
jack@glados ~/dev/pgx ±zolstein-collect-rows-perf⚡ » got -run=^$ -bench=PR1949
goos: darwin
goarch: arm64
pkg: github.com/jackc/pgx/v5
BenchmarkPR1949AppendRowsAddrOfStruct-16           	    2160	    469062 ns/op
BenchmarkPR1949AppendRowsUsingAddrOfStruct-16      	    4170	    269547 ns/op
BenchmarkPR1949AppendRowsStruct-16                 	    2636	    464332 ns/op
BenchmarkPR1949AppendRowsUsingStruct-16            	    5186	    243114 ns/op
BenchmarkPR1949AppendRowsCustomScannerStruct-16    	    5310	    238816 ns/op
BenchmarkPR1949ManualScanAppendStruct-16           	    5314	    237211 ns/op
BenchmarkPR1949ManualScanIntoStruct-16             	    5545	    226621 ns/op
BenchmarkPR1949AppendRowsInt-16                    	    7552	    153954 ns/op
BenchmarkPR1949AppendRowsUsingInt-16               	    8491	    138888 ns/op
PASS
ok  	github.com/jackc/pgx/v5	11.379s
```

However, it's interesting to investigate exactly where the gains are coming from. Note these 5 benchmarks in particular. They all load structs into a slice. The only difference is how.

```
BenchmarkPR1949AppendRowsStruct-16                 	    2636	    464332 ns/op
BenchmarkPR1949AppendRowsUsingStruct-16            	    5186	    243114 ns/op
BenchmarkPR1949AppendRowsCustomScannerStruct-16    	    5310	    238816 ns/op
BenchmarkPR1949ManualScanAppendStruct-16           	    5314	    237211 ns/op
BenchmarkPR1949ManualScanIntoStruct-16             	    5545	    226621 ns/op
```

The bottom one, `BenchmarkPR1949ManualScanIntoStruct`,  gives a best case baseline. It manually uses `rows.Next()` and `rows.Scan()` and scans directly into the slice instead of using append. And the top one, `BenchmarkPR1949AppendRowsStruct`, is our current, inefficient approach. The second one, `BenchmarkPR1949AppendRowsUsingStruct`, is your new, faster approach.

However, what surprised me was  `BenchmarkPR1949AppendRowsCustomScannerStruct`. It uses the existing `pgx.AppendRows` with a custom scanner function instead of using `pgx.RowToStructByPos`.

```go
		records, err = pgx.AppendRows(records, rows, func(row pgx.CollectableRow) (Record, error) {
			var r Record
			err := row.Scan(&r.ID, &r.Name, &r.Age, &r.CreatedAt)
			return r, err
		})
```

This was slightly faster than `pgx.AppendRowsUsing` with `pgx.pgx.RowIntoStructByPos`.

This makes me think that the overwhelming portion of the improvement is from the scanning logic itself, not from the scanning directly into slice vs. append.

I further tested that with `BenchmarkPR1949ManualScanAppendStruct` and it appears to confirm that the scan directly into slice is only a very marginal improvement.

I tried to see what your new scanning objects do differently that makes such a difference, but I wasn't able to figure it out. But whatever it is, maybe it could be adapted to the existing `CollectRows` / `AppendRows` pattern. In particular, even though they take a scanner function, the function could be a method of an object. That could allow memoizing the expensive reflect operations.

```go
type Person struct {
	Name string
	Age  sql.NullInt64
}

type Record struct {
	ID int64
	Person
	CreatedAt time.Time
}

const createTablePeopleSQL = `create temporary table people (
	id bigint primary key generated by default as identity,
	name text not null,
	age bigint,
	created_at timestamptz not null
);

insert into people (name, age, created_at)
select gen_random_uuid()::text, ceil(random() * 100), clock_timestamp()
from generate_series(1, 1000);`

func BenchmarkPR1949AppendRowsAddrOfStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]*Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRows(records, rows, pgx.RowToAddrOfStructByPos[Record])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsUsingAddrOfStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]*Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRowsUsing(records, rows, pgx.RowIntoAddrOfStructByPos[Record])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRows(records, rows, pgx.RowToStructByPos[Record])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsUsingStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRowsUsing(records, rows, pgx.RowIntoStructByPos[Record])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsCustomScannerStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRows(records, rows, func(row pgx.CollectableRow) (Record, error) {
			var r Record
			err := row.Scan(&r.ID, &r.Name, &r.Age, &r.CreatedAt)
			return r, err
		})
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949ManualScanAppendStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}

		for rows.Next() {
			var r Record
			err = rows.Scan(&r.ID, &r.Name, &r.Age, &r.CreatedAt)
			if err != nil {
				b.Fatal(err)
			}
			records = append(records, r)
		}
		if rows.Err() != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949ManualScanIntoStruct(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	mustExec(b, conn, createTablePeopleSQL)
	records := make([]Record, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select id, name, age, created_at from people")
		if err != nil {
			b.Fatal(err)
		}

		for rows.Next() {
			if len(records) < cap(records) {
				records = records[:len(records)+1]
			} else {
				records = append(records, Record{})
			}
			r := &records[len(records)-1]
			err = rows.Scan(&r.ID, &r.Name, &r.Age, &r.CreatedAt)
			if err != nil {
				b.Fatal(err)
			}
		}
		if rows.Err() != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsInt(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	records := make([]int32, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select n::int4 from generate_series(1, 1000) n")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRows(records, rows, pgx.RowTo[int32])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}

func BenchmarkPR1949AppendRowsUsingInt(b *testing.B) {
	ctx := context.Background()
	conn := mustConnect(b, mustParseConfig(b, os.Getenv("PGX_TEST_DATABASE")))
	defer closeConn(b, conn)

	records := make([]int32, 0, 1000)

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		records = records[:0]
		rows, err := conn.Query(ctx, "select n::int4 from generate_series(1, 1000) n")
		if err != nil {
			b.Fatal(err)
		}
		records, err = pgx.AppendRowsUsing(records, rows, pgx.RowInto[int32])
		if err != nil {
			b.Fatal(err)
		}

		if len(records) != 1000 {
			b.Fatalf("expected 1000 records, got %d", len(records))
		}
	}
}
```

zolstein:
> I tried to see what your new scanning objects do differently that makes such a difference, but I wasn't able to figure it out.

The gist here is that at the start of the query it does introspection on the struct type and column fields to determine which columns correspond to which fields. Then, when actually scanning into the structs, it can just look up the addresses of the structs which is comparatively much cheaper, especially when mapping columns to struct is done by string comparison.

This part of the improvement could be refactored to work with the existing public API, though you'd lose some efficiency by virtue of (presumably) needing to look up the values from a synchronized map for each row. Also, when mapping by name, the map key would need to include both the struct type and a representation of the row fields, since you'd need a different mapping if the fields were in a different order and/or if only some fields were included.

> However, what surprised me was BenchmarkPR1949AppendRowsCustomScannerStruct. It uses the existing pgx.AppendRows with a custom scanner function instead of using pgx.RowToStructByPos.

This also surprises me. It might help to see benchmark output with allocations included. My understanding was that the main benefit from scanning directly into the struct was that it avoided the record getting heap-allocated for scanning. Maybe these versions are simple enough that the compiler can successfully run escape analysis? At that point, the difference should be small. I also wonder if these benchmarks run long enough for GC to have much effect.

zolstein:
I pushed out a PR that just caches the results of reflection and uses the more efficient scanning functions. My testing still shows it leaving some noticeable value on the table due to the extra allocations, but it's still an improvement. Curious what your benchmarks would show. (I couldn't manage to make the pgx test suite connect to my database.)

https://github.com/jackc/pgx/pull/1959

jackc:
Merged #1959 instead.

PR #1953, opened 2024-03-18
Add support for macaddr8 type

exekias:
Postgres also has a `macaddr8` type, this PR adds support for it, using the same codec as `macaddr`


jackc:
LGTM

PR #1960, opened 2024-03-25
Remove unnecessary Rows.Close()

bbaktaeho:
Hi, I frequently use the `CollectOneRow` function and noticed that Close() is redundantly called. There's no issue if it's already closed, but please remove it for cleaner code.


jackc:
That `Close()` is required.

The code above only reads a single row. If the result set had more than one row then `rows.Err()` would be returned before the deferred `rows.Close()`. This would be incorrect.

Issue #1950, opened 2024-03-17
Support MultiQueryTracer


uh-zz:
**Is your feature request related to a problem? Please describe.**

By using the `MultiQueryTracer` proposed in the following PR, I was able to set tracer and logger respectively.
I thought this is a case that other people might need.

- https://github.com/jackc/pgx/issues/1061

Do you plan to incorporate this into the pgx itself?

**Describe the solution you'd like**

I'd be happy to see it implemented in the pgx itself.

jackc:
> Do you plan to incorporate this into the pgx itself?

I'm open to something like that being merged, but I haven't seen any actual PRs.

One thing I would be concerned about is how to handle multiple tracers that each implement different hooks. I'm sure it could be handled but it might be a little tricky to both have a clean interface and avoid a performance hit.

PR #1976, opened 2024-03-31
feat: tracer supports MultiQueryTracer

uh-zz:
It able to set tracer and logger respectively.

https://github.com/jackc/pgx/issues/1950

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -25,6 +25,24 @@ type TraceQueryEndData struct {
 	Err        error
 }
 
+type MultiQueryTracer struct {
+	Tracers []QueryTracer
+}
+
+func (m *MultiQueryTracer) TraceQueryStart(ctx context.Context, conn *Conn, data TraceQueryStartData) context.Context {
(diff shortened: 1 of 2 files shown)


jackc:
As I mentioned on my comment to #1950:

> One thing I would be concerned about is how to handle multiple tracers that each implement different hooks. I'm sure it could be handled but it might be a little tricky to both have a clean interface and avoid a performance hit.

As far as I can tell this does not account for this case at all. What if one of tracers implemented `pgx.BatchTracer`?

rubiagatra:
Any update @uh-zz ?

uh-zz:
@rubiagatra 
Thank you for the mention. I haven't been able to work on this issue for a while. As @jackc san commented, I haven't considered the performance hit when supporting multiple tracers, so further verification may be necessary.

Until it is supported, the following issue might be helpful:
- https://github.com/jackc/pgx/issues/1061

rubiagatra:
Maybe I can help, I will take a look the following issue

PR #1982, opened 2024-04-07
Remove useless Close() calls

sylr:


batch.go
```diff
@@ -28,7 +28,6 @@ func (qq *QueuedQuery) Query(fn func(rows Rows) error) {
 		if err != nil {
 			return err
 		}
-		rows.Close()
```
jackc, line 31: This is not useless. There is no guarantee that fn() will have read all rows. In this case the `rows.Close()` is necessary for the rows to be closed before the error is read in `return rows.Err()`.

rows.go
```diff
@@ -461,7 +461,6 @@ func CollectOneRow[T any](rows Rows, fn RowToFunc[T]) (T, error) {
 		return value, err
 	}
 
-	rows.Close()
```
jackc, line 464: Especially required here as only one row is read.

PR #1992, opened 2024-04-18
remove keepalive and rely on GOLANG default (since go 1.13 default is 15s)

h0nIg:
set keepalive to appropiate value (since go 1.13 default is 15s).

5 minute for initial keepalive checks are way to high, we should orient us here according to go defaults

https://www.reddit.com/r/golang/comments/d7v7dn/psa_go_113_introduces_15_sec_server_tcp/

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -800,7 +800,8 @@ func parsePort(s string) (uint16, error) {
 }
 
 func makeDefaultDialer() *net.Dialer {
-	return &net.Dialer{KeepAlive: 5 * time.Minute}
+	// rely on GOLANG KeepAlive settings
+	return &net.Dialer{}
 }
 
 func makeDefaultResolver() *net.Resolver {


paudley:
This may make sense for TCP connections in general, but it probably doesn't for PostgreSQL connections, where the setup and teardown are expensive enough to want longer keep-alives.

paudley:
> @paudley "paudley reacted with thumbs down emoji" why?

Sorry, was a mistake!  (undone)   I'm doing some testing on the effects of shorter keepalives in our environments but I don't have results yet.

jackc:
That keep alive value goes all the way back to b46ee0a9512cb0a375e3bcd0c4fbc6c05bdf3843 in 2014. At the time Go did not have TCP keepalive enabled by default.

If a change is to be made here, I believe the correct change would be to remove the 5 minute value and actually use the Go default rather than hard code it to the current Go default.

h0nIg:
@jackc done

pgconn/config.go
```diff
@@ -800,7 +800,8 @@ func parsePort(s string) (uint16, error) {
 }
 
 func makeDefaultDialer() *net.Dialer {
-	return &net.Dialer{KeepAlive: 5 * time.Minute}
+	// rely on GOLANG KeepAlive settings
+	return &net.Dialer
```
jackc, line 804: That's not valid syntax.

pgconn/config.go
```diff
@@ -800,7 +800,8 @@ func parsePort(s string) (uint16, error) {
 }
 
 func makeDefaultDialer() *net.Dialer {
-	return &net.Dialer{KeepAlive: 5 * time.Minute}
+	// rely on GOLANG KeepAlive settings
+	return net.Dialer
```
jackc, line 804: This still doesn't compile. Please try running the tests locally.

jackc (changes requested):

Issue #1566, opened 2023-04-04
Implementations of driver.Valuer are no longer called for nil receivers in v5


alnr:
The [v5 changelog](https://github.com/jackc/pgx/blob/master/CHANGELOG.md#null-representation) mentions this:

> Previously, a type that implemented driver.Valuer would have the Value method called even on a nil pointer. All nils whether typed or untyped now represent NULL.

That's a huge change in behavior from v4. Previously, it was very simple to implement custom types which would serialize themselves into a non-SQL-NULL value from a nil receiver.

```sql
CREATE TABLE tab (col text NOT NULL);
```

```go
type MySlice []string

func (m MySlice) Value() (driver.Value, error) {
	if len(m) == 0 {
		return "[]", nil
	}

	encoded, err := json.Marshal(&m)
	return string(encoded), errors.WithStack(err)
}

func main() {
	var m MySlice
	_, err := conn.Exec(context.Background(), "INSERT INTO tab (col) VALUES ($1)", m))
}
```

This would work in v4, but fails in v5 with a NOT NULL constraint violation, because `(MySlice).Value()` is never called.

Is there a comparably simple way to achieve this pattern in pgx/v5?

Thanks for your work and this great library!

glerchundi:
It seems like your concerns were partially answered and resolved in subsequent minor and patch releases, right? At least ours were with the latest 5.3.1.

jackc:
Sorry, I actually don't know any good way for an application to get that behavior back.

Typed nils caused a lot of headaches in pgtype and normalizing them at the border simplified things quite a bit. It wasn't clear to me at the time that this would have negative downstream effects. While calling methods with a nil receiver is allowed in Go, it almost always is an error. But I suppose it is a reasonable thing to consider with your example type.

As far as changes to pgx that could improve this use case, there are only two options that come to mind.

1. A new interface that when implemented signals that typed nils should not be normalized.
2. A flag that turns off the nil normalization.

Not sure either of these is a good idea though. I'm not confident that any typed nils allowed past the border wouldn't crash later in code not expecting a nil. Maybe if it was only enabled for database/sql though...

aeneasr:
> It seems like your concerns were partially answered and resolved in subsequent minor and patch releases, right? At least ours were with the latest 5.3.1.

The patch used in gobuffalo/pop does use `github.com/jackc/pgx/v5 v5.3.1` so I think the problem persists.

[...]

PR #1993, opened 2024-04-22
add option to handle nil Valuers

sajanalexander:
This is a fix for #1566 to add an option to handle cases where we've passed a value that is considered `nil` by `anynil`, but may have a value which can be substituted, like in the case where something implements the `Valuer` interface and returns a default value. 

Note that I needed to add an option to `pgx.Conn` to be able to modify the `conn.TypeMap()`.  I couldn't figure out another way to do this when using the stdlib.

Please let me know if this is on the right track and/or needs any modifications.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,6 +17,8 @@ import (
 	"github.com/jackc/pgx/v5/pgtype"
 )
 
+type AfterPgxConnectFunc func(ctx context.Context, pgxconn *Conn) error
+
 // ConnConfig contains all the options used to establish a connection. It must be created by ParseConfig and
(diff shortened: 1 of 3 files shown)


jackc:
pgx supports for `driver.Valuer` in its native interface. Fixing this issue only for stdlib / database/sql would mean that `driver.Valuer` would behave differently depending on whether the type was used with pgx directly or through database/sql.

So I don't think this approach will work. However, it was still valuable. It got me thinking about this problem again and I think I found a simpler solution than was discussed in #1566. See PR #2019 .

PR #1996, opened 2024-04-29
make QueuedQuery.Fn property public, closes #1878

pashagolub:
This commit fixes the overlook of the #1886 where `QueuedQuery.SQL` and `QueuedQuery.Arguments` properties were exposed.

And if for `SQL` and `Arguments` we had an opportunity to use reflection, for `QueuedQuery.fn` it's impossible.

One more benefit exposing `QueuedQuery.Fn` for developers is the ability to execute batched functions without knowing the kind of a queued query and mandatory calling `BatchResults.Close()`, e.g.

```go
batch := &pgx.Batch{}
batch.Queue("select func42($1)", 42).QueryRow(...)
batch.Queue("select * from foo").Query(...)
batch.Queue("insert").Exec(...)
...
batch.Queue("truncate foo").Exec(...)

br = conn.SendBatch(ctx, batch)
// check first two queued queries
err := batch.QueuedQueries[0].Fn()
err = errors.Join(err, batch.QueuedQueries[1].Fn())
// let BatchResults execute all that left and close 
err = errors.Join(err, br.Close())
if err != nil {
	fmt.Printf("SendBatch error: %v", err)
	return
}
```

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -12,7 +12,7 @@ import (
 type QueuedQuery struct {
 	SQL       string
 	Arguments []any
-	fn        batchItemFunc
+	Fn        batchItemFunc
 	sd        *pgconn.StatementDescription
 }
 
@@ -20,7 +20,7 @@ type batchItemFunc func(br BatchResults) error
 
 // Query sets fn to be called when the response to qq is received.
 func (qq *QueuedQuery) Query(fn func(rows Rows) error) {
-	qq.fn = func(br BatchResults) error {
+	qq.Fn = func(br BatchResults) error {
 		rows, _ := br.Query()
 		defer rows.Close()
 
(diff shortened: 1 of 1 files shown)


pashagolub:
When this PR will be merged, one will be able to test Batches with pgxmock:

```go
func TestBatch(t *testing.T) {
	t.Parallel()
	mock, _ := NewConn()
	a := assert.New(t)

	// define our expectations
	eb := mock.ExpectBatch()
	eb.ExpectQuery("select").WillReturnRows(NewRows([]string{"sum"}).AddRow(2))
	eb.ExpectExec("update").WithArgs(true, 1).WillReturnResult(NewResult("UPDATE", 1))

	// run the test
	batch := &pgx.Batch{}
	batch.Queue("select 1 + 1").QueryRow(func(row pgx.Row) error {
		var n int32
		return row.Scan(&n)
	})
	batch.Queue("update users set active = $1 where id = $2", true, 1).Exec(func(ct pgconn.CommandTag) (err error) {
		if ct.RowsAffected() != 1 {
			err = errors.New("expected 1 row to be affected")
		}
		return
	})

	err := mock.SendBatch(ctx, batch).Close()
	a.NoError(err)
	a.NoError(mock.ExpectationsWereMet())
}
```

Issue #1747, opened 2023-09-26
add the Acquire connection tracer


taran507:
Hello, I found that pgpool does not track the time for connection allocation. I often have to deal with the fact that the request takes a long time to work out precisely because of the long receipt of the connection. I suggest adding AcquireTracer, which will help detect a similar problem
```
package pgx

// AcquireTracer - traces Acquire.
type AcquireTracer interface {
	// TraceAcquireStart is called at the beginning of Acquire.
	// The returned context is used for the rest of the call and will be passed to the TraceAcquireEnd.
	TraceAcquireStart(ctx context.Context, data TraceAcquireStartData) context.Context
	TraceAcquireEnd(ctx context.Context, data TraceAcquireEndData)
}

type TraceAcquireStartData struct {
	ConnConfig *pgconn.Config
}

type TraceAcquireEndData struct {
	Err error
}
```

Add the acquireTracer variable to the Pool structure
```
type Pool struct {
	// ...
	acquireTracer pgx.AcquireTracer
	// ...
}
```

Declare it in the same way as other tracers, but only when creating pgxpool
```
func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
	// Default values are set in ParseConfig. Enforce initial creation by ParseConfig rather than setting defaults from
	// zero values.
	if !config.createdByParseConfig {
		panic("config must be created by ParseConfig")
	}

	p := &Pool{
		config:                config,
		beforeConnect:         config.BeforeConnect,
		afterConnect:          config.AfterConnect,
		beforeAcquire:         config.BeforeAcquire,
		afterRelease:          config.AfterRelease,
		beforeClose:           config.BeforeClose,
		minConns:              config.MinConns,
		maxConns:              config.MaxConns,
		maxConnLifetime:       config.MaxConnLifetime,
		maxConnLifetimeJitter: config.MaxConnLifetimeJitter,
		maxConnIdleTime:       config.MaxConnIdleTime,
		healthCheckPeriod:     config.HealthCheckPeriod,
		healthCheckChan:       make(chan struct{}, 1),
		closeChan:             make(chan struct{}),
	}

	if t, ok := config.ConnConfig.Tracer.(pgx.AcquireTracer); ok {
		p.acquireTracer = t
	}

```
And call tracer methods at the beginning and end of connection allocation
```
func (p *Pool) Acquire(ctx context.Context) (*Conn, error) {
	if p.acquireTracer != nil {
[...]

PR #2008, opened 2024-05-10
Add pgx pool acquire tracer

ngavinsir:
resolves https://github.com/jackc/pgx/issues/1747

### DIFF
--- a/pgxpool/conn.go
+++ b/pgxpool/conn.go
@@ -13,6 +13,7 @@ import (
 type Conn struct {
 	res *puddle.Resource[*connResource]
 	p   *Pool
+	ctx context.Context
 }
 
 // Release returns c to the pool it was acquired from. Once Release has been called, other methods must not be called.
@@ -27,7 +28,7 @@ func (c *Conn) Release() {
 	c.res = nil
 
 	if c.p.releaseTracer != nil {
-		ctx := c.p.releaseTracer.TraceReleaseStart(context.Background(), c.p, TraceReleaseStartData{Conn: conn})
+		ctx := c.p.releaseTracer.TraceReleaseStart(c.ctx, c.p, TraceReleaseStartData{Conn: conn})
 		defer func() {
 			c.p.releaseTracer.TraceReleaseEnd(ctx, c.p, TraceReleaseEndData{})
 		}()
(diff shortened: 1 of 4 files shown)


jackc:
Looks pretty good, but I have a few suggestions.

I think that `TraceAcquireStart` and `TraceAcquireEnd` should receive the `*Pool` as an argument. This would match the precedent set by `QueryTracer` of passing the item that is being traced. This would also mean that `TraceAcquireStartData` would not need to include `ConnConfig`. `TraceAcquireStartData` would still exist for possible future changes, but it would be empty.

It's a little confusing at first whether `TraceAcquireEnd` is called when the connection is acquired or when the acquisition is complete (i.e. when it is released). I think the docs should make that clearer. `TraceAcquireEndData` should probably also  include the `*pgx.Conn`.

---

This might be a separate feature, but it might also make sense to have a `ReleaseTracer` interface to allow tracing the entire time the the connection is acquired. In addition to the normal metrics usage of tracing, this would allow building diagnostics that would make it easier to track down connection leaks.

ngavinsir:
@jackc updated, but not sure how to pass the current context to the release method, so it won't have the same parent trace as the acquire one

jackc:
@ngavinsir I guess the release tracer is a bit trickier than I thought.

It is, perhaps, a little weird to use the context from Acquire for the Release. But I suppose context is where things like opentelemetry are stored so it is necessary. 🤔 At least it should be documented that it is the Acquire Context and/or made part of `TraceRelease(Start|End)Data`. Storing it in the `pgxpool.Conn` is fine. It's private so we can refactor it if necessary.

I also wonder if release tracing needs Start and End. Release should be consistently extremely fast and non-blocking. Perhaps it only needs one hook?

jackc:
🤔 And on even further reflection, maybe the release tracer doesn't need context at all. As of 6f0deff0156a7ffcd557eac1011ebb5ce73739d3 there is a custom data map available on connections. Any release tracer that needed it could simply store it there.

ngavinsir:
@jackc I see, so for example someone can save the passed context on TraceAcquireEnd in the TraceAcquireEndData.Conn.PgConn().CustomData() and access it again inside trace release from TraceReleaseData.Conn.PgConn().CustomData()

jackc:
Thanks!

PR #2011, opened 2024-05-13
Add support for custom JSON marshal and unmarshal

### LINKED ISSUE #2005: Custom JSON marshal/unmarshal

I have two use cases where I think custom function to do JSON marshal/unmarshal would help (instead of default `json.Marshal`/`json.Unmarshal`:

* I want to marshal with `SetEscapeHTML(false)` set, because there is no point in escaping HTML when storing into a database. Similarly, I want to set `KnownFields(true)` when unmarshal, to detect errors early.
* I want to store a polymorphic JSON into a database (one where depending on its `type` field it gets unmarshal into different structs), the type being an interface. So I would use a custom unmarshal function which first checks the type and then does the rest.

Some easy way to specify the default `json.Marshal` and `json.Unmarshal` function to be used.

It seems I could remove the json codec and create my own copy of it, with only difference being that it does not call `json.Marshal`. I would like to avoid having to maintain such a copy where the only difference is in the function which gets called.

mitar:
Fixes #2005.


pgtype/json_test.go
```diff
@@ -224,3 +225,20 @@ func TestJSONCodecEncodeJSONMarshalerThatCanBeWrapped(t *testing.T) {
...
+			Name: "json", OID: pgtype.JSONOID, Codec: &pgtype.JSONCodec{
+				Marshal: func(v any) ([]byte, error) {
+					return []byte(`{"custom":"value"}`), nil
+				},
+				Unmarshal: json.Unmarshal,
+			}})
+	}
+
+	pgxtest.RunValueRoundTripTests(context.Background(), t, connTestRunner, pgxtest.KnownOIDQueryExecModes, "json", []pgxtest.ValueRoundTripTest{
+		{[]byte(`"hello"`), new(string), isExpectedEq(`{"custom":"value"}`)},
```
mitar, line 242: @jackc I do not know how to write this test. This does not work. It seems like Marshal is simply not called? Any insight? Maybe I should not use `RunValueRoundTripTests`?

jackc:
`[]byte` is sent directly without being marshaled.

```go
func (c JSONCodec) PlanEncode(m *Map, oid uint32, format int16, value any) EncodePlan {
	switch value.(type) {
	case string:
		return encodePlanJSONCodecEitherFormatString{}
	case []byte:
		return encodePlanJSONCodecEitherFormatByteSlice{}
```

You probably want to test something like a `map[string]any`.

mitar:
@jackc Thanks! This is ready for review now I think.

jackc:
LGTM

Issue #1860, opened 2024-01-02
Valuer interface is not respected for nil values


SajanAlexander-okta:
**Describe the bug**
When using a struct that implements `driver.Valuer` but is backed by a nilable type, `Value` is never called when the value is `nil`.  This is used by some types where a default should be stored within the DB.  The [default converter](https://github.com/golang/go/blob/master/src/database/sql/driver/types.go#L237) for `database/sql` will call this on a nil.  

I'm using the stdlib compatibility layer.  When tracing through the code, the `database/sql` code will call the [CheckNamedValue](https://github.com/jackc/pgx/blob/master/stdlib/sql.go#L539) method on `*stdlib.Conn`, which will then just pass through the values directly.

Later on, when the prepared statement is being built, `anynil.NormalizeSlice` is [called](https://github.com/jackc/pgx/blob/master/extended_query_builder.go#L26) on the arguments, which then sets the value of the arg as nil and erases any type information about the arg.  

The nil value hits a [nil check](https://github.com/jackc/pgx/blob/master/extended_query_builder.go#L116), preventing any Encode plan from being executed to convert the value / call `Value` on it.  

**To Reproduce**
See code below.  The first insert will result in a `NULL` value to be inserted, while the second one will insert an empty object of `{}`. 

```go
package main

import (
	"database/sql"
	"database/sql/driver"
	"encoding/json"
	"fmt"
	"os"

	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/stdlib"
)

// JSONText code was copied from github.com/jmoiron/sqlx types.JSONText so there's not another import required outside of pgx and the Go stdlib
type JSONText json.RawMessage

var emptyJSON = JSONText("{}")

func (j JSONText) Value() (driver.Value, error) {
	var m json.RawMessage
	var err = j.Unmarshal(&m)
	if err != nil {
		return []byte{}, err
	}
	return []byte(j), nil
}

func (j *JSONText) Unmarshal(v interface{}) error {
	if len(*j) == 0 {
		*j = emptyJSON
	}
	return json.Unmarshal([]byte(*j), v)
}

func main() {
	connConfig, err := pgx.ParseConfig(os.Getenv("PGX_TEST_DATABASE"))
	if err != nil {
		fmt.Println("error parsing config: ", err)
		return
	}

	dn := stdlib.RegisterConnConfig(connConfig)

	db, err := sql.Open("pgx", dn)
	if err != nil {
[...]

PR #2019, opened 2024-05-18
Fix encode driver.Valuer on pointer

jackc:
pgx v5 introduced nil normalization for typed nils. This means that []byte(nil) is normalized to nil at the edge of the encoding system. This simplified encoding logic as nil could be encoded as NULL and type specific handling was unneeded.

However, database/sql compatibility requires Value to be called on a nil pointer that implements driver.Valuer. This was broken by normalizing to nil.

This commit changes the normalization logic to not normalize pointers that directly implement driver.Valuer to nil. It still normalizes pointers that implement driver.Valuer through implicit derefence.

e.g.

```go
type T struct{}

func (t *T) Value() (driver.Value, error) {
  return nil, nil
}

type S struct{}

func (s S) Value() (driver.Value, error) {
  return nil, nil
}
```

`(*T)(nil)` will not be normalized to nil but `(*S)(nil)` will be.

https://github.com/jackc/pgx/issues/1566


internal/anynil/anynil.go
```diff
@@ -1,17 +1,47 @@
...
-	switch refVal.Kind() {
+	kind := refVal.Kind()
+	switch kind {
 	case reflect.Chan, reflect.Func, reflect.Map, reflect.Ptr, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
-		return refVal.IsNil()
+		if !refVal.IsNil() {
+			return false
+		}
+
+		if kind == reflect.Ptr {
```
sajanalexander, line 33: For my use-case, the restriction to just pointers won't work since the type I'm using is an alias to a slice.  See https://github.com/jackc/pgx/issues/1860 for some example code.  

Allowing `reflect.Slice` here would fix my use-case, but it would probably make sense to allow for others as well.

jackc: Good point. It now extends the same logic to all nilable values.

This extends the behavior change a little further than I had anticipated. But it is a fix for database/sql compatibility and I don't anticipate the change breaking any real world code.

sajanalexander (approved):
LGTM.   Thanks for fixing this!

jackc:
I just refactored the whole typed nil system to exist entirely in `pgtype.Map.Encode`. This change keeps the original `nil` `driver.Valuer` available while still treating other typed nils as SQL NULL.

If my design and implementation is correct, this approach should restore full driver.Valuer support while still not leaking any typed nils into the Codec system.

sajanalexander:
I like the refactor.  Tracing how nils were handled prior to the change was a bit complicated and this definitely simplifies it.

PR #2028, opened 2024-05-25
Enable composite tests on cockroachdb

jackc:


rubiagatra:
Want to continue it @jackc ?

Issue #2030, opened 2024-05-26
Provide improved support for loading types


nicois:
I am using pgx with a database which defines a very large number of custom types. Virtually every stored procedure returns a custom type, and any given application needs at least 50-100 types, often more.
The current `LoadType` connection method is helpful, but is limited:
- `LoadType` generates two queries to the database when one would be sufficient.
- it can only be given one type at a time, and then requiring at least one round-trip to the database to complete. Coupled with the impracticality of running multiple `LoadType` calls concurrently, this adds significant overhead to an application's startup, before the dataclass connection/pool can be used at all.
- it is painful to have to manually calculate the types which will be required in the application. It is not only necessary to specify the desired type, but to recursively calculate all the other types these depend on, and to load them in, in the correct order.

I would like to create a PR to resolve these issues, either in separate PRs or together, if I get the green light. Are these seen as valid concerns, and would improvements in these areas be accepted?

To briefly describe what I am doing locally, and which would form the basis of my PR(s):
- (the first point above is quite trivial: the first `QueryRow` calls made in `LoadType` is used in the second query)
- add a new method named `LoadTypes` which takes a `[]string` of type names. Modify the (now) single SQL query to use `= ANY($1)` to find all OIDs matching the provided type names, and return `[]*pgtype.Type`.
- add a helper method `GetTypeDependencies` which, given a connection and a list of type names, recursively identifies the sub-types they implicitly also need registered. This can either be called internally by `LoadType` and the new `LoadTypes` or if that is considered too "magicial", can be kept separate for the user to invoke. e.g. `conn.LoadTypes(ctx, conn.GetTypeDependencies(ctx, "my_type1", "my_type_2", ...))`
 
The SQL I am using to assist in the above:
```sql
WITH RECURSIVE typeinfo AS (
    SELECT c.relname AS parent, regexp_replace(pg_catalog.format_type(a.atttypid, a.atttypmod), '\[\]$', '') as child, 'f'::bool AS has_array
    FROM pg_catalog.pg_class c
        LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
        JOIN pg_catalog.pg_attribute a ON (a.attrelid = c.oid)
        JOIN pg_type ON (pg_type.typname = regexp_replace(pg_catalog.format_type(a.atttypid, a.atttypmod), '\[\]$', ''))
    WHERE pg_catalog.pg_table_is_visible(c.oid)
[...]

PR #2046, opened 2024-06-15
Faster/easier loading of types

nicois:
As discussed in https://github.com/jackc/pgx/issues/2030 , improve the `LoadType` method to use a single SQL call.

Add `LoadTypes()` which adds multiple types in a single call, and supports recursively loading of other implicitly-required types in the same SQL call.

### DIFF
--- a/derived_types.go
+++ b/derived_types.go
@@ -1,0 +1,260 @@
+package pgx
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+/*
+buildLoadDerivedTypesSQL generates the correct query for retrieving type information.
+
+	pgVersion: the major version of the PostgreSQL server
+	typeNames: the names of the types to load. If nil, load all types.
+*/
+func buildLoadDerivedTypesSQL(pgVersion int64, typeNames []string) string {
+	supportsMultirange := (pgVersion >= 14)
+	var typeNamesClause string
+
+	if typeNames == nil {
+		// This should not occur; this will not return any types
+		typeNamesClause = "= ''"
+	} else {
+		typeNamesClause = "= ANY($1)"
+	}
+	parts := make([]string, 0, 10)
+
+	// Each of the type names provided might be found in pg_class or pg_type.
+	// Additionally, it may or may not include a schema portion.
+	parts = append(parts, `
+WITH RECURSIVE
+-- find the OIDs in pg_class which match one of the provided type names
+selected_classes(oid,reltype) AS (
+    -- this query uses the namespace search path, so will match type names without a schema prefix
+    SELECT pg_class.oid, pg_class.reltype
+    FROM pg_catalog.pg_class
+        LEFT JOIN pg_catalog.pg_namespace n ON n.oid = pg_class.relnamespace
+    WHERE pg_catalog.pg_table_is_visible(pg_class.oid)
+      AND relname `, typeNamesClause, `
+UNION ALL
+    -- this query will only match type names which include the schema prefix
+    SELECT pg_class.oid, pg_class.reltype
+    FROM pg_class
+    INNER JOIN pg_namespace ON (pg_class.relnamespace = pg_namespace.oid)
+    WHERE nspname || '.' || relname `, typeNamesClause, `
+),
+selected_types(oid) AS (
+    -- collect the OIDs from pg_types which correspond to the selected classes
+    SELECT reltype AS oid
+    FROM selected_classes
+UNION ALL
+    -- as well as any other type names which match our criteria
+    SELECT pg_type.oid
+    FROM pg_type
+    LEFT OUTER JOIN pg_namespace ON (pg_type.typnamespace = pg_namespace.oid)
+    WHERE typname `, typeNamesClause, `
+        OR nspname || '.' || typname `, typeNamesClause, `
+),
+-- this builds a parent/child mapping of objects, allowing us to know
+-- all the child (ie: dependent) types that a parent (type) requires
+-- As can be seen, there are 3 ways this can occur (the last of which
+-- is due to being a composite class, where the composite fields are children)
+pc(parent, child) AS (
+    SELECT parent.oid, parent.typelem
+    FROM pg_type parent
+    WHERE parent.typtype = 'b' AND parent.typelem != 0
+UNION ALL
+    SELECT parent.oid, parent.typbasetype
+    FROM pg_type parent
+    WHERE parent.typtypmod = -1 AND parent.typbasetype != 0
+UNION ALL
+    SELECT pg_type.oid, atttypid
+    FROM pg_attribute
+    INNER JOIN pg_class ON (pg_class.oid = pg_attribute.attrelid)
+    INNER JOIN pg_type ON (pg_type.oid = pg_class.reltype)
+    WHERE NOT attisdropped
+      AND attnum > 0
+),
+-- Now construct a recursive query which includes a 'depth' element.
+-- This is used to ensure that the "youngest" children are registered before
+-- their parents.
+relationships(parent, child, depth) AS (
+    SELECT DISTINCT 0::OID, selected_types.oid, 0
+    FROM selected_types
+UNION ALL
+    SELECT pg_type.oid AS parent, pg_attribute.atttypid AS child, 1
+    FROM selected_classes c
+    inner join pg_type ON (c.reltype = pg_type.oid)
+    inner join pg_attribute on (c.oid = pg_attribute.attrelid)
+UNION ALL
+    SELECT pc.parent, pc.child, relationships.depth + 1
+    FROM pc
+    INNER JOIN relationships ON (pc.parent = relationships.child)
+),
+-- composite fields need to be encapsulated as a couple of arrays to provide the required information for registration
+composite AS (
+    SELECT pg_type.oid, ARRAY_AGG(attname ORDER BY attnum) AS attnames, ARRAY_AGG(atttypid ORDER BY ATTNUM) AS atttypids
+    FROM pg_attribute
+    INNER JOIN pg_class ON (pg_class.oid = pg_attribute.attrelid)
+    INNER JOIN pg_type ON (pg_type.oid = pg_class.reltype)
+    WHERE NOT attisdropped
+      AND attnum > 0
+    GROUP BY pg_type.oid
+)
+-- Bring together this information, showing all the information which might possibly be required
+-- to complete the registration, applying filters to only show the items which relate to the selected
+-- types/classes.
+SELECT typname,
+       pg_namespace.nspname,
+       typtype,
+       typbasetype,
+       typelem,
+       pg_type.oid,`)
+	if supportsMultirange {
+		parts = append(parts, `
+       COALESCE(multirange.rngtypid, 0) AS rngtypid,`)
+	} else {
+		parts = append(parts, `
+       0 AS rngtypid,`)
+	}
+	parts = append(parts, `
+       COALESCE(pg_range.rngsubtype, 0) AS rngsubtype,
+       attnames, atttypids
+    FROM relationships
+    INNER JOIN pg_type ON (pg_type.oid = relationships.child)
+    LEFT OUTER JOIN pg_range ON (pg_type.oid = pg_range.rngtypid)`)
+	if supportsMultirange {
+		parts = append(parts, `
+    LEFT OUTER JOIN pg_range multirange ON (pg_type.oid = multirange.rngmultitypid)`)
+	}
+
+	parts = append(parts, `
+    LEFT OUTER JOIN composite USING (oid)
+    LEFT OUTER JOIN pg_namespace ON (pg_type.typnamespace = pg_namespace.oid)
+    WHERE NOT (typtype = 'b' AND typelem = 0)`)
+	parts = append(parts, `
+    GROUP BY typname, pg_namespace.nspname, typtype, typbasetype, typelem, pg_type.oid, pg_range.rngsubtype,`)
+	if supportsMultirange {
+		parts = append(parts, `
+        multirange.rngtypid,`)
+	}
+	parts = append(parts, `
+        attnames, atttypids
+    ORDER BY MAX(depth) desc, typname;`)
+	return strings.Join(parts, "")
+}
+
+type derivedTypeInfo struct {
+	Oid, Typbasetype, Typelem, Rngsubtype, Rngtypid uint32
+	TypeName, Typtype, NspName                      string
+	Attnames                                        []string
+	Atttypids                                       []uint32
+}
+
+// LoadTypes performs a single (complex) query, returning all the required
+// information to register the named types, as well as any other types directly
+// or indirectly required to complete the registration.
+// The result of this call can be passed into RegisterTypes to complete the process.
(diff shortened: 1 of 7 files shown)


nicois:
@jackc when tests run against cockroachdb, I see:
```
ERROR: unknown signature: array_agg(name[]) (SQLSTATE 42883)
```

I don't have any experience with this database, and I don't see any mention of being incompatible with this function. If pgx needs to support cockroachdb, how would you suggest we do so? I could try using an alternate syntax, but I'd like your advice before doing so.

nicois:
All regression tests apart from cockroachdb are passing: https://github.com/nicois/pgx/actions/runs/9526586068

jackc:
Regarding Cockroach DB, if it is something that is not supported by that database engine then there is a helper method to skip tests. But as far as I know, `array_agg` is supported by CRDB. I'm not sure where that error is coming from but the `name[]` is suspicious. I would expect the call to aggregate `name` not `name[]`. As in, it looks like it is trying to aggregate an array of `name` into an array of arrays of `name`s instead of aggregating `name` into an array of `name`.

I also was wondering if the loading all custom types includes table types.

nicois:
`array_agg` is being used by the new single-SQL statement, to select everything in one go - so it's not viable to simply skip the test.
This is related to supporting composite fields, where I am collecting multiple 2-tuples of composite fields into an array of 2-tuples, so it is in fact intended that this is an array of arrays of strings. 
I will try to flatten the structure to a single array of 2`X` elements, where `X` is the number of composite fields.

Regarding the loading of all types: yes, there is further work needed there. I've added a simple test and there are complications. I think this needs some experimentation/discussion, and it's probably better to cover that in its own PR. I will remove the public method to invoke the loading of all types from this PR.

nicois:
It's passing with cockroachdb now. I've also reinstated the original code for `LoadType` to minimise the chance of causing any regressions for existing users.
When loading multiple types, it's pretty much essential to load each type into the map as it's found, so I've also renamed the methods to more clearly capture this fact.

nicois:
I will refine the SQL further, to deal with more complex heirachies of types/domains/arrays/composite types.

jackc:
I suppose there is no way around the bit recursive query. But, wow, that is scary looking. I'd hate to have to debug that. 🤷 

I also think the server version should be an internal function.

nicois:
Sure, I'll make it internal. I thought it might be helpful for a client, but perhaps it's better to not expose it.

That query isn't nice, and it's quite possible it could be optimised in some way too; it takes longer than I'd like it to. However, my upcoming pgxpool PR has support for caching this mapping (for non-replicated environments), meaning the overhead is only felt once per invocation.

nicois:
I'll add a few comments to the scary SQL to make it a little less nightmarish.

nicois:
I've changed the `serverVersion` to being private, added some comments to the SQL, and also modified the structure a little, so there is a `LoadTypes` and `RegisterTypes` call. This is both more similar to the `LoadType` which exists, and will be a better structure for what's coming.

jackc:
Hmmm... The functionality looks good, but I'm not so sure about the interface.

At first glance, I didn't like `TypeInfo`. I didn't understand why `pgtype.Type` wouldn't be used directly. But then I realized that since types dependent on other types directly refer to the registered `Codec` of those types it wouldn't be possible to safely share the type. Be that as it may, it's still confusing to have two types that are so similar. Not sure what the solution is.

Then `LoadTypes` and `RegisterTypes` are named such that I would expect they are exactly the same as the singular versions except they do multiple at a time. But they are different from their singular counterparts. It is counterintuitive that `LoadType` returns a `*pgtype.Type` but `LoadTypes` returns `[]*TypeInfo`.

I think I understand at this point why `*pgtype.Type` couldn't be used directly, but the naming, at least, is very unfortunate.

Upon further reflection, I also question if `LoadTypes` and `RegisterTypes` should be members of `*pgx.Conn` at all. `LoadTypes` doesn't use any private values or functions of the `*pgx.Conn` except `serverVersion` and that doesn't need to be a member either. `RegisterTypes` has even less justification to be a member function. Not only does it not use the receiver it actually uses the `registerWith` argument instead.

---

So here's some suggestions for your consideration. I'm not sure if they are good ideas, but I think they are worth looking into.

* Rename `LoadTypes` to disambiguate loading a `pgtype.Type` and `pgx.TypeInfo`.
* Rename `RegisterTypes` to disambiguate registering a `pgtype.Type` and `pgx.TypeInfo`.
* Make `LoadTypes` a function instead of a method.
* Make `RegisterTypes` a function instead of a method. Or perhaps it should be a method on `*pgtype.Map`.
* Perhaps `TypeInfo` should be part of the `pgtype` instead of the `pgx` package.
* I wish there was a overall naming scheme. Perhaps something to do with derived or dependent types would be better. `LoadType(s)` only works on types that are derived from known types. e.g. You can't use `LoadType` for an entirely unknown type like something from PostGIS. But you do use `pgtype.Map.RegisterType`. Maybe there should be something like `LoadDerivedTypes` / `RegisterDerivedType` / `DerivedTypeInfo`. 🤷‍♂️ 
* Possibly `Conn.LoadType` should be deprecated in favor of this new functionality.
 
---

Also, one other random thing. I noticed `LoadTypes` is purposely using the simple protocol. Any particular reason?

nicois:
I've addressed the feedback.

- Renamed methods and struct as suggested
- Turned methods into functions where possible
- Moved some things into pgtype
- Made the registration part into a method on `Map`
- Added tests, including showing how arrays work: https://github.com/jackc/pgx/pull/2046/files#diff-af46ecc2001b2a06a9990fee3299e7e2cd95e99cbbb914cd7fd90dc1d2c186d4R33

There are a couple of reasons I didn't change the original `LoadType`:
- when only a single type is being loaded, in some situations it might be faster, if the database connection is low latency and there are many other types which would be scanned by the mega query
- people who want speed will switch over to the new syntax, probably via the (coming) pgxpool PR
- if there are any bugs, they can be ironed out without upsetting existing users

nicois:
Regarding the use of the simple protocol: when I remove this, at least when running locally the test fails:

```
While scanning type information: can't scan into dest[1]: cannot scan char (OID 18) in binary format into *string
```

jackc:
Regarding scanning `"char"`, that is a special type. I believe the issue is scanning `pg_type.typtype` into a string. Try scanning it into a `byte` or `rune` instead.

---

I like the progress that is being made here. But I've been reviewing this PR, looking back at the existing pgx type code, and mulling it over. And I think there might be more fundamental improvements that can be made.

As I understand it, you are trying to improve two issues. 1. This PR is making loading types faster. 2. The pgxpool PR is trying to make it unnecessary to introspect the database for each connection. They are somewhat related and improvements in one may affect what is desirable or necessary for the other.

Regarding loading types faster, as we've discussed above, I consider it unfortunate that we end up with 2 similar things `pgtype.Type` and `pgtype.DerivedTypeInfo`. The naming, at least, is better now, but two similar things still bother me. That's makes me wonder what if instead of adding `pgtype.DerivedTypeInfo` we instead added all of its fields to `pgtype.Type`? `LoadDerivedTypes` would again become `LoadTypes`.

... 🤔 ... 🤔 ... 🤔 ...

Actually, wait a second. Why do we need the `DerivedTypeInfo` fields at all?

My previous observation was wrong:

> At first glance, I didn't like TypeInfo. I didn't understand why pgtype.Type wouldn't be used directly. But then I realized that since types dependent on other types directly refer to the registered Codec of those types it wouldn't be possible to safely share the type. 

It **is** possible to safely share the `*pgtype.Type`. According to the docs `*pgtype.Type` and `*pgtype.Codec` must not be altered once they are registered with a `pgtype.Map`.

I think this means PR this and the pgxpool PR can be significantly simplified. We still need the complicated SQL query, but I think much of the rest could go away.

Here's an idea: We add a `LoadTypes` function. It runs your big SQL query, but it returns `[]*pgtype.Type`. It's slightly tricky because it needs to build up its own map of types in addition to the `*pgtype.Map. For example, `_mytype` depends on `mytype` which is a composite which contains `_myothertype`. The `*pgtype.Type`s would need to be created in the order `myothertype`, `_myothertype`, `mytype`, `_mytype`. So `LoadTypes` would have to keep track of what `*pgtype.Type` it has created.

`RegisterDerivedTypes` goes away. The existing `*Map.RegisterType` function is used.

Add another small change and we also solve the `_hstore` problem discussed on the other PR. `LoadTypes` could support types where typtype = 'b' (base type). It would leave the `Codec` nil in that case. Derived types like `_hstore` would still be initialized because they would have a pointer to the `pgtype.Type` for `hstore`. The caller of `LoadTypes` would be responsible for filling in the `Codec` before registering the types.

This should allow for all type introspection to be done in a single SQL query regardless of how many types or what kind types are used.

Regarding the pgxpool change to avoid having to introspect the database for each connection, I'm not sure how much of that would still be necessary either. Since `pgtype.Type` and `pgtype.Codec` are immutable once registered, there's no reason we can't just copy one `pgtype.Map`'s type information to another `pgtype.Map`.

nicois:
Ok, I see where you're going with this. I will try to avoid the need to define a second type, if I can. 

Regarding pgxpool: the type map has to be populated for each connection, doesn't it? And there is still the need to support situations where the pool does not always connect to servers sharing the same OID mapping. Assuming that, there are wins in making it easy to let the custom and automatic loaders be defined more easily, and to cache the results when it's safe to do so.

Either way, it's good to keep the PRs separate as the first one is needed either way, and once it's solid, it's easier to look at the incremental benefit of the other.

nicois:
I've made some changes which didn't take very long at all, and am pretty happy with the results:
- there is now `LoadTypes` which returns `[]*pgtype.Type`. It copies the connection's `Map` and then mutates that copy while it builds the list of types, preventing types from complaining about not being able to find other types which they expect to already be registered. But this is just an temporary internal map; what is returned is a clean list of `Type`s. Note that this also deals with the `hstore` issue: as long as someone has registered `hstore` in their connection's `Map` before calling `LoadTypes`, that `hstore` type will be found and respected.
- I created a `Map.Copy()` method to support the above
- A `Map.RegisterTypes()` convenience method is trivial, and can be removed if you think it's too trivial.

nicois:
> Regarding scanning "char", that is a special type. I believe the issue is scanning pg_type.typtype into a string. Try scanning it into a byte or rune instead.

I've tried removing the simple protocol designator, then using `byte` and `rune`, and they both error. I'll leave it as-is for now, unless you can suggest something else.

jackc:
I mentioned this idea on #2056, but I will restate it here.

What if `pgx.ConnConfig` had a list of PG type names to load when the connection is established. For base types like `hstore` or PostGIS types, a map of name to function could be provided that would contain the loading logic.

This would remove the need to manually load types after connection, and would probably remove most uses of the after connect hooks provided by pgxpool and stdlib.

It would also remove the need to have any new functions or methods. The only new interface would be additional fields on `*pgx.ConnConfig`.

This would be an even simpler interface for loading types.

nicois:
@jackc I have tried to modify these 3 PRs to minimise the change in the interface, as you suggested. It did work out a bit less messy than I anticipated, so let me know what you think.

jackc:
@nicois 

Thanks for making all these changes. I still think this configuration should not live on the pool. But now actually seeing it on the connection and especially seeing how it complicates #2048 (`ConnectConfigReusingTypeMap` is a rather inelegant necessity caused by this change), I'm not sure `AutoLoadTypes` on the `pgx.ConnConfig` is right either.

Here's what I think we should do. Let's go back to some of your original design.

* Back out this last change of `AutoLoadTypes` on `pgx.ConnConfig`.
* Make `LoadTypes` a method of `*pgx.Conn` again. Given that it now really is the collection version of `LoadType` it should live next to it. If I could do it all over again I might not make `LoadType` a `Conn` method, but at this point consistency is probably the priority.

I'll merge just that core functionality. Then we can evaluate what is the most convenient interface for exposing the functionality.

---

With regard to that convenient interface, the pgxpool interface and possibly configuration on `Conn`, I would like us to evaluate using https://pkg.go.dev/sync#OnceValue to load the types. This, possibly with a wrapping helper function, might be a better abstraction than the config options.

nicois:
I've pushed the earlier version back, with the duplicate creation of types for the namespaced version of each type's name. You can more easily see this here: https://github.com/jackc/pgx/compare/4660f1d7546adb2f6091625ed9fe42f393daf522..dcb32f94021386bf4f450b858281687faa4a0641

derived_types.go
```diff
@@ -0,0 +1,261 @@
...
+buildLoadDerivedTypesSQL generates the correct query for retrieving type information.
+
+	pgVersion: the major version of the PostgreSQL server
+	typeNames: the names of the types to load. If nil, load all types.
+*/
+func buildLoadDerivedTypesSQL(pgVersion int64, typeNames []string) string {
+	supportsMultirange := (pgVersion >= 14)
+	var typeNamesClause string
+
+	if typeNames == nil {
```
jackc, line 23: Is this branch still valid? As far as I can see the only caller, LoadTypes, enforces len(typeNames) > 1.

derived_types.go
```diff
@@ -0,0 +1,261 @@
...
+// or indirectly required to complete the registration.
+// The result of this call can be passed into RegisterTypes to complete the process.
+func (c *Conn) LoadTypes(ctx context.Context, typeNames []string) ([]*pgtype.Type, error) {
+	m := c.TypeMap()
+	if typeNames == nil || len(typeNames) == 0 {
+		return nil, fmt.Errorf("No type names were supplied.")
+	}
+
+	serverVersion, err := serverVersion(c)
+	if err != nil {
```
jackc, line 167: Should serverVersion really be able to error? Or rather, do we care? As far as I can see, the server version is only used to determine whether multirange is supported? Falling back to no multirange might improve support for non-PostgreSQL servers.

It still may fail on non-PostgreSQL servers later, but it would probably be an error that more directly indicated the underlying issue.

derived_types.go
```diff
@@ -0,0 +1,261 @@
...
+			if !ok {
+				return nil, fmt.Errorf("Array element OID %v not registered while loading pgtype %q", ti.Typelem, ti.TypeName)
+			}
+			type_ = &pgtype.Type{Name: ti.TypeName, OID: ti.Oid, Codec: &pgtype.ArrayCodec{ElementType: dt}}
+		case "c": // composite
+			var fields []pgtype.CompositeCodecField
+			for i, fieldName := range ti.Attnames {
+				//if fieldOID64, err = strconv.ParseUint(composite_fields[i+1], 10, 32); err != nil {
+				//	return nil, fmt.Errorf("While extracting OID used in composite field: %w", err)
+				//}
```
jackc, line 201: Should this be removed?

jackc (approved):
I made a few notes above in the review, but none of them are critical. We also might want to add a test for the namespaced type path.

But I think this is mergeable at this point. Just let me know if you are planning on making any more changes.

nicois:
I'll address your feedback above, then let you know. We can follow up further if I need be.

derived_types.go
nicois: I would prefer this branch to be kept, as the alternative is to fail at runtime if a `nil` is passed in. It is an unexercised branch right now, and the compiler will optimise it away. 
If there were a way to make it a compile-time error, if a `nil` was passed in, it would be even better.

nicois: I have modified this branch to run, but not match any records. This shouldn't usually happen, but at least avoids a runtime error.

nicois:
ok, this should be good to merge now.

PR #2048, opened 2024-06-17
Simplify type autoloading with pgxpool

nicois:
Provide some backwards-compatible configuration options for pgxpool
which streamlines the use of the bulk loading of custom types:
- AutoLoadTypes: a list of type (or class) names to automatically
  load for each connection, automatically also loading any other
  types these depend on.
- ReuseTypeMaps: if enabled, pgxpool will cache the typemap information,
  avoiding the need to perform any further queries as new connections
  are created.

ReuseTypeMaps is disabled by default as in some situations, a
connection string might resolve to a pool of servers which do not share
the same type name -> OID mapping.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -12,14 +12,17 @@ import (
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/jackc/puddle/v2"
 )
 
-var defaultMaxConns = int32(4)
-var defaultMinConns = int32(0)
-var defaultMaxConnLifetime = time.Hour
-var defaultMaxConnIdleTime = time.Minute * 30
-var defaultHealthCheckPeriod = time.Minute
+var (
+	defaultMaxConns          = int32(4)
+	defaultMinConns          = int32(0)
+	defaultMaxConnLifetime   = time.Hour
+	defaultMaxConnIdleTime   = time.Minute * 30
+	defaultHealthCheckPeriod = time.Minute
+)
 
 type connResource struct {
 	conn       *pgx.Conn
@@ -100,6 +103,48 @@ type Pool struct {
 
 	closeOnce sync.Once
 	closeChan chan struct{}
+}
+
+// LoadTypesAfterConnect is suitable for assigning to the AfterConnect configuration setting.
+// It will automatically load the named types for each connection in an efficient manner,
+// performing a single query to the database backend. The underlying call to pgx.LoadTypes
+// is smart enough to also retrieve any related types required to support the definition of the
+// named types.
+// If reuseTypeMap is enabled, it is assumed that the OID mapping is stable across all database
+// backends in this pool, resulting in only needing to query when creating the initial connection;
+// subsequent connections will reuse the same OID type mapping.
+// Because it is not always possible for a client to know the database topology in the final usage
+// context, PGXPOOL_REUSE_TYPEMAP, when given a value of y or n, will take precedence over this argument.
+func LoadTypesAfterConnect(typeNames []string, reuseTypeMap bool) func(context.Context, *pgx.Conn) error {
+	if reuseTypeMap {
+		mutex := new(sync.Mutex)
+		var types []*pgtype.Type
+		return func(ctx context.Context, conn *pgx.Conn) error {
+			mutex.Lock()
+			defer mutex.Unlock()
+			var err error
+
+			if types != nil {
+				conn.TypeMap().RegisterTypes(types)
+				return nil
+			}
+			types, err = conn.LoadTypes(ctx, typeNames)
+			if err != nil {
+				types = nil
+				return err
+			}
+			conn.TypeMap().RegisterTypes(types)
+			return nil
+		}
+	}
+	return func(ctx context.Context, conn *pgx.Conn) error {
+		types, err := conn.LoadTypes(ctx, typeNames)
+		if err != nil {
+			return err
+		}
+		conn.TypeMap().RegisterTypes(types)
+		return nil
+	}
 }
 
 // Config is the configuration struct for creating a pool. It must be created by [ParseConfig] and then it can be


nicois:
A recommended way of using this would be something like:
```
        config.AutoLoadTypes = []string{"my_type", "another_type"}
        config.ReuseTypeMaps = true
        config.MinConns = 10
```
Which, compared to previously, would:
- make a single SQL call as soon as the pgxpool was created (as `MinConns` > 0, the pool will immediately begin filling)
- the type map information will be loaded, without needing to worry about explicitly naming other custom types; they will be automatically detected
- subsequent connections will not need to perform any SQL queries to load custom types while being initialised

Together this should result in a significant improvement both to startup times and whenever new connections are being added to the connection pool, whenever custom types are being registered.

jackc:
I think I'd like to wait and see what happens with #2046 before looking too much into this. But there are a few things that come to mind.


1. How does `AutoLoadTypes` work with non-derived types. e.g. I have to manually register `hstore`. How do I register `_hstore`? A `[]string` of types to load and register is really convenient, but I'm not sure it would work in every case.
2. That might affect how the types are shared in pgxpool. 
3. If these can be resolved perhaps some of this logic could / should be pushed down into something usable from

nicois:
Agreed, let's get the other PR sorted before finessing this PR.

Related array types should be handled by the new loader. I haven't experimented with this in depth, as I haven't run into any problems with the code, but if there is an example/test where array types aren't supported, I can remedy that, probably in the other PR.

jackc:
> Related array types should be handled by the new loader.

Right, I expect it would work - once the underlying type is loaded. But I wasn't sure if the `AutoLoadTypes` would be handled first, leaving it unusable in those cases.

nicois:
> > Related array types should be handled by the new loader.
> 
> Right, I expect it would work - once the underlying type is loaded. But I wasn't sure if the `AutoLoadTypes` would be handled first, leaving it unusable in those cases.

If it works correctly, [this line](https://github.com/jackc/pgx/pull/2046/commits/ca18af49ba56c7b00fbbd1e9d1a0913b333b0318#diff-4f427d2b022907c552328e63f137561f6de92396d7a6e8f6c2ea1bcf0db52654R1453) should collect any dependent array types. I can look at adding a test in the other PR to explicitly ensure this works.

jackc:
What I mean is, what happens when it tries to load `_hstore` when `hstore` is not registered yet?

nicois:
As `_hstore` depends on `hstore`, requesting to load the array would implicitly load the underlying data type first. You can see this in the original PR's tests; this behaviour is not specific to the autoloader.

nicois:
Thinking a bit more about manual registration of certain types, and how to reduce the number of queries there: what do you think of adding something like a `map[string] func(c conn.Conn, oid uint32) error` field to the pgxpool config. Then if they want to register say `hstore` on each connection, they can assign their custom registration function to this map against the `"hstore"` key, and the pgxpool will retrieve the OIDs for each of these types, then call the functions with that information? 
Not only would this be cleaner, but if the `reuse...` configuration boolean was set, pgxpool will be able to internally cache the type name -> OID mapping, and also only need to make a database call when the first connection is created.

nicois:
I have rebased this on top of the other PR, and added a comment relating to the autoload config settings to clarify that custom type registration done in `AfterConnect` will be respected.
Discussion about further optimising the loading of the custom registration can be talked about if/when this PR is merged.

nicois:
Thinking about https://pkg.go.dev/sync#OnceValue : I can see how this can replace the use of a mutex. Were you thinking that the pgxpool's struct would contain a function used to register the types? And that if the reuse flag was set, that function would be replaced by the `OnceValue` version of it?

The primary question is whether the user provides the autoload information via the pgxpool's config, or via some other mechanism.

jackc:
> Were you thinking that the pgxpool's struct would contain a function used to register the types?

Yes, something like this. Though since pgxpool already has a AfterConnect hook it might just be using AfterConnect in a specific way.

> And that if the reuse flag was set, that function would be replaced by the OnceValue version of it?

The reuse flag wouldn't have to exist. The function would use OnceValue internally. From pgxpool's point of view, it has a function it calls to get (or get and register) the type information. This technique could be documented, or maybe even a new standalone function that encapsulates this concept and returns a type loading function that internally uses OnceValue.

Potentially, this lets the user control autoload and type reuse without pgxpool needing any explicit support.

nicois:
Without reuse being optional, how would heterogeneous servers be supported, where oids could differ?

Whether it's via a config setting or environment variable, if this is a possible situation, we should support it, shouldn't we?

If not, I certainly agree that it would be nice to eliminate the need to support such a workflow. This was based on what you suggested earlier.

jackc:
Reuse would still be optional. It would be up to the caller to implement.

Here is what I'm thinking: 

```go
	makeLoadTypesOnce := func() func(ctx context.Context, conn *pgx.Conn) error {
		var mux sync.Mutex
		var loaded bool
		var types []*pgtype.Type

		return func(ctx context.Context, conn *pgx.Conn) error {
			mux.Lock()
			defer mux.Unlock()

			if loaded {
				return types, nil
			}

			var err error
			types, err = conn.LoadTypes(ctx, "mytype", "myothertype")
			if err != nil {
				return err
			}

			loaded = true
			return nil
		}
	}

	dbconfig.AfterConnect = makeLoadTypesOnce()
```

I haven't actually tested it, but I think it should work. I wasn't able to figure out a way to get `sync.OnceValue` to work as it needs the `ctx` and `conn` from the first call. But the `sync.Mutex` works fine.

Also, prototyping this code made me even more in favor of this logic being activated through the existing `AfterConnect` hook. The project that I was working in had this `AfterConnect` hook already installed:

```go
	dbconfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
		pgxuuid.Register(conn.TypeMap())
		return nil
	}
```

It's using `github.com/jackc/pgx-gofrs-uuid` to integrate with `github.com/gofrs/uuid`. It is important that this type be registered before the any composites that have a UUID field or the wrong UUID type could be used.

It could be modified as follows to load types once after registering the custom UUID type:

```go
	loadTypesOnce := makeLoadTypesOnce()
	dbconfig.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
		pgxuuid.Register(conn.TypeMap())
		err := loadTypesOnce(ctx, conn)
		if err != nil {
			return err
		}
		return nil
	}
```

---

Not sure if we would want a function that creates / simplifies the load once logic or not. It would be more convenient. But I can imagine different applications having slightly different preferences so it might be easier to document the pattern instead.

nicois:
We can make this something where each user needs to write/adapt their AfterConnect to do this. I just thought that type registration was such a common pattern that we could ensure it would done in a consistent way through providing helpers to manage the autoloading, and to keep AfterConnect free for anything special they want done apart from type registration.
The other benefit to `reuse` being a configuration option is that it is more likely to be exposed to end-users via a DSN, so if when all they have is a precompiled binary and the author didn't consider supporting the reuse of type registration, those end-users can still opt in.

nicois:
@jackc I have made a helper which can be used to build an `AfterCommit` hook. I have looked at the uuid code yet, but wanted to share what I've got now, as it's already fairly compact.

I think this strikes a reasonable balance, where this helper can keep things sufficiently DRY while still allowing people to decide whether to chain this with other `AfterConnect` logic.

I'm also interested in what you think of the environment variable. I do think that since we aren't using the pgxpool config for this any more, it wouldn't be possible for a DSN to control the reuse of type information, so an optional environment variable seems like a good way to let this setting be controlled by end users.

nicois:
If the approach shown here is one you are happy with, it might actually make sense for me to integrate this with https://github.com/jackc/pgx/pull/2049 , as the helper function's signature will change when support for custom registration functions is added.

pgxpool/pool.go
```diff
@@ -102,6 +106,60 @@ type Pool struct {
...
+	}
+	if reuseTypeMap {
+		mutex := new(sync.Mutex)
+		var types []*pgtype.Type
+		return func(ctx context.Context, conn *pgx.Conn) error {
+			if types != nil {
+				// avoid acquiring the mutex if the types are already available
+				conn.TypeMap().RegisterTypes(types)
+				return nil
+			}
```
jackc, line 134: I'm pretty sure this is not safe. The read on line 130 could happen concurrently with the write on line 144. I don't think there is any way around some sort of synchronization.

But also, this only happens when creating a connection. Any performance gains would be insignificant compared to the complexity beyond doing the simplest locking.

pgxpool/pool.go
```diff
@@ -102,6 +106,60 @@ type Pool struct {
...
+// It will automatically load the named types for each connection in an efficient manner,
+// performing a single query to the database backend. The underlying call to pgx.LoadTypes
+// is smart enough to also retrieve any related types required to support the definition of the
+// named types.
+// If reuseTypeMap is enabled, it is assumed that the OID mapping is stable across all database
+// backends in this pool, resulting in only needing to query when creating the initial connection;
+// subsequent connections will reuse the same OID type mapping.
+// Because it is not always possible for a client to know the database topology in the final usage
+// context, PGXPOOL_REUSE_TYPEMAP, when given a value of y or n, will take precedence over this argument.
+func AutoloadAfterConnect(typeNames []string, reuseTypeMap bool) func(context.Context, *pgx.Conn) error {
```
jackc, line 119: Not sure about the name. Maybe `LoadTypesAfterConnect`?

jackc (commented):
I'm actually not in favor of the environment variable for two reasons.

1. The caller could read an environment variable if they wanted to. This could be as simple as `AutoloadAfterConnect([]string{"mytype", "myothertype"}, os.Getenv("PGXPOOL_REUSE_TYPEMAP") == "y")`. I don't think we gain anything by having it internal to this function.
2. pgx's use of environment variables is limited to when creating / parsing the connection / pool config. An environment variable is a global variable, and isolating their usage to a single point makes the code easier to reason about.

---

> If the approach shown here is one you are happy with, it might actually make sense for me to integrate this with https://github.com/jackc/pgx/discussions/2049 , as the helper function's signature will change when support for custom registration functions is added.

I do think we will want to wait to merge until we also have figured out how to handle custom registration for types like `hstore` and PostGIS types.

I think the custom registration functions are worth looking into.

I also think it's worth considering letting `LoadTypes` load base types and have it leave the `Codec` empty. The caller could populate the `Codec` for base types. This would allow a single query to suffice for all type registration.

It also might be worth considering doing nothing. There are relatively few base types that will be used by any given project relative to all the derived types. A few extra queries might be inconsequential relative to the complexity of optimizing them away.

pgxpool/pool.go
nicois: I think this is safe. Whatever is happening in the mutex, either this variable will be set to `nil` (if there is nothing yet holding the mutex, or if that goroutine hasn't yet updated the variable), or it will be set to its final value.
If it's `nil`, we then acquire the mutex and proceed safely. But there is no need to acquire the mutex if the value has already been set.
The reason this is safe is that the transition is only in one direction, from `nil` to its final value.
I can't find any documentation indicating that a goroutine will get a runtime error or similar; just that the value could be changed "underneath it", which in the situation is fine, as it would result in the next step being to acquire the mutex and re-verify the value.
The purpose of this check before acquiring the mutex is to cater for the common case when the types are already known, allowing new connections to instantly access the types without needing to wait for any other connections also in the process of running this code.

Here's a silly little gist showing how there is no problem reading a shared variable without a mutex, as long as you don't care about timing: https://gist.github.com/nicois/d4aabaa6fc1c0ed998a1e4cc9bd102f5#file-no-mutex-go

pgxpool/pool.go
nicois: sure, I'll rename this.

nicois:
The reason for wanting the environment variable is because I expect there are a significant number of projects which use `pgx`. Many of them will not bother exposing the reuse setting, either disabling it for safety, or enabling it as 99% of users will benefit from the extra speed, particularly with large numbers of connections.
Because the DSN doesn't provide the ability to control this boolean (and as it's not part of ConnConfig, but only defined in a bundled function), an environment variable would avoid the situation where people are trying to use an pgx-based application and are otherwise unable to access this setting. 
This is the time when we have the opportunity to define a consistent method to control this flag. But I can remove it if you don't like it; this is your project.

pgxpool/pool.go
jackc: Try running it with the race detector:

```
jack@glados ~/dev/gorun » go run -race main.go
==================
WARNING: DATA RACE
Read at 0x000100606028 by goroutine 10:
  main.reader()
      /Users/jack/dev/gorun/main.go:38 +0xc0
  main.main.gowrap2()
      /Users/jack/dev/gorun/main.go:54 +0x4c

Previous write at 0x000100606028 by goroutine 6:
  main.writer()
      /Users/jack/dev/gorun/main.go:21 +0x94
  main.main.gowrap1()
      /Users/jack/dev/gorun/main.go:51 +0x4c

Goroutine 10 (running) created at:
  main.main()
      /Users/jack/dev/gorun/main.go:54 +0x150

Goroutine 6 (running) created at:
  main.main()
      /Users/jack/dev/gorun/main.go:51 +0x12c
==================
Read 238100 values
Read 246484 values
Read 239879 values
Read 237489 values
Read 238356 values
Read 230082 values
Written 10056 values
Read 242796 values
Read 231918 values
Read 234697 values
Read 266612 values
nothing unexpected
Found 1 data race(s)
exit status 66
```

jackc:
I'd prefer to hold off on the environment variable. We can always add it later, but once it's there we can't remove it.

nicois:
OK, I've removed the environment variable reference.

jackc:
@nicois

There is still the race condition mentioned above.

But there is one other thing that needs to be figured out that was brought to my attention by #2089. `EnumCodec` has mutable state. The `Type` and `Codec` interfaces require that they are not modified after registration, but I suppose it is unclear whether they should be able to modify themselves. `EnumCodec` modifies itself, so it would not be safe to share between multiple connections. Not sure what the best solution is here.

nicois:
I'll take a look at that race condition. I didn't notice your comment, sorry.

Relating to state mutation: if types and codecs shouldn't be mutated, should they have a `Copy()` member which produces an equivalent object with a separate internal state? If so, we could then iterate over these items using this method to generate copies for the new connection.
Potentially `Copy()` could return a pointer to itself, if the type/codec is one which is safe to reuse.

jackc:
> Relating to state mutation: if types and codecs shouldn't be mutated, should they have a Copy() member which produces an equivalent object with a separate internal state? If so, we could then iterate over these items using this method to generate copies for the new connection.

Seems reasonable.

> Potentially Copy() could return a pointer to itself, if the type/codec is one which is safe to reuse.

I would make it an optional additional interface a `Codec` could also implement. So it would be assumed safe to use concurrently if it was not implemented.

pgxpool/pool.go
```diff
@@ -12,14 +12,17 @@ import (
...
+	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/jackc/puddle/v2"
 )
 
-var defaultMaxConns = int32(4)
-var defaultMinConns = int32(0)
-var defaultMaxConnLifetime = time.Hour
-var defaultMaxConnIdleTime = time.Minute * 30
-var defaultHealthCheckPeriod = time.Minute
+var (
```
Syed-Tajamul, line 19: galat ha chee ye

PR #2061, opened 2024-06-25
Add support for sslrootcert=system

yann-soubeyrand:
Hello,

libpq supports using system certificate authorities to verify the server TLS certificate: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT. This PR tries to add compatible behaviour to pgx.

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -654,6 +654,36 @@ func configTLS(settings map[string]string, thisHost string, parseConfigOptions P
 
 	tlsConfig := &tls.Config{}
 
+	if sslrootcert != "" {
+		var caCertPool *x509.CertPool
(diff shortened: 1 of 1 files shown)


yann-soubeyrand:
Hello, yes, I didn’t know how to add tests for that (setting up an environment on my workstation to make tests seemed a little bit complicated for the time I had unfortunately…). I guess we can have a test similar to the test for `sslrootcert=./path/to/ca-cert.crt`, but with a step to install the CA certificate in the system pool?

jackc:
Well, I think I will just merge it. The code looks right. And I don't see any good way to test it either.

PR #2065, opened 2024-06-29
Add support for PG 17 interval infinity values

toofishes:
Although Postgres 17 is still in beta, I thought it might make sense to get ahead of one of the changes being made.

Added in this commit: https://github.com/postgres/postgres/commit/519fc1bd9
Upstream release notes: https://www.postgresql.org/docs/17/release-17.html#RELEASE-17-DATATYPES

The wire format for infinity/-infinity adds no breaking changes- it is just the various values set to either minimum or maximum possible int64/int32 values.

I attempted to match how InfinityModifier works in other types, such as date and timestamptz.

Please let me know what, if anything, I forgot to update here, and any other feedback is definitely welcome.

It feels like this is relatively safe and backward compatible with older versions, since you can just choose not to use infinity values with older versions, and since there were no wire format changes. However, I'm happy to make changes or documentation updates if that is necessary.

### DIFF
--- a/pgtype/interval.go
+++ b/pgtype/interval.go
@@ -4,6 +4,7 @@ import (
 	"database/sql/driver"
 	"encoding/binary"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 
@@ -27,10 +28,11 @@ type IntervalValuer interface {
 }
 
 type Interval struct {
-	Microseconds int64
-	Days         int32
-	Months       int32
-	Valid        bool
+	Microseconds     int64
+	Days             int32
+	Months           int32
+	InfinityModifier InfinityModifier
+	Valid            bool
 }
 
 func (interval *Interval) ScanInterval(v Interval) error {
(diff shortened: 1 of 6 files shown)


pgtype/interval.go
```diff
@@ -27,10 +28,11 @@ type IntervalValuer interface {
...
 
 type Interval struct {
-	Microseconds int64
-	Days         int32
-	Months       int32
-	Valid        bool
+	Microseconds     int64
+	Days             int32
+	Months           int32
+	InfinityModifier InfinityModifier
```
toofishes, line 34: Is it worth adding a field comment here along the lines of this?

```golang
// InfinityModifier for intervals is only supported in Postgres 17 or newer
InfinityModifier InfinityModifier
```

jackc: Wouldn't hurt.

jackc (approved):

PR #2066, opened 2024-06-30
Fix: use `slices.Contains` according to the TODO comment

yuki0920:
I used `slices.Contains` according to the TODO comment.

```
// TODO replace by slices.Contains when experimental package will be merged to stdlib
// https://pkg.go.dev/golang.org/x/exp/slices#Contains
```


jackc:
Thanks. I also updated the go.mod file to represent that pgx required Go 1.21+.

PR #2082, opened 2024-07-10
don't print URL when url.Parse returns an error

heavycrystal:
When `pgx.Connect`or `pgx.ParseConfig` is used, it indirectly calls url.Parse which by default returns the URL as part of its error type in case it failed to parse it. This leads to the error message from pgx leaking the unredacted URL in any logs.

Fixed by typecasting to `url.Error` and then returning only the appropriate bit of the error, needed to change a variable name along with this.

## Minimal Reproduction

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/jackc/pgx/v5"
)

func main() {
	conn, err := pgx.Connect(context.Background(), "postgres://postgres:password@ localhost:5432/postgres")
	if err != nil {
		os.Exit(1)
	}
}
```

### Without the change
```
Unable to connect to database: cannot parse `postgres://postgres:xxxxxx@ localhost:5432/postgres`: failed to parse as URL (parse "postgres://postgres:password@ localhost:5432/postgres": invalid character " " in host name)
```

### With the change
```
Unable to connect to database: cannot parse `postgres://postgres:xxxxxx@ localhost:5432/postgres`: failed to parse as URL (invalid character " " in host name)
```

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -467,14 +467,17 @@ func parseEnvSettings() map[string]string {
 func parseURLSettings(connString string) (map[string]string, error) {
 	settings := make(map[string]string)
 
-	url, err := url.Parse(connString)
+	parsedURL, err := url.Parse(connString)
 	if err != nil {
-		return nil, err
+		if urlErr := new(url.Error); errors.As(err, &urlErr) {
+			return nil, urlErr.Err
+		}
+		return nil, errors.New("failed to parse URL")
 	}
 
-	if url.User != nil {
-		settings["user"] = url.User.Username()
-		if password, present := url.User.Password(); present {
+	if parsedURL.User != nil {
+		settings["user"] = parsedURL.User.Username()
+		if password, present := parsedURL.User.Password(); present {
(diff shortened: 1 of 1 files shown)


pgconn/config.go
```diff
@@ -467,22 +467,25 @@ func parseEnvSettings() map[string]string {
...
 	settings := make(map[string]string)
 
-	url, err := url.Parse(connString)
+	parsedURL, err := url.Parse(connString)
 	if err != nil {
-		return nil, err
+		if urlErr := new(url.Error); errors.As(err, &urlErr) {
+			return nil, urlErr.Err
+		}
+		return nil, errors.New("failed to parse URL")
```
serprex, line 475:
```suggestion
		return nil, err
```
for non url.Error should be able to return underlying error

jackc: I agree with change out of general principal. Though this path will never be reached, url.Parse only returns url.Error.

heavycrystal: changed

PR #2083, opened 2024-07-11
V1 XMLCodec supports encoding + scanning XML column type

nickcruess-soda:
We have a need for the XML column type but observed that it isn't currently supported. I've taken the JSONCodec as a starting point and adapted it for XML. They turned out to be quite similar. I left in some of the issue-referencing comments from the JSONCodec which seemed relevant here as well, such as the one explaining the `driver.Valuer` vs `xml.Marshaler` ordering in `PlanEncode`.

While I was at it I also noticed some unused code related to the JSONCodec (`scanPlanJSONToBytesScanner`), which I removed in the second commit. Happy to drop that if this isn't the place/time for it.

### DIFF
--- a/pgtype/xml.go
+++ b/pgtype/xml.go
@@ -1,0 +1,198 @@
+package pgtype
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"encoding/xml"
+	"fmt"
+	"reflect"
+)
+
+type XMLCodec struct {
+	Marshal   func(v any) ([]byte, error)
+	Unmarshal func(data []byte, v any) error
+}
+
+func (*XMLCodec) FormatSupported(format int16) bool {
+	return format == TextFormatCode || format == BinaryFormatCode
+}
+
+func (*XMLCodec) PreferredFormat() int16 {
+	return TextFormatCode
+}
+
+func (c *XMLCodec) PlanEncode(m *Map, oid uint32, format int16, value any) EncodePlan {
+	switch value.(type) {
+	case string:
+		return encodePlanXMLCodecEitherFormatString{}
+	case []byte:
+		return encodePlanXMLCodecEitherFormatByteSlice{}
+
+	// Cannot rely on driver.Valuer being handled later because anything can be marshalled.
+	//
+	// https://github.com/jackc/pgx/issues/1430
+	//
+	// Check for driver.Valuer must come before xml.Marshaler so that it is guaranteed to be used
+	// when both are implemented https://github.com/jackc/pgx/issues/1805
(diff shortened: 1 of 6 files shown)


stdlib/sql.go
```diff
@@ -795,6 +795,16 @@ func (r *Rows) Next(dest []driver.Value) error {
...
+			case pgtype.XMLOID:
+				var d []byte
+				scanPlan := m.PlanScan(dataTypeOID, format, &d)
+				r.valueFuncs[i] = func(src []byte) (driver.Value, error) {
+					err := scanPlan.Scan(src, &d)
+					if err != nil {
+						return nil, err
+					}
+					return d, nil
+				}
```
nickcruess-soda, line 807: Not totally sure if this is necessary / correct -- we don't use the standard library but observed that JSON was registered this way, so I followed suit. Open to advice or correction.

jackc: To be honest my memory is a little fuzzy on that. But I think what is happening is that ensures the type of value passed to a `sql.Scanner` is a `[]byte` instead of a `string`. Not sure why that was preferred. But it should be safe to do the same thing for xml.

jackc:
It looks like the tests will need to skip CockroachDB.

stdlib/sql.go
nickcruess-soda: Ahh, understood

pgtype/xml.go
```diff
@@ -0,0 +1,198 @@
...
+	return dstBuf, nil
+}
+
+func (c *XMLCodec) DecodeValue(m *Map, oid uint32, format int16, src []byte) (any, error) {
+	if src == nil {
+		return nil, nil
+	}
+
+	var dst any
+	err := c.Unmarshal(src, &dst)
```
lyuboxa, line 196: @nickcruess-soda  @jackc I seem to be late to the game but this will always return nil, since `Unmarshal(..)` expects the variable to be of slice, struct or string. https://pkg.go.dev/encoding/xml#Unmarshal

```
Unmarshal parses the XML-encoded data and stores the result in the value pointed to by v, 
which must be an arbitrary struct, slice, or string. 
Well-formed data that does not fit into v is discarded.
```
https://go.dev/play/p/yLdFqZPaoUo

Returning the bytes seem to be the better choice. 

Am I missing something?

jackc: @lyuboxa I think you're right.

@nickcruess-soda What do you think?

lovromazgon: @nickcruess-soda @jackc Any update on this bug?

PR #2084, opened 2024-07-11
Implement 'MultiTracer'

EpicStep:
Hi. I'm tried to implement multi tracer (https://github.com/jackc/pgx/discussions/1677)
I took the fact that in ConnConfig the main tracer is QueryTracer and I decided to split from it into different tracers inside the constructor. If the API is good, I'll write the tests.


jackc:
Seems reasonable.

EpicStep:
Should we support pgxpool Acquire/Release Tracer at MultiTracer?

jackc:
> Should we support pgxpool Acquire/Release Tracer?

Yes.

EpicStep:
I'm done. I moved multi tracer to a separate package to avoid an import cycle.

PR #2088, opened 2024-07-16
Lock EnumCodec.membersMap during reads and writes

jamesroutley:
We've been seeing a nondeterministic `fatal error: concurrent map writes` error while using PGX. I traced it to the `c.membersMap[s] = s` line, and I think that this is a true issue - there doesn't appear to be anything preventing concurrent access to this map.

This PR adds a `sync.RWMutex` to `EnumCodec` and calls the appropriate `RLock` or `Lock` methods around read or write access.

### DIFF
--- a/pgtype/enum_codec.go
+++ b/pgtype/enum_codec.go
@@ -3,24 +3,26 @@ package pgtype
 import (
 	"database/sql/driver"
 	"fmt"
+	"sync"
 )
 
 // EnumCodec is a codec that caches the strings it decodes. If the same string is read multiple times only one copy is
 // allocated. These strings are only garbage collected when the EnumCodec is garbage collected. EnumCodec can be used
 // for any text type not only enums, but it should only be used when there are a small number of possible values.
 type EnumCodec struct {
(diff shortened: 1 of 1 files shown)


pgtype/enum_codec.go
```diff
@@ -3,24 +3,26 @@ package pgtype
...
 // allocated. These strings are only garbage collected when the EnumCodec is garbage collected. EnumCodec can be used
 // for any text type not only enums, but it should only be used when there are a small number of possible values.
 type EnumCodec struct {
-	membersMap map[string]string // map to quickly lookup member and reuse string instead of allocating
+	membersMap   map[string]string // map to quickly lookup member and reuse string instead of allocating
+	membersMutex sync.RWMutex
 }
 
-func (EnumCodec) FormatSupported(format int16) bool {
+func (*EnumCodec) FormatSupported(format int16) bool {
```
jamesroutley, line 17: I needed to make these pointer receivers to avoid `FormatSupported passes lock by value: github.com/jackc/pgx/v5/pgtype.EnumCodec contains sync.RWMutex` go vet errors

Issue #2043, opened 2024-06-13
Tracelog uses "time" when logging which can cause issues with JSON payloads


vicentsegui:
**Describe the bug**
The tracelog packlage (github.com/jackc/pgx/v5/tracelog") uses "time" as a label to print the time used to process. This however causes conflicts with other logging systems that will use "time" as a label to mean timestamp of log produced. This causes log ingestors to fail parsing the JSON as there is a duplicate key.

**To Reproduce**

Use github.com/jackc/pgx/v5/tracelog and "github.com/jackc/pgx-zerolog" and a logger configured roughly like so:

```
	log = zerolog.New(output).
			Level(zerolog.Level(logLevel)).
			With().
			Caller().
			Stack().
			Timestamp().
			Str("scm_version", version.ScmVersion).
			Logger()
```
and configure pgx pool like so:

```
	pgLogTrace := &tracelog.TraceLog{
		Logger:   zerologadapter.NewLogger(log),
		LogLevel: tracelog.LogLevelTrace,
	}
	pgConfig.ConnConfig.Tracer = pgxtrace.CompositeQueryTracer{
		pgLogTrace,
	}
	l.Info().Msg("Connecting to Postgres database")
	pgClient, err := pgxpool.NewWithConfig(ctx, pgConfig)
```

Use the pgClient to run some queries and observe the logs:

```json
{"level":"info",
 "module":"pgx",
"alreadyPrepared":false,
"name":"stmtcache_11578b0b996ce09b82a573adef17a56ff464aab9bbc4a4af",
"pid":426317,
"sql":"SELECT *********", 
"caller":"external/gazelle~~go_deps~com_github_jackc_pgx_zerolog/adapter.go:119", 
"time":3.17694, 
"time":"2024-06-13T15:20:12.937558654Z",
"message":"Prepare",
....}
```

Notice the duplicate "time" entry,.

**Expected behavior**

The tracelog package should avoid using generic labels like "time", etc as those can likely collide with other labels. I'd suggest adding a prefix (perhaps configurable) so that it will be logged like "pgx.QueryTime", for all tracelog labels.

**Version**
 - Go: `$ go version` -> 1.22
 - pgx:  `github.com/jackc/pgx/v5 v5.6.0`
- pgx-zerolog: `github.com/jackc/pgx-zerolog v0.0.0-20230315001418-f978528409eb`

jackc:
[...]

PR #2098, opened 2024-07-26
add TraceLogConfig with customizable TimeKey and ensureConfig method for default initialization

stringintech:
### Description

This PR introduces the `TraceLogConfig` struct to the `tracelog` package. The new struct includes a customizable `TimeKey` field with a default value of "time", allowing flexibility for logging configurations.

Additionally, the `ensureConfig` method is introduced to ensure that the `Config` field is initialized with default values if it is nil. Trace methods have been updated to call `ensureConfig()` to ensure that `Config` is properly initialized before logging.

### References

This PR addresses the issue discussed in Issue #2043.

### Feedback Request

Are there any additional key names or configurations that should be considered for inclusion in `TraceLogConfig` to further enhance customization? 

I would appreciate any feedback on whether I am following the correct procedure. Thank you.

### DIFF
--- a/tracelog/tracelog.go
+++ b/tracelog/tracelog.go
@@ -117,11 +117,31 @@ func logQueryArgs(args []any) []any {
 	return logArgs
 }
 
+// TraceLogConfig holds the configuration for key names
+type TraceLogConfig struct {
+	TimeKey string
+}
+
+// DefaultTraceLogConfig returns the default configuration for TraceLog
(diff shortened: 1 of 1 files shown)


jackc:
LGTM.

> Are there any additional key names or configurations that should be considered for inclusion in TraceLogConfig to further enhance customization?

I don't know of any.

PR #2102, opened 2024-08-04
Easy CRUD Operations Support

niting3c:
## Description 
This is a CRUD Repository Implementation for PGX pool connections and provides an easier way of using CRUD and pagination support for GET Queries.

### How to Use :
Inputs: 1st Argument is your SQL Query, followed by the arguments in order needed by the query.
It also supports transaction management , can also be used in a context based setting.

Sample Service: https://github.com/niting3c/gin-wire-starter 
Specific DB FIle: https://github.com/niting3c/gin-wire-starter/blob/main/internal/app/repository/userRepository.go

### Testing: 
- Unit tests are provided 
- It has been tested and in use in multiple of my Corporate Projects


jackc:
Was this PR intended for the main pgx repo? Everything is in an `internal` path so it wouldn't be usable.

PR #2112, opened 2024-08-21
support text scanner for binary format for uint32

jennifersp:


pgtype/uint32.go
```diff
@@ -282,6 +284,22 @@ func (scanPlanBinaryUint32ToUint32Scanner) Scan(src []byte, dst any) error {
 	return s.ScanUint32(Uint32{Uint32: n, Valid: true})
 }
 
+type scanPlanBinaryTextToUint32Scanner struct{}
```
jackc, line 287: I would expect this to be named `scanPlanBinaryUint32ToTextScanner`.

jackc:
Aside from the name, LGTM.

PR #2113, opened 2024-08-22
Add comment for pgtype.Interval struct

mateuszkowalke:
Comment explicitly states that pgtype.Interval is a nullable postgres type.


pgtype/interval.go
```diff
@@ -26,6 +26,21 @@ type IntervalValuer interface {
 	IntervalValue() (Interval, error)
 }
 
+// Interval represents an interval that may be null.
+// Interval implements the [Scanner] interface so
+// it can be used as a scan destination:
```
jackc, line 31: Implementing Scanner is not why it can be used as a destination in pgx. That's only when used through database/sql.

pgtype/interval.go
```diff
@@ -26,6 +26,21 @@ type IntervalValuer interface {
...
+//	var i Interval
+//	err := conn.QueryRow(context.Background(), "SELECT elapsed_time FROM foo WHERE id=$1", id).Scan(&i)
+//	...
+//	if i.Valid {
+//	   // use i.Days...
+//	} else {
+//	   // NULL value
+//	}
+//
+// When using as a parameter for prepared statement
```
jackc, line 42: Whether or not a prepared statement is used makes no difference.

pgtype/interval.go
```diff
@@ -26,6 +26,21 @@ type IntervalValuer interface {
...
+// the [Valid] field has to be explicitly set to [true] by the user.
```
jackc, line 43: Valid has to be set if it is not `NULL`. It is perfectly reasonable for `Valid` to be false when representing a `NULL`.

jackc (commented):
Sorry, but there are several errors in the comments.

But beyond that, I really don't think that this is the place to explain `Valid`. Otherwise, we would need more or less duplicate comments on all the types. A package level comment on the `pgtype` package would be more reasonable.

mateuszkowalke:
Many thanks for your comments. Made me have a deeper look into pgtype. I added a short notice to the `doc.go`.
Anyway I get your point about the behavior of pgtype types being similar to database/sql and why it might be a bit redundant to add another explanation in pgtype, so feel free to simply reject changes.

Anyhow, wanted to thank you for your awesome work!

PR #2138, opened 2024-10-02
add message body size limits in frontend and backend

zenkovev:


zenkovev:
Refers to https://github.com/jackc/pgx/issues/2139

zenkovev:
Thanks for the reply. Regarding the control over the frontend creation, yes, it's fair, then global atomics are really not needed. There remains control over the length of the frontend message at the structure level - PR has been updated.

jackc:
The Frontend changes look okay. Seem to be just porting the existing system from the backend. But I'm not sure why there are any changes on Backend.

Also, the comment style of having the comments below the field instead of above or beside looks odd to me (Frontend.maxBodyLen).

zenkovev:
I fixed the comments. About the changes in the Backend: there is one change transferred from the Frontend, checking that a positive number is set in `b.bodyLen`, that is, that the specified length of the message is correct.

jackc:
👍

PR #2151, opened 2024-10-19
handling double pointer on sql.Scanner interface when scanning rows

### LINKED ISSUE #2146: PGX seems not calling the Scanner interface with JSONB column
**Describe the bug**
When implementing scanner and valuer interface of a type used with pgx, it seems that scanner interface is not used to read data in case of JSONB columns.  

> It is called when the column is not JSONB, I've tried with a bytea/varchar columns and it works perfectly.

**To Reproduce**

I've created the following repo to reproduce the issue: https://github.com/ludusrusso/pgx-protobuf

As you can see if you run `go run main.go`, the program panics with 
```
INFO[0000] calling valuer                               
FATA[0000] can't get tests: can't scan into dest[1]: json: cannot unmarshal string into Go struct field Test.timestamp of type timestamppb.Timestamp 
exit status 1
```

The problem is due to the fact that pgx is tring to unmarshal test with the `json` packages, while the struct has been marshalled with  `protojson`.

As you can see from logs, the `Scan` method is never called.

```go
func (t *Test) Scan(value interface{}) error {
	logrus.Info("calling scanner")

	bytes, ok := value.([]byte)
	if !ok {
		return errors.New("type assertion to []byte failed")
	}

	return protojson.Unmarshal(bytes, t)
}

func (t *Test) Value() (driver.Value, error) {
	logrus.Info("calling valuer")

	return protojson.Marshal(t)
}
```

**Expected behavior**
Scan method should be called.

**Actual behavior**
Scan method is not called.

**Version**
 - Go: `$ go version` -> go version go1.23.1 darwin/arm64
 - PostgreSQL: `$ psql --no-psqlrc --tuples-only -c 'select version()'` -> PostgreSQL 15.8 (Homebrew) on aarch64-apple-darwin23.4.0, compiled by Apple clang version 15.0.0 (clang-1500.3.9.4), 64-bit 
 - pgx: `$ grep 'github.com/jackc/pgx/v[0-9]' go.mod` -> v5.7.1

ludusrusso:
fix #2146 

This PR handle the case of double pointer on sql.Scanning by reversing the pointer three until we found a sql.Scanner interface

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/pgtype/json.go
+++ b/pgtype/json.go
@@ -143,16 +143,32 @@ func (c *JSONCodec) PlanScan(m *Map, oid uint32, format int16, target any) ScanP
 	case BytesScanner:
 		return scanPlanBinaryBytesToBytesScanner{}
 
+	}
+
 	// Cannot rely on sql.Scanner being handled later because scanPlanJSONToJSONUnmarshal will take precedence.
 	//
 	// https://github.com/jackc/pgx/issues/1418
-	case sql.Scanner:
+	if isSQLScanner(target) {
 		return &scanPlanSQLScanner{formatCode: format}
 	}
 
 	return &scanPlanJSONToJSONUnmarshal{
 		unmarshal: c.Unmarshal,
 	}
+}
+
+// we need to check if the target is a pointer to a sql.Scanner (or any of the pointer ref tree implements a sql.Scanner).
+//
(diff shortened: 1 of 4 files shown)


jackc:
LGTM

PR #2156, opened 2024-10-23
Update TimestampCodec  to be equal to TextCodec for pointer reference

mesmerx:
working with 
https://github.com/manniwood/pgx-protobuf-timestamp i receive  
`cannot use TimestampCodec{} (value of type TimestampCodec) as pgtype.Codec value in struct literal: TimestampCodec does not implement pgtype.Codec (method DecodeDatabaseSQLValue has pointer receiver)`

and i found that is because of timestamp have pointers where textecodec(athat works) dont have, this is to normalize the both

sems that is introduced in https://github.com/jackc/pgx/commit/8649231bb3bc00b4b9c180ce557a54ae41c28ce2
idk if its the better approach, but it need to be fixed for 

```
type TextCodec struct {
	pgtype.TextCodec
}


...

// TimestampCodec embeds pgtype.TimestampCodec, which implements pgtype.Codec interface
type TimestampCodec struct {
	pgtype.TimestampCodec
}


// Register registers the github.com/gofrs/uuid integration with a pgtype.Map.
func Register(tm *pgtype.Map) {
	tm.TryWrapEncodePlanFuncs = append(
		[]pgtype.TryWrapEncodePlanFunc{TryWrapTimestampEncodePlan},
		tm.TryWrapEncodePlanFuncs...)
	tm.TryWrapScanPlanFuncs = append(
		[]pgtype.TryWrapScanPlanFunc{TryWrapTimestampScanPlan},
		tm.TryWrapScanPlanFuncs...)

	tm.RegisterType(&pgtype.Type{
		Name:  "timestamp",
		OID:   pgtype.TimestampOID,
		Codec: TimestampCodec{},
	})
}
```

to work

i believe that i can overwrite some functions to make work in my side, but i think we need to set this here

### DIFF
--- a/pgtype/timestamp.go
+++ b/pgtype/timestamp.go
@@ -122,15 +122,15 @@ type TimestampCodec struct {
 	ScanLocation *time.Location
 }
 
-func (*TimestampCodec) FormatSupported(format int16) bool {
+func (TimestampCodec) FormatSupported(format int16) bool {
 	return format == TextFormatCode || format == BinaryFormatCode
 }
 
-func (*TimestampCodec) PreferredFormat() int16 {
+func (TimestampCodec) PreferredFormat() int16 {
 	return BinaryFormatCode
 }
 
-func (*TimestampCodec) PlanEncode(m *Map, oid uint32, format int16, value any) EncodePlan {
+func (TimestampCodec) PlanEncode(m *Map, oid uint32, format int16, value any) EncodePlan {
 	if _, ok := value.(TimestampValuer); !ok {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


mesmerx:
usage of 
```
 tm.RegisterType(&pgtype.Type{
		Name:  "timestamp",
		OID:   pgtype.TimestampOID,
		Codec: &TimestampCodec{},
	})
```
seems to fix the issue too, but i will test more here

mesmerx:
closing because 
```
 tm.RegisterType(&pgtype.Type{
		Name:  "timestamp",
		OID:   pgtype.TimestampOID,
		Codec: &TimestampCodec{},
	})
```
fix without messing pointers

Issue #2168, opened 2024-11-19
target_session_attrs cannot be used with replication connection


thedolphin:
**Describe the bug**

target_session_attrs cannot be used with replication connection

**To Reproduce**

```go

config, _ := pgconn.ParseConfig("target_session_attrs=read-write")
config.RuntimeParams["replication"] = "database"
pgConn, err := pgconn.ConnectConfig(ctx, config)
if err != nil {
    log.Print(err)
}
```

**Expected behavior**

Primary server was found and connection succeeded

**Actual behavior**
```text
ValidateConnect failed: ERROR: extended query protocol not supported in a replication connection (SQLSTATE 08P01)
```

**Version**
 - Go: `go version go1.22.7 linux/amd64`
 - PostgreSQL: 16
 - pgx: `github.com/jackc/pgx/v5 v5.7.1`

**Additional context**

Seems that `ExecParams` in `ValidateConnectTargetSessionAttrs*` functions is redundant and could be replaced with `Exec` function, while it triggers extended query protocol usage
https://github.com/jackc/pgx/blob/b9e2b20fb13a7eb8e6bf105c074c291b349b6dec/pgconn/config.go#L864

PR #2169, opened 2024-11-19
Switch from ExecParams to Exec in ValidateConnectTargetSessionAttrs

thedolphin:
Resolves https://github.com/jackc/pgx/issues/2168


jackc:
👍

Issue #2177, opened 2024-11-22
Ability to use ":" instead of "@" in named args


KoNekoD:
Add ability to use ":" instead of "@" in named args

Sql linter in golang like this

KoNekoD:
Just add ":" to here https://github.com/jackc/pgx/blob/master/named_args.go#L110

KoNekoD:
PR https://github.com/jackc/pgx/pull/2178

### Before(bad)

![image](https://github.com/user-attachments/assets/f507c3b1-5c98-4ff8-9534-8c71f42b38b5)


### After (good)

![image](https://github.com/user-attachments/assets/437ba977-c1cf-4475-97f2-813cca8b0bab)

KoNekoD:
@jackc I don't think it will bring any significant problems, but only solve the “mad linter in goland” problem.

In fact, your implementation provokes goland to warnings, wouldn't it be better to let users choose whether to continue using '@' or to refuse explicit conversion (it will be even a plus because of the cost of conversion itself), but it will be code without warnings.

What do you think about this improvement?

KoNekoD:
@jackc  I see your point, I'll make the edits according to your concerns.

KoNekoD:
@jackc Done! Commit 2aa476d2f92d5599543e5db4e8f6aa14d3cb9369

jackc:
The code looks more solid now.

But as I said above, I want to keep the simplicity of only having a single delimiter for named arguments.

I think this support for `:` should be pulled out into your own library or application.

KoNekoD:
@jackc Are you sure that library users will be comfortable being forced to adjust to your understanding of good? There seem to be some people who don't like it when their IDE frowns on '@' substitutions. Maybe we should pay attention to the convenience for the end user of the library?

Nerobrain:
@KoNekoD Are you sure that library users will be comfortable being forced to adjust to your understanding of good? There seem to be some people who know difference between `errors` and `linter faults`. Maybe we should pay attention to the workload for the autor of the library?

devsalesexpress:
> `:`é complicado porque também é usado para conversão de tipo, por exemplo `select somecol::text`. Até onde eu sei, #2178 falharia quando `::`fosse encontrado.
> 
> Mas mesmo que o analisador fosse corrigido, eu preferiria que houvesse apenas uma maneira de denotar um argumento nomeado.
> 
[...]

PR #2178, opened 2024-11-22
add ability to use ':' in named args

KoNekoD:


named_args.go
```diff
@@ -107,6 +107,19 @@ func rawState(l *sqlLexer) stateFn {
 			return singleQuoteState
 		case '"':
 			return doubleQuoteState
+		case ':':
+			nextRune, _ := utf8.DecodeRuneInString(l.src[l.pos:])
+			prevRune := rune(0)
+			if l.pos > 1 {
```
KoNekoD, line 113: I added this check to avoid panic when : is at the beginning of a line.

named_args.go
```diff
@@ -107,6 +107,19 @@ func rawState(l *sqlLexer) stateFn {
...
+				prevRune, _ = utf8.DecodeRuneInString(l.src[l.pos-2:])
+			}
+			if nextRune != ':' && prevRune != ':' && (isLetter(nextRune) || nextRune == '_') {
```
KoNekoD, line 116: It seems to me that we can omit the first check nextRune != ':' as there will be a more specific check next, but in addition I would like to say that this check is much easier than the next ones, and will cut off cast types a little faster.... But type casts are not done so often to leave this prevenient check.

What do you think, should I remove `nextRune != ':' &&` ?

named_args.go
```diff
@@ -107,6 +107,19 @@ func rawState(l *sqlLexer) stateFn {
...
+				if l.pos-l.start > 0 {
+					l.parts = append(l.parts, l.src[l.start:l.pos-width])
+				}
+				l.start = l.pos
+				return namedArgState
```
KoNekoD, line 121: Copy from '@' case

named_args_test.go
```diff
@@ -160,3 +160,155 @@ func TestStrictNamedArgsRewriteQuery(t *testing.T) {
 		}
 	}
 }
+
+func TestNamedArgsRewriteQuery2(t *testing.T) {
```
KoNekoD, line 164: I also added a test that should confirm the clarity of my implementation

devsalesexpress:
The implementation of KoNekoD makes perfect sense, I also have this need and I made the change in my project and it worked perfectly as expected, because I have queries that run on Postgres, Oracle and MySql and by default we use the named parameter with ':', for example :id

Ability to use ":" instead of "@" in named arg

jackc:
See https://github.com/jackc/pgx/issues/2177#issuecomment-2585345296.

PR #2189, opened 2024-12-12
Update golang.org/x/crypto v0.27.0 => v0.31.0 to fix vulnerability

pankona:
Hi!

This is a vulnerability report related to golang.org/x/crypto. https://github.com/advisories/GHSA-v778-237x-gjrc 
It says the vulnerability is fix golang.org/x/crypto v0.31.0 or after. pgx is now depends on v0.27.0. How about to update it to v0.31.0? Thanks!

---
I've do followings:
- Update go.mod to bump crypto version to v0.31.0
- Then `go mod tidy`


jackc:
I suppose we can placate vulnerability scanners, but pgx actually doesn't use the vulnerable functionality in golang.org/x/crypto.

PR #2195, opened 2024-12-17
Add CommitQuery to transaction options

LucasHild:
This PR adds the ability to override the query which is used to commit a transaction.

This can be helpful for investigating performance issues (see https://knock.app/blog/hacking-postgresql-driver). For example, instead of committing a transaction with `commit`, we did override it to `commit -- POST /my/endpoint`. This way, you can break down the commit performance by endpoints in tools like RDS Performance Insights.


jackc:
Thanks

Issue #2205, opened 2024-12-23
pgxpool.Stat: Missing EmptyAcquireWaitTime()


pquerna:
**Is your feature request related to a problem? Please describe.**
- `EmptyAcquireWaitTime()` is available in `puddle` since https://github.com/jackc/puddle/pull/34
- But it wasn't re-exposed in `pgxpool.Stat{}`

**Describe the solution you'd like**
- add an accessor function.

**Additional context**
- Related issue https://github.com/jackc/pgx/issues/1669

vamshiaruru:
This would a pretty easy thing to add, right? Is someone adding it? I don't mind raising a PR for this as well.

jackc:
I think it should be pretty easy - and no one is working on it as far as I know.

vamshiaruru:
Created this PR: https://github.com/jackc/pgx/pull/2208

jackc:
Merged #2208.

PR #2208, opened 2024-12-26
Expose puddle.Pool's EmptyAcquireWaitTime in pgxpool's Stats

vamshiaruru:
Addresses: https://github.com/jackc/pgx/issues/2205


vamshiaruru:
I haven't added any tests it. Looks like methods like `EmptyAcquireCount` weren't tested too, so I just left it as is.

PR #2213, opened 2024-12-30
Fix #2204

### LINKED ISSUE #2204: Regression in `Rows.Scan`: incorrect handling of `sql.Scanner` implementation in v5.7.2
The `Rows.Scan` method of the `pgx` package is documented as follows:  

> 	// Scan reads the values from the current row into dest values positionally.
> 	// dest can include pointers to core types, values implementing the Scanner
> 	// interface, and nil. nil will skip the value entirely. It is an error to
> 	// call Scan without first calling Next() and checking that it returned true.

According to the documentation, it accepts as arguments:  

1. pointers to core types,  
2.  values implementing the `Scanner` interface,  
3.  `nil`.  

Starting from version v5.7.2, point (2) is no longer true because:  

1. It no longer accepts non-pointer values implementing the `sql.Scanner` interface.  
2. It accepts pointer values that **do not** implement `sql.Scanner` but point to types that do.  

Both changes contradict the method's documentation. In particular:  

- The first change breaks existing programs that relied on the documented behavior.  
- The second change allows invalid values to pass through, potentially hiding parameter-passing errors.  

These issues were introduced in commit 5c9b565. The issue #2146, which this commit intended to resolve, did not need fixing in the first place because the original behavior was correct. The [example code](https://github.com/jackc/pgx/issues/2146#issuecomment-2406721196) in the issue is incorrect. Specifically:

```go  
var r Row  
err := rows.Scan(&r.ID, &r.Data)  
...  
type Row struct {  
	ID   int        `json:"id"`  
	Data *Timestamp `json:"data"` // <- this is a pointer  
}  
...  
func (t *Timestamp) Scan(value interface{}) error { ... }  
```  

The second argument to `rows.Scan` has type `**T`, but `**T` does not implement `sql.Scanner`. Therefore, the `Scan` method has no `Scan` implementation to invoke.  

The example code should be corrected as follows:  

```go  
err := rows.Scan(&r.ID, r.Data)  
```  

Here, `r.Data` must not be passed as a pointer.  

As a result, commit 5c9b565 should be reverted to restore the correct behavior, which was consistent with both the documentation and expected usage.

moukoublen:
Fix issues discussed in #2204 

  1. Add the missing check in both in `getSQLScanner` (used in `json` and `jsonb` codecs `PlanScan`) and `getSQLScanner` (used in `scanPlanSQLScanner.Scan`) in case the destination value is not a pointer **but** it is an `sql.Scanner`. At the same time I left the reflection to be used only in cases where chain-following is needed. Otherwise (when the type is directly an `sql.Scanner` the reflection will not be used which is a performance gain.
  2. Updated `scanPlanJSONToJSONUnmarshal.Scan` so as not to aggressively assume that the target destination is a pointer because if it is not, it leads to panic. Followed an approach similar to what `json.Unmarashal` [does](https://cs.opensource.google/go/go/+/master:src/encoding/json/decode.go;l=171-175).


jackc:
Thanks!

Issue #2215, opened 2025-01-02
Unmarshaling JSON from pgtype.Timestamp is broken


pconstantinou:
**Describe the bug**
The change here:
https://github.com/jackc/pgx/commit/7803ec3661b8f10241ee7094c850f9184d95f07a
introduces an error where JSON marsharing and unmarshalling fails.

The unit test below shows that if you create a `pgtype.Timestamp` in a struct, marshal it to JSON and then unmarshal it back into the struct, the JSON unmarshaling fails with an error: 
`Received unexpected error:
                                parsing time "2025-01-02T20:17:44.033105ZZ": extra text: "Z"
`

**To Reproduce**
See this: https://goplay.tools/snippet/fO_ceIqGn7d
or 

```package main

import (
	"encoding/json"
	"fmt"
	"testing"
	"time"

	"github.com/jackc/pgx/v5/pgtype"
	"github.com/stretchr/testify/assert"
)

type TestStruct struct {
	TS pgtype.Timestamp `json:"t"`
}

// ToTimestampFromTime converts time.Time
func ToTimestampFromTime(t time.Time) pgtype.Timestamp {
	var pgt pgtype.Timestamp
	if err := pgt.Scan(t.UTC()); err != nil {
		fmt.Printf("Failed to populate pgtype.Timestamp %v", err)
	}
	return pgt
}

func TestTimestampRoundTrip(t *testing.T) {
	ts := ToTimestampFromTime(time.Now())
	s := TestStruct{TS: ts}
	b, err := json.Marshal(s)
	assert.NoError(t, err)
	newMap := make(map[string]any)
	err = json.Unmarshal(b, &newMap)
	assert.NoError(t, err)
	var s2 TestStruct
	err = json.Unmarshal(b, &s2)
	assert.NoError(t, err)
	assert.Equal(t, ts, s2.TS)
	assert.Equal(t, ts, s.TS)
}

```

This unit test fails on:

```
                Error Trace:                    Error:          Received unexpected error:
                                parsing time "2025-01-02T20:17:44.033105ZZ": extra text: "Z"
```

Please run your example with the race detector enabled. For example, `go run -race main.go` or `go test -race`.

**Expected behavior**
Round-trip marshal and unmarshal should be possible without an error.

**Actual behavior**
An error occurs when unmarshaling `pgtype.Timestamp`

**Version**
 - Go: go version go1.23.0 darwin/arm64
 - pgx: github.com/jackc/pgx/v5 v5.7.2

PR #2216, opened 2025-01-02
Timestamp incorrectly adds 'Z' when serializing from JSON to indicate GMT, fixes bug #2215

pconstantinou:
Fixes https://github.com/jackc/pgx/issues/2215

### DIFF
--- a/pgtype/timestamp_test.go
+++ b/pgtype/timestamp_test.go
@@ -116,9 +116,9 @@ func TestTimestampMarshalJSON(t *testing.T) {
 		result string
 	}{
 		{source: pgtype.Timestamp{}, result: "null"},
-		{source: pgtype.Timestamp{Time: tm, Valid: true}, result: "\"2012-03-29T10:05:45Z\""},
-		{source: pgt, result: "\"2012-03-29T10:05:45Z\""},
-		{source: pgtype.Timestamp{Time: time.Date(2012, 3, 29, 10, 5, 45, 555*1000*1000, time.UTC), Valid: true}, result: "\"2012-03-29T10:05:45.555Z\""},
+		{source: pgtype.Timestamp{Time: tm, Valid: true}, result: `"2012-03-29T10:05:45"`},
+		{source: pgt, result: `"2012-03-29T10:05:45"`},
+		{source: pgtype.Timestamp{Time: time.Date(2012, 3, 29, 10, 5, 45, 555*1000*1000, time.UTC), Valid: true}, result: `"2012-03-29T10:05:45.555"`},
 		{source: pgtype.Timestamp{InfinityModifier: pgtype.Infinity, Valid: true}, result: "\"infinity\""},
 		{source: pgtype.Timestamp{InfinityModifier: pgtype.NegativeInfinity, Valid: true}, result: "\"-infinity\""},
 	}
@@ -137,6 +137,7 @@ func TestTimestampMarshalJSON(t *testing.T) {
 		t2 := tsStruct
 		err = json.Unmarshal(b, &t2)
 		assert.NoErrorf(t, err, "failed to unmarshal %v with %s", tt.source, err)
+		assert.True(t, tsStruct.TS.Time.Unix() == t2.TS.Time.Unix())
 	}
 }
 
--- a/pgtype/timestamp_test.go
+++ b/pgtype/timestamp_test.go
@@ -2,12 +2,14 @@ package pgtype_test
 
 import (
 	"context"
+	"encoding/json"
 	"testing"
 	"time"
 
 	pgx "github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/jackc/pgx/v5/pgxtest"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


pconstantinou:
@s-montigny-desautels any chance you could review this?

jackc:
Now that I'm looking at this functionality a little closer, it seems to me that both `MarshalJSON` and `UnmarshalJSON` are wrong. Why is `time.RFC3339Nano` being used at all? A `timestamp` doesn't have a time zone.

I would expect pgtype.Timestamp's JSON behavior to match PostgreSQL.

```
postgres@[local]:5015 pgx_test=# select to_json(now()::timestamp);
           to_json
──────────────────────────────
 "2025-01-03T18:57:40.099807"
(1 row)
```

pconstantinou:
> Now that I'm looking at this functionality a little closer, it seems to me that both `MarshalJSON` and `UnmarshalJSON` are wrong. Why is `time.RFC3339Nano` being used at all? A `timestamp` doesn't have a time zone.
> 
> I would expect pgtype.Timestamp's JSON behavior to match PostgreSQL.
> 
> ```
> postgres@[local]:5015 pgx_test=# select to_json(now()::timestamp);
>            to_json
> ──────────────────────────────
>  "2025-01-03T18:57:40.099807"
> (1 row)
> ```

It looks like RFC3339  is endorsed by ISO 8601. - https://www.rfc-editor.org/rfc/rfc3339
The RFC allows a timezone to be present but doesn't require it

```
	RFC3339     = "2006-01-02T15:04:05Z07:00"
	RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
```

The issue I encountered was that the library added the "Z" even if it was there. I'd be happy if the whole change in https://github.com/jackc/pgx/commit/7803ec3661b8f10241ee7094c850f9184d95f07a was rolled back.

s-montigny-desautels:
The fix I did was to match the postgres JSON implementation, by adding the missing "Z" to the returned string. As @jackc said, `timestamp` don't have timezone, but `time.RFC3339Nano` requires it. That's why I've added the missing "Z", assuming the timezone is UTC and user could change it afterwards. I didn't want to change too much code, but maybe a custom format should be used instead of the `time.RFC3339Nano`.

I don't think reverting 7803ec3 is an good option either, since it will break UnmarshalJSON when getting JSON timestamp from postgres.

@pconstantinou Where are you getting your timezone from? Maybe you are not using the proper type, since your timestamp has a timezone information in it.

pconstantinou:
> @pconstantinou Where are you getting your timezone from? Maybe you are not using the proper type, since your timestamp has a timezone information in it.

```
	var pgt pgtype.Timestamp
	if err := pgt.Scan(t.UTC()); err != nil {
		fmt.Printf("Failed to populate pgtype.Timestamp %v", err)
	}
	return pgt
```


See: https://github.com/jackc/pgx/issues/2215 with a succinct example here: https://goplay.tools/snippet/fO_ceIqGn7d

s-montigny-desautels:
I see. Then, maybe, we should modify the `MarshalJSON` function instead? I don't think it should add the timezone information, since timestamp don't have one. What do you think?

pconstantinou:
> I see. Then, maybe, we should modify the `MarshalJSON` function instead? I don't think it should add the timezone information, since timestamp don't have one. What do you think?

I've made `MarshalJSON` write out standard ISO 8601 JSON formatted timestamp with only the date and time (no timezone). The internal representation of `time.Time` seems to try to standardize on `UTC` which could a little problematic but I'm not sure that there's a better option. 
Modifying `MarshalJSON` should make it clear that the `Timestamp` object does not include a Timezone.

As a consequence, this change does modify the expected results in the unit tests. The old version of the unit tests *expected* a timezone, I've removed that.

When Unmarshalling, I've changed the implementation to start parsing using with `RFC3339Nano` and if that is unsuccessful then parse with 8601 setting the timezone to UTC. This is cleaner than just appending an `Z` which will do the wrong thing in several situations. This also makes the implementation in line with the JSON spec.

For my particular usecase, I've persisted JSON generated from the older version of the library, the appending `Z` broke backwards compatibility. Fixing that remains important to me and it seems like an absolute requirement for any fix.

pconstantinou:
@jackc can you take another look? I'd love to get this merged or decide on a different approach.

pgtype/timestamp_test.go
```diff
@@ -116,12 +129,32 @@ func TestTimestampMarshalJSON(t *testing.T) {
 			t.Errorf("%d: %v", i, err)
 		}
 
-		if string(r) != tt.result {
+		if !assert.Equal(t, tt.result, string(r)) {
```
jackc, line 132: I don't understand this change.

jackc:
So if I understand correctly, this change makes marshalling and unmarshalling match PostgreSQL? If so then this should be fine.

pgtype/timestamp_test.go
pconstantinou: The `! assert.Equal(...)` function evaluates to the same as `string(r) != tt.result` but will also output a standard error message from `testify` showing clearly the expected and actual value.

pconstantinou:
> So if I understand correctly, this change makes marshaling and unmarshalling match PostgreSQL? If so then this should be fine.

Yes. It will match PostgresSQL marshaling and unmarshalling, by default excluding the timezone.

Additionally, if the data was marshaled using an older version of the `pgx` library (which used to include the timezone) it will continue to work without error.

jackc:
Thanks!

PR #2225, opened 2025-01-09
Include the field name in error messages when scanning structs

logicbomb:
This is helpful when the order of the columns in the row is unknown (e.g. when issuing a `select *` query).

### DIFF
--- a/pgtype/json_test.go
+++ b/pgtype/json_test.go
@@ -6,6 +6,7 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


jackc:
The idea of including the column header is reasonable. But just the column name in parenthesis seems like it could be ambiguous. I could see people mistaking that for the name of the variable they are attempting to scan into (even though that's not actually possible in Go). Not sure what else to do without making the error message too long though.

Beyond that, why use a `*string` for field name instead of just a `string`? That leaves the possibility for nil as well as pinning the whole field descriptions slice in memory.

The tests would need to be fixed as well.

logicbomb:
@jackc i've updated the code & tests, one test is failing locally but it looks related to my timezone settings and not the changes I'm proposing.

```
?       github.com/jackc/pgx/v5/testsetup       [no test files]
--- FAIL: TestConnCopyWithAllQueryExecModes (0.07s)
    --- FAIL: TestConnCopyWithAllQueryExecModes/exec (0.01s)
        copy_from_test.go:71: Input rows and output rows do not equal: [[0 1 2 abc 2010-02-03 04:05:06 -0500 EST] [<nil> <nil> <nil> <nil> <nil>]] -> [[0 1 2 abc 2010-02-03 09:05:06 +0000 +0000] [<nil> <nil> <nil> <nil> <nil>]]
    --- FAIL: TestConnCopyWithAllQueryExecModes/simple_protocol (0.01s)
        copy_from_test.go:71: Input rows and output rows do not equal: [[0 1 2 abc 2010-02-03 04:05:06 -0500 EST] [<nil> <nil> <nil> <nil> <nil>]] -> [[0 1 2 abc 2010-02-03 09:05:06 +0000 +0000] [<nil> <nil> <nil> <nil> <nil>]]
FAIL
```

logicbomb:
this may be a bit painful to get through, I'm not sure how to run the full suite locally, so, i can only see test failures here.  I'll make updates as I see errors coming though.  Is there a way for me to run the test suite in my github account w/out needing to wait for you to press the button?

logicbomb:
alright, i was able to get the cockroach db tests running, i see failures but seem to be unrelated to the changes in this PR

```
Received unexpected error:
                                ERROR: temporary tables are only supported experimentally (SQLSTATE XCEXF)
                Test:           TestLogCopyFrom/describe_exec```

jackc:
Thanks!

Issue #1231, opened 2022-06-21
Mocking pgx pool connection for unit testing 


super3programmer:
Hello,
	I am trying to write a unit test for a function in which I fetched pgxpool connection from another microservice. But I am not understanding how to write unit test for this function and how to get mock pgxpool.Conn
	
	E.g.
	Func (a *ADB) Create(ctx *gin.Context, userID uuid){
		pgxpoolConn, err := a.dbservice.GetConnection(ctx)
		//GetConnection is function of DBService(microservice) which returns the single connection of  pgxpool

		pgxpoolConn.QueryRow(ctx, query, arguments)
		//Using that pgxpool connection takes the actions on the DB
	}

Thanks!

nikitacrit:
Hello!
1. Create your own interface, build mock with standard mock library (gomock, mockery). **- complex solution**
2. Use https://github.com/DATA-DOG/go-sqlmock. You also need to create your own interface because of capability between standard sql db and pgxpool signatures. Core problem - go-sqlmock support standart go sql db interface **- complex solution**
3. Just use https://github.com/pashagolub/pgxmock. All described above already done by community. **- ok solution**

bfontaine:
ping @jackc you might have missed that one

jackc:
There is no built-in mocking framework for pgx.

But here are three approaches for testing with pgx.

First, I typically integrate over the database when possible.

Second, when testing lower level functionality, the pgmock package allows mocking the database server. This is very low level and should probably **not** be used for application level testing, but it is possible.

Lastly, the application can define its own connection interface. e.g.

```go
type DB interface {
	Query(ctx context.Context, sql string, optionsAndArgs ...interface{}) (pgx.Rows, error)
	QueryRow(ctx context.Context, sql string, optionsAndArgs ...interface{}) pgx.Row
	Exec(ctx context.Context, sql string, arguments ...any) (pgconn.CommandTag, error)
}
```

In your case `a.dbservice.GetConnection(ctx)` would return that `DB` interface instead of an actual pgx type. In testing your service could return a mock interface.

StevenACoffman:
This comes up a lot and this issue gets referenced a lot, so I just wanted to add a bit more nuance to what has been said.

[...]

Issue #616, opened 2019-09-26
Local Testing: How to mock a connection


Nokel81:
I have an application that I am trying to test where I am using a pgx connection. I would like to be able to inject a mock connection into my code where I can then expect/return expectations for that connection.

I tried to use https://github.com/DATA-DOG/go-sqlmock but pgx rejects it. I have looked at pgmock but that doesn't seem to do this at all.

Any ideas?

jackc:
You can use pgmock to create a mock server that pgx can connect to. It's fairly low level. It basically requires specifying what you want to happen at the wire protocol level. A higher level layer could be built on top of it...

If your looking for more of a fake *pgx.Conn that doesn't have an underlying connection then that doesn't exist out of the box. But database/sql and pgx v3, the pgx v4 Rows type is an interface instead of a pointer. This was done specifically to allow for mocking the entire database connection. The way to do that would be for your application to work with interfaces instead of an actual *pgx.Conn.   Then you could pass in objects that responded to Query and Exec however you wanted.

So to summarize, the low level plumbing is there to do this at the wire level or at the connection handle level, but there are no high level mocking tools in pgx.

Nokel81:
Okay thanks for the response. However I don't quite see how the `Rows` interface can help with mocking the whole connection since it is only returned by `conn.Query` and similar so there doesn't seem to be a way to have them returned when possessing a dummy connection.

jackc:
How it helps is that you can make an entire mock connection. If `Rows` was a struct with private members you couldn't mock `Query` because you would have no way to build the `Rows` it returns.

Here's a bigger picture explanation. Your code that is using a DB connection probably shouldn't be using a reference to a Conn or a Pool.

Instead define your own interface with the subset of methods you use. This is what I have in one of my projects:

```go
type dbconn interface {
	Begin(ctx context.Context) (pgx.Tx, error)
	Exec(ctx context.Context, sql string, arguments ...interface{}) (pgconn.CommandTag, error)
	Query(ctx context.Context, sql string, optionsAndArgs ...interface{}) (pgx.Rows, error)
	QueryRow(ctx context.Context, sql string, optionsAndArgs ...interface{}) pgx.Row
}
```

Then I have a method that uses the database. e.g.

```go
func CreateBook(ctx context.Context, db dbconn, book Book) (*Book, error) {
// ...
}
```

[...]

PR #2226, opened 2025-01-10
Add variants of Pool.Acquire et al. to return a Connection interface

cleroux:
This PR aims to serve as a proof-of-concept and basis to continue the discussion in #1231 which was about mocking connections obtained from pgxpool.

For some context about my use case, I am migrating a microservice from database/sql to pgx. This service already has a comprehensive suite of unit tests including some tests using [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock) which I have mostly migrated to [pgxmock](https://github.com/pashagolub/pgxmock). Unfortunately there is one blocker. We use `Pool.Acquire` to get a `Conn` from the `Pool` so that we can `Exec` a pre-flight query to set session variables before each user's query. These session variables are used by our Foreign Data Wrapper to propagate distributed tracing information and debug flags to our distributed storage engine. Unfortunately, [pgxmock is unable to mock Pool.Acquire](https://github.com/pashagolub/pgxmock/issues/42) because `Pool.Acquire` returns a `Conn` struct which cannot be mocked.

I have found some prior discussions about mocking pgx in other issues (https://github.com/jackc/pgx/issues/616, https://github.com/jackc/pgx/issues/1231) but after some attempts I found the suggested solutions are not ideal for my use case.

The purpose of the microservice I am migrating is to route client queries to the appropriate customer database and enforce user query quotas and rate-limiting. This is the functionality being tested in the unit tests I am referring to, along with some query error-handling and cancellation tests. The components under test are agnostic to the queries so using a real Postgres instance or simulated database connection using pgmock seems like unnecessary complexity. We do also have some integration tests using a real Postgres instance but I think it would be difficult to achieve the same granularity of testing we currently have. Not to mention the additional costs of setting up and tearing down the database for tests where it doesn't actually matter.

In this case I really think being able to mock `Pool.Acquire` is appropriate because the connection pool is a dependency but the application behavior being tested is a separate concern from any DB functionality.

After some consideration and a few different attempts, it seems to me the least difficult solution would be if `Pool.Acquire` returned a Connection interface. And perhaps you might consider that if a you decide to release another major version. However, in the interest of not making breaking changes and minimizing effort for everyone, I am proposing the addition of some redundant functions for `Pool.Acquire` et al. that return a `Connection` interface instead of a `Conn` struct. These new functions would allow `Pool.Acquire` to be mocked and I have a working proof-of-concept PR for pgxmock which I will submit to those maintainers depending on the outcome of this discussion.

Would you be willing to consider some solution to allow mocking `Pool.Acquire`?

Thanks for pgx, and thanks for your attention and consideration in this issue.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -62,6 +62,28 @@ func (cc *ConnConfig) Copy() *ConnConfig {
 // ConnString returns the connection string as parsed by pgx.ParseConfig into pgx.ConnConfig.
 func (cc *ConnConfig) ConnString() string { return cc.connString }
 
+// Connection is a PostgreSQL connection handle. It is not safe for concurrent usage. Use a connection pool to manage
+// access to multiple database connections from multiple goroutines.
+//
+// Connection is an interface to allow tests to mock Pool.AcquireConnection.
+type Connection interface {
+	Close(ctx context.Context) error
+	Prepare(ctx context.Context, name, sql string) (sd *pgconn.StatementDescription, err error)
+	Deallocate(ctx context.Context, name string) error
+	DeallocateAll(ctx context.Context) error
+	WaitForNotification(ctx context.Context) (*pgconn.Notification, error)
+	IsClosed() bool
+	Ping(ctx context.Context) error
+	PgConn() *pgconn.PgConn
+	TypeMap() *pgtype.Map
+	Config() *ConnConfig
+	Exec(ctx context.Context, sql string, arguments ...any) (pgconn.CommandTag, error)
+	Query(ctx context.Context, sql string, args ...any) (Rows, error)
+	QueryRow(ctx context.Context, sql string, args ...any) Row
+	SendBatch(ctx context.Context, b *Batch) (br BatchResults)
+	LoadType(ctx context.Context, typeName string) (*pgtype.Type, error)
+}
+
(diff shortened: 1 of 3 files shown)


jackc:
To be honest, I'm not very positive on adding parallel / duplicate functionality solely in place for mocking. This is both because of the Go preference of returning normal values rather than interfaces as well as my opinion that mocking at the database driver layer is usually the wrong approach.

In your specific case, perhaps you could wrap pgxpool and use and mock your wrapped pool?

cleroux:
I completely agree with your sentiment about adding duplicate functionality and returning structs instead of interfaces.

With regard to your wrapper suggestion, I did actually start down that path a couple weeks ago and I came to the conclusion that in addition to implementing the wrapper, I would also need to implement much of the functionality of pgxmock and I decided to step back and look for an easier solution.

This PR was the least-effort and greatest benefit solution I arrived at. (Ie. pgxmock and all its users would also benefit)  So I felt I had to at least ask the question to see where this might lead.  I appreciate your response nonetheless.

If there comes a time to release a new major version of pgx, I hope you might still consider returning an interface from these functions as there does seem to be some precedent to support mocks based on the comments for [Row](https://github.com/jackc/pgx/blob/master/rows.go#L78-L81) and [Rows](https://github.com/jackc/pgx/blob/master/rows.go#L23-L26).

Issue #2227, opened 2025-01-10
XMLCodec.DecodeValue always returns nil


lovromazgon:
**Describe the bug**

When trying to decode an XML value into a Go type, the driver always returns `nil`, because it tries to unmarshal the value into an `any` typed value in [this line](https://github.com/jackc/pgx/blob/c2175fe46e3d6f43af14a21b47386739d15e4ee0/pgtype/xml.go#L196). According to the documentation, [`xml.Unmarshal`](https://pkg.go.dev/encoding/xml#Unmarshal) will discard data that doesn't fit into the supplied value:

```
Unmarshal parses the XML-encoded data and stores the result in the value pointed to by v, 
which must be an arbitrary struct, slice, or string. 
Well-formed data that does not fit into v is discarded.
```

**To Reproduce**

Runnable example: https://go.dev/play/p/yLdFqZPaoUo

```go
package main

import (
	"encoding/xml"
	"fmt"
)

func main() {
	bxml := []byte(`<root><key1>v1</key1><key2>v2</key2></root>`)

	var vxml any
	if err := xml.Unmarshal(bxml, &vxml); err != nil {
		fmt.Println(err)
	}
	
	// vxml is always nil
	fmt.Printf("xml value: %v, type: %T\n", vxml, vxml)
}
```

**Expected behavior**

Probably that raw XML bytes are returned.

**Actual behavior**

The returned value contains `nil`.

**Version**
 - pgx: v5.7.0+

**Additional context**

Discovered by @lyuboxa https://github.com/jackc/pgx/pull/2083#discussion_r1755768269.

felix-roehrich:
Please provide an example using pgx. Your example is not equivalent to the line you highlighted and the line itself is correct.

lyuboxa:
> Please provide an example using pgx. Your example is not equivalent to the line you highlighted and the line itself is correct.

https://github.com/jackc/pgx/blob/c2175fe46e3d6f43af14a21b47386739d15e4ee0/pgtype/xml.go#L195-L196

lovromazgon:
There you go.

```go
package main

import (
	"encoding/xml"
	"fmt"

	"github.com/jackc/pgx/v5/pgtype"
)

func main() {
	codec := &pgtype.XMLCodec{Marshal: xml.Marshal, Unmarshal: xml.Unmarshal}
	out, err := codec.DecodeValue(
		pgtype.NewMap(),
		uint32(pgtype.XMLOID),
		0,
[...]

PR #2228, opened 2025-01-11
XMLCodec: fix DecodeValue to return a []byte

jackc:
Previously, DecodeValue would always return nil with the default Unmarshal function.

fixes https://github.com/jackc/pgx/issues/2227


felix-roehrich:
While this does fix this specific issue, it no longer respects a custom `Unmarshal`. Thus it is backward incompatible, i.e. a breaking change. While I don't believe that somebody ever relied on this, I think the proper change would be to update the default codec.

felix-roehrich (approved):
Thanks!

nickcruess-soda (approved):
Makes perfect sense to me, nice fix.

lyuboxa:
@jackc 🙇

PR #2230, opened 2025-01-14
Add support for field name based composite type scanning

WhiskeyJack96:
This is pretty rough, but before going through the process to polish it and test it I wanted to get something up to see if this would be accepted (or if theres a "better" way)

### DIFF
--- a/pgtype/builtin_wrappers.go
+++ b/pgtype/builtin_wrappers.go
@@ -743,6 +743,25 @@ func (w *ptrStructWrapper) ScanIndex(i int) any {
 	return w.exportedFields[i].Addr().Interface()
 }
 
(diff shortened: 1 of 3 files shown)


jackc:
The fundamental difficulty with scanning by names in composite types is that composite field names are not included over the wire. They are sent positionally. So I don't think it is possible for `row(...)` or subquery results. I guess it might be possible for specific registered types to use the OID to look up the previously determined field names. I'm guessing that's the approach you are taking? I looked at the code but I didn't quite follow it.

I suppose this might be a reasonable approach to solve the problem. But I actually wonder if the problem needs to be solved. Composites are always returned in a specifically defined order. I just use a struct defined where the position of the fields matches the position of the fields on the PostgreSQL type. I haven't encountered the need for anything past that.

Issue #1691, opened 2023-07-16
Not all Json null case is handled


bat-bold:
**Describe the bug**

```select 'null'::jsonb``` 

Starting from v5.3.0, in some case, it does not `Scan` as `Nil` .
v5.2.0 works.

**To Reproduce**

```go
func TestJSONBCodecUnmarshalSQLNull(t *testing.T) {
	defaultConnTestRunner.RunTest(context.Background(), t, func(ctx context.Context, t testing.TB, conn *pgx.Conn) {
		// Slices are nilified
		slice := []string{"foo", "bar", "baz"}
		err := conn.QueryRow(ctx, "select null::jsonb").Scan(&slice)
		require.NoError(t, err)
		require.Nil(t, slice)

                sliceJsonNull := []string{"foo1", "bar1", "baz1"}
		err = conn.QueryRow(ctx, "select 'null'::jsonb").Scan(&sliceJsonNull)
		require.NoError(t, err)
/*Pass*/	require.Nil(t, sliceJsonNull)  //// Pass

		// Maps are nilified
		m := map[string]any{"foo": "bar"}
		err = conn.QueryRow(ctx, "select null::jsonb").Scan(&m)
		require.NoError(t, err)
		require.Nil(t, m)

                mJsonNull := map[string]any{"foo": "bar"}
		err = conn.QueryRow(ctx, "select 'null'::jsonb").Scan(&mJsonNull)
		require.NoError(t, err)
/*Pass*/	require.Nil(t, mJsonNull)  //// Pass

		m = map[string]interface{}{"foo": "bar"}
		err = conn.QueryRow(ctx, "select null::jsonb").Scan(&m)
		require.NoError(t, err)
		require.Nil(t, m)

                mJsonNull = map[string]interface{}{"foo": "bar"}
		err = conn.QueryRow(ctx, "select 'null'::jsonb").Scan(&m)
		require.NoError(t, err)
/*Pass*/	require.Nil(t, m)  //// Pass

		// Pointer to pointer are nilified
		n := 42
		p := &n
		err = conn.QueryRow(ctx, "select null::jsonb").Scan(&p)
		require.NoError(t, err)
		require.Nil(t, p)

                n1 := 44
		p1 := &n1
		err = conn.QueryRow(ctx, "select 'null'::jsonb").Scan(&p1)
		require.NoError(t, err)
/*Fail*/	require.Nil(t, p1)  // *p1 == 0

                type ImageSize struct {
			Height int    `json:"height"`
[...]

Issue #1470, opened 2023-01-15
Migrate jackc/pgtype CIDR to its pgx/v5/pgtype alternative


antonprokopovich:
I'm trying to migrate from `pgx/v4` to `pgx/v5` and in particular I need to rewrite code that uses `CIDR` type from `jackc/pgtype` to it's v5 alternative. I see `v5/pgtype` has the `InetCodec` type but I can't figure out how to use it instead of CIDR from v4.

Here the code I need to rewrite:

```
package main

import (
	"net"

	pgtypeV4 "github.com/jackc/pgtype" // TODO: Migrate to "github.com/jackc/pgx/v5/pgtype"
)

func toPgCIDRs(a []*net.IPNet) pgtypeV4.CIDRArray {
	elements := make([]pgtypeV4.CIDR, len(a))
	for i, e := range a {
		elements[i] = pgtypeV4.CIDR(pgtypeV4.Inet{IPNet: e, Status: pgtypeV4.Present})
	}

	return pgtypeV4.CIDRArray{
		Status:     pgtypeV4.Present,
		Elements:   elements,
		Dimensions: []pgtypeV4.ArrayDimension{{Length: int32(len(a)), LowerBound: 0}},
	}
}
```

Any kind of help would be greatly appreciated!

jackc:
You should be able to use the core Go types directly now. There isn't a pgx CIDR type because it shouldn't be necessary anymore. You probably want https://pkg.go.dev/net/netip#Prefix.

antonprokopovich:
@jackc Thank you for your reply! I’ll try it out.

May I ask just one more questing regarding pgtype V5 vs V4?

I have code that queries a `jsonb` filed from BD into a *string field of a struct. It worked fine with V4 but after migrating to V5 I get `scan row into struct fields: can't scan into dest[7]: json: cannot unmarshal object into Go value of type string`. 

I understand that it tries to umarshal the json and if I use `map[string]interface` instead of `*string` then error goes away. But I would like to keep using `*string` type as before and can’t figure out how to fix it.

jackc:
Can you make an example program that demonstrates this?

antonprokopovich:
[...]

PR #2236, opened 2025-01-21
Alternative implementation for JSONCodec.PlanScan

### LINKED ISSUE #2229: Potential regression in `sql.Scanner` handling when using `pgx.CollectRows`
**Describe the bug**

As of `v5.7.2` When using `pgx.CollectRows` to collect rows into a struct that contains a pointer to a struct that implements `sql.Scanner`, if the column in the database is `NULL`, the corresponding struct is no longer `nil`.

* `v5.7.1` - the example below prints `testData[1].Value was nil struct` and the `testData[1].Value` is `nil` as expected.
* `v5.7.2` - the example below prints `expected testData[1].Value to be nil struct` and the `testData[1].Value` is no longer nil.

This seems to have been introduced with https://github.com/jackc/pgx/pull/2151, is this an expected change? 
Is it possible to work around this to ensure that the pointer values are still nil?

**To Reproduce**

If possible, please provide runnable example such as:

```
package main

import (
	"context"
	"database/sql/driver"
	"encoding/json"
	"errors"
	"fmt"
	"log/slog"
	"os"

	embeddedpostgres "github.com/fergusstrange/embedded-postgres"
	"github.com/jackc/pgx/v5"
)

type TestData struct {
	Id    string     `db:"id"`
	Value *TestValue `db:"value"`
}

func (e *TestData) RewriteQuery(ctx context.Context, conn *pgx.Conn, sql string, args []any) (newSQL string, newArgs []any, err error) {
	return `
		INSERT INTO
			pgx_test (
				id,
				value
			)
		VALUES
			($1, $2)
	`, []any{e.Id, e.Value}, nil
}

type TestValue struct {
	Data string `json:"data"`
}

func (t *TestValue) Scan(src interface{}) error {
	if src == nil {
		return nil
	}

	var source []byte
	switch val := src.(type) {
	case string:
		source = []byte(val)
	case []byte:
		source = val
	default:
		return errors.New("incompatible source type")
	}

	if err := json.Unmarshal(source, &t); err != nil {
		return fmt.Errorf("unable to unmarshal source data: %w", err)
	}

	return nil
}

func (t *TestValue) Value() (driver.Value, error) {
	if t == nil {
		return nil, nil
	}

	testValueBytes, err := json.Marshal(t)
	if err != nil {
		return nil, fmt.Errorf("unable to marshal test value data: %w", err)
	}

	return testValueBytes, nil
}

func main() {
	ctx := context.TODO()

	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
	slog.SetDefault(logger)

	postgres := embeddedpostgres.NewDatabase(embeddedpostgres.DefaultConfig().Port(5555).Logger(nil))
	if err := postgres.Start(); err != nil {
		slog.Error("unable to start embedded postgres", "error", err.Error())
		os.Exit(1)
	}
	defer func() {
		if err := postgres.Stop(); err != nil {
			slog.Error("unable to stop embedded postgres", "error", err.Error())
			os.Exit(1)
		}
	}()

	test(ctx)

}

func test(ctx context.Context) {

	connectionUrl := "postgres://postgres:postgres@127.0.0.1:5555/postgres"

	pgxConfig, err := pgxpool.ParseConfig(connectionUrl)
	if err != nil {
		slog.Error("cannot parse postgres config", "error", err.Error())
		return
	}

	pool, err := pgxpool.NewWithConfig(ctx, pgxConfig)
	if err != nil {
		slog.Error("unable to create pgx connection pool", "error", err.Error())
		return
	}

	if _, err := pool.Exec(ctx, `CREATE TABLE "public"."pgx_test" ("id" text NOT NULL PRIMARY KEY, "value" JSONB);`); err != nil {
		slog.Error("unable to create table 'pgx_test'", "error", err.Error())
		return
	}

	first := &TestData{
		Id: "first",
		Value: &TestValue{
			Data: "test",
		},
	}

	if _, err := pool.Exec(ctx, "", first); err != nil {
		slog.Error("unable to insert first value into table 'pgx_test'", "error", err.Error())
		return
	}

	second := &TestData{
		Id:    "second",
		Value: nil,
	}

	if _, err := pool.Exec(ctx, "", second); err != nil {
		slog.Error("unable to insert second value into table 'pgx_test'", "error", err.Error())
		return
	}

	rows, err := pool.Query(ctx, "SELECT id, value from pgx_test")
	if err != nil {
		slog.Error("unable to query rows from table 'pgx_test'", "error", err.Error())
		return
	}

	testData, err := pgx.CollectRows(rows, pgx.RowToAddrOfStructByName[TestData])
	if err != nil {
		slog.Error("unable to collect rows from table 'pgx_test'", "error", err.Error())
		return
	}

	if testData[0].Value == nil || testData[0].Value.Data != "test" {
		slog.Error("expected testData[0].Value.Data to be 'test'")
		return
	}
	slog.Info("testData[0].Value.Data was 'test'")

	if testData[1].Value != nil {
		slog.Error("expected testData[1].Value to be nil struct")
		return
	}

	slog.Info("testData[1].Value was nil struct")
}
```

*  Previous behaviour - `go mod tidy && go get github.com/jackc/pgx/v5@v5.7.1 && go run main.go`
* New behaviour - `go mod tidy && go get github.com/jackc/pgx/v5@v5.7.2 && go run main.go`

**Expected behavior**
I expected the behaviour to continue between versions, and that the `NULL` column would map to a `nil` struct value.

**Actual behavior**
 The `NULL` column no longer produces a `nil` struct value.

**Version**
 - Go: `$ go version` -> go version go1.23.1 darwin/arm64

**Additional context**

felix-roehrich:
The fix from #2151 just pushed the original problem from #1470 and #1691 further along. This should be a comprehensive solution; also resolves #2229.

### DIFF
--- a/pgtype/json.go
+++ b/pgtype/json.go
@@ -71,6 +71,27 @@ func (c *JSONCodec) PlanEncode(m *Map, oid uint32, format int16, value any) Enco
 	}
 }
 
+// JSON needs its on scan plan for pointers to handle 'null'::json(b).
+// Consider making pointerPointerScanPlan more flexible in the future.
+type jsonPointerScanPlan struct {
+	next ScanPlan
+}
+
+func (p jsonPointerScanPlan) Scan(src []byte, dst any) error {
+	el := reflect.ValueOf(dst).Elem()
+	if src == nil || string(src) == "null" {
+		el.SetZero()
+		return nil
+	}
+
+	el.Set(reflect.New(el.Type().Elem()))
+	if p.next != nil {
+		return p.next.Scan(src, el.Interface())
+	}
+
+	return nil
+}
+
 type encodePlanJSONCodecEitherFormatString struct{}
 
 func (encodePlanJSONCodecEitherFormatString) Encode(value any, buf []byte) (newBuf []byte, err error) {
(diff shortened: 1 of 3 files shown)


felix-roehrich:
I will add a test to prevent regression from #2229 tomorrow.

felix-roehrich:
@jackc Looking at the following examples, I believe that the current scanning behavior is in inconsistent.

```go
type Foo struct{}

var dst *Foo
err = conn.QueryRow(context.Background(), `select 'null'::jsonb`).Scan(&dst)
// err == nil
// dst == nil

var dst *Foo
err = conn.QueryRow(context.Background(), `select null::jsonb`).Scan(&dst)
// err == nil
// dst == nil

var dst Foo
err = conn.QueryRow(context.Background(), `select 'null'::jsonb`).Scan(&dst)
// err == nil
// dst == Foo{}

var dst Foo
err = conn.QueryRow(context.Background(), `select null::jsonb`).Scan(&dst)
// err == can't scan into dest[0] (col: jsonb): cannot scan NULL into *pgx_test.Foo
// dst == Foo{}
```

I believe either both cases of scanning `Foo` should error or none (or change scanning of `*Foo`). What is your opinion?

felix-roehrich:
I decided to revert #2151 partially. This is because this should never be necessary, as such cases would be handled by the `PointerPointerScanPlan`. JSON requires special care to handle the scanning of `'null'::jsonb`.

jackc:
Thanks! `sql.Scanner` especially plus json has been such a pain. Hopefully, this finally resolves it.

emreisikligil:
Can we add this to the next patch version v5.7.3 please as v5.7.2 has the regression when scanning null jsonb columns into a pointer?

jackc:
@emreisikligil Just released v5.7.3.

emreisikligil:
@jackc Thanks a lot 🙏

alicebob:
Just as a FYI, this did make the behavior of scanning a raw message with the content `null`. If I pass in a jSON value `null`, then I don't get back what I put into it, but a NULL/nil. I don't know the whole history of this change, and I'm sure it's needed somehow, so I put this comment here as a warning/gotcha, maybe it helps someone:

```
    ["CREATE TABLE pgx (jsoncol JSONB NOT NULL DEFAULT 'null')"]
    ["INSERT INTO pgx (jsoncol) VALUES ('null')"]
    var j json.RawMessage
    require.NoError(t, db.QueryRow(ctx, `SELECT jsoncol FROM pgx`).Scan(&j))
    require.Equal(t, json.RawMessage([]byte("null")), j)
```

this now fails in 5.7.3 with:
```
        	Error:      	Not equal: 
        	            	expected: json.RawMessage{0x6e, 0x75, 0x6c, 0x6c}
        	            	actual  : json.RawMessage(nil)
```

The test passes on 5.7.2.

pgtype/json.go
```diff
@@ -202,7 +197,7 @@ type scanPlanJSONToJSONUnmarshal struct {
 }
 
 func (s *scanPlanJSONToJSONUnmarshal) Scan(src []byte, dst any) error {
-	if src == nil {
+	if src == nil || string(src) == "null" {
```
felix-roehrich, line 200: This is due to this change. Which was made with https://github.com/jackc/pgx/pull/2236#issuecomment-2604154556 in mind, however looking at the godoc for the json package, not erroring is intentional. Though I personally don't agree with the rational, it's probably for the best to revert this change.

> // The JSON null value unmarshals into an interface, map, pointer, or slice
// by setting that Go value to nil. Because null is often used in JSON to mean
// “not present,” unmarshaling a JSON null into any other Go type has no effect
// on the value and produces no error.

jackc: I agree. I think this should be reverted.

alicebob: Thanks both for the quick reply, and many more thanks for a great library.

PR #2240, opened 2025-01-24
Unwatch and close connection on a batch write error

bonnefoa:
On a write error, a conn.Write simply unlock pgconn, leaving the connection as Idle and reusable while the multiResultReader would be closed. From this state, calling multiResultReader.Close won't try to receiveMessage and thus won't unwatch and close the connection since it is already closed. This leaves the connection "open" and the next time it's used, a "Watch already in progress" panic could be triggered.

This patch fixes the issue by unwatching and closing the connection on a batch write error, similar to what's done on a receiveMessage error. The same was done on Sync.Encode error even if the path is unreachable as Sync.Error never returns an error.


jackc:
Thanks!

PR #2257, opened 2025-02-17
Change connection logic to be more forgiving

### LINKED ISSUE #1581: No retry when pg_hba.conf error when using sslmode=allow or prefer
in postgres [Documentation](https://www.postgresql.org/docs/)
```
sslmode
This option determines whether or with what priority a secure SSL TCP/IP connection will be negotiated with the server. There are six modes:

disable
only try a non-SSL connection

allow
first try a non-SSL connection; if that fails, try an SSL connection

prefer (default)
first try an SSL connection; if that fails, try a non-SSL connection
```
VERSION：
I'm using pgx/v4 v4.16.1

Detail：
When using sslmode=allow to establish a connection, according to the documentation, it should first try to establish a non-ssl connection and then try to establish an ssl connection after failure.
In the implementation, the ConnectConfig function will handle some special errors to skip subsequent retries.

```
       for _, fc := range fallbackConfigs {
		pgConn, err = connect(ctx, config, fc)
		if err == nil {
			break
		} else if pgerr, ok := err.(*PgError); ok {
			err = &connectError{config: config, msg: "server error", err: pgerr}
			const ERRCODE_INVALID_PASSWORD = "28P01"                    // wrong password
			const ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION = "28000" // wrong password or bad pg_hba.conf settings
			const ERRCODE_INVALID_CATALOG_NAME = "3D000"                // db does not exist
			const ERRCODE_INSUFFICIENT_PRIVILEGE = "42501"              // missing connect privilege
			if pgerr.Code == ERRCODE_INVALID_PASSWORD ||
				pgerr.Code == ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION ||
				pgerr.Code == ERRCODE_INVALID_CATALOG_NAME ||
				pgerr.Code == ERRCODE_INSUFFICIENT_PRIVILEGE {
				break
			}
		}
	}
```
When the previous attempt triggers a bad pg_hba.conf settings error, there will be no subsequent retries, but in actual use, occasionally it is necessary to set pg_hba.conf to block all non-ssl connections, and the client in allow mode will be unusable at this time
Is there any way that can be used in this scenario?Or should it not break on ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION error?

### LINKED ISSUE #2246: Problem with retry/fallback to simplier auth from LDAP (ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION)
**Describe the bug**
Hello everyone find this [issue](https://github.com/jackc/pgx/issues/1581). When I was trying to understand why this driver behave differently from what we have in jdbc, as I understand libpq should have same behaviour as menssioned above. The problem that this ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION should make a fallback, because when you connect via LDAP and configure LDAP settings in pg_hba.conf Go driver doesn't fallback whereas other do. Can we fix it? I want to have default behavior when I use driver.

The error: server error: FATAL: LDAP authentication failed for user "postgres" (SQLSTATE 28000)

**Expected behavior**
fallback without auth to LDAP, should work the same as in the official C driver

felix-roehrich:
See discussion in #2246.

Closes #2246.
Closes #1581.

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -268,12 +268,15 @@ func connectPreferred(ctx context.Context, config *Config, connectOneConfigs []*
 
 		var pgErr *PgError
 		if errors.As(err, &pgErr) {
-			const ERRCODE_INVALID_PASSWORD = "28P01"                    // wrong password
-			const ERRCODE_INVALID_AUTHORIZATION_SPECIFICATION = "28000" // wrong password or bad pg_hba.conf settings
-			const ERRCODE_INVALID_CATALOG_NAME = "3D000"                // db does not exist
-			const ERRCODE_INSUFFICIENT_PRIVILEGE = "42501"              // missing connect privilege
+			// pgx will try next host even if libpq does not in certain cases (see #2246)
+			// consider change for the next major version
+
+			const ERRCODE_INVALID_PASSWORD = "28P01"
+			const ERRCODE_INVALID_CATALOG_NAME = "3D000"   // db does not exist
+			const ERRCODE_INSUFFICIENT_PRIVILEGE = "42501" // missing connect privilege
+
+			// auth failed due to invalid password, db does not exist or user has no permission
 			if pgErr.Code == ERRCODE_INVALID_PASSWORD ||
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

Issue #1066, opened 2021-08-16
pgx.Batch behavior when one statement raises an exception


jschaf:
Hello again, I'd like to clarify the expected behavior of a sequence of pgx batch calls where one of the statements throws an exception in two cases: 1) with explicit transaction commands, 2) with implicit transaction commands.

First case with explicit transactions:
```go
		batch.Queue(`BEGIN`)
		batch.Queue(`INSERT INTO foo (id) VALUES (1)`)
		batch.Queue(`COMMIT`)
		batch.Queue(`SELECT 1/0`) // exception
		batch.Queue(`INSERT INTO foo (id) VALUES (2)`)
```

Second case with implicit transactions:
```go
		batch.Queue(`INSERT INTO foo (id) VALUES (1)`)
		batch.Queue(`SELECT 1/0`) // exception
		batch.Queue(`INSERT INTO foo (id) VALUES (2)`)
```

I wrote up tests with the above commands and found that:

- Execution continues until the first exception, the `SELECT 1/0`. The last insert statement does not run in either case.

- In the first case, the first insert with value `1` inserts successfully since there's an explicit `COMMIT`.
 
- In the second case with implicit transactions, the first insert is rolled back, so nothing is written into the `foo` table.

The Postgres docs for error handling for the extended query protocol states:

> When an error is detected while processing any extended-query message, the backend issues ErrorResponse, then reads and discards messages until a Sync is reached, then issues ReadyForQuery and returns to normal message processing. (But note that no skipping occurs if an error is detected while processing Sync — this ensures that there is one and only one ReadyForQuery sent for each Sync.

I'm not sure how a Sync relates to pgx. My guess is pgx only issues a Sync after sending all queries in a batch.

To summarize:

1. Is a batch wrapped with an implicit transaction block if there's no explicit transaction commands like `BEGIN`?
2. If a statement in a batch fails, will pgx execute any more messages in that batch?

jackc:
> My guess is pgx only issues a Sync after sending all queries in a batch.

Correct.

> Is a batch wrapped with an implicit transaction block if there's no explicit transaction commands like BEGIN?

Correct.

> If a statement in a batch fails, will pgx execute any more messages in that batch?

pgx sends them all. But PostgreSQL will ignore them until it gets the `Sync` message.

vorishirne:
>Is a batch wrapped with an implicit transaction block if there's no explicit transaction commands like BEGIN?

[...]

PR #2266, opened 2025-02-26
added BatchExec and BatchExecTx func

nar10z:
Hey, everybody!
I decided to add 2 new functions to simplify working with pgx.Batch{}. 
One can be executed with both pgx.Conn and pgx.Tx. The second one is executed only with pgx.Conn, but specifies the transaction explicitly.

### DIFF
--- a/batch_exec.go
+++ b/batch_exec.go
@@ -1,0 +1,88 @@
+package pgx
+
+import (
+	"context"
+	"errors"
+)
+
+var (
+	ErrNoSupportTxInBatchExecTx = errors.New("no support tx in batch exec transaction")
+)
+
+// BatchExec Executes queries using pgx.Batch without using an explicit transaction
+// Example:
+//
+//	pgx.BatchExec(ctx, conn, func (b *pgx.Batch){
+//		batcher.Queue("INSERT INTO users (name, email) VALUES ($1, $2)", "Alice", "alice@example.com")
+//		batcher.Queue("INSERT INTO users (name, email) VALUES ($1, $2)", "Bob", "bob@example.com")
+//	})
+//
+// OR:
+//
+//	tx, err := conn.Begin(ctx)
+//	if err != nil {
+//		panic(err)
+//	}
+//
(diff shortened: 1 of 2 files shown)


jackc:
I don't understand how this should be used.

BatchExec appears to do the same thing as:

```go
err := SendBatch(batch).Close()
```

And batches are already implicitly transactional, so I'm not sure when BatchExecTx would be used.

nar10z:
> And batches are already implicitly transactional, so I'm not sure when BatchExecTx would be used.

So if I execute the code below, then the whole batch request will not be executed? Not even the first request?

```go
b := &pgx.Batch{}

b.Queue("INSERT INTO t1 (id, name) VALUES ($1, $2);", 1, "some")
b.Queue("SOME ERROR")
b.Queue("INSERT INTO t2 (id, name) VALUES ($1, $2);", 2, "some")

err := conn.SendBatch(ctx, b).Close()
if err != nil {
  return err
}
```

nar10z:
And is there then a difference between example 1 and example 2?

```go
b := &pgx.Batch{}

// Example 1
b.Queue("BEGIN;")
b.Queue("INSERT INTO t1 (id, name) VALUES ($1, $2);", 1, "some")
b.Queue("INSERT INTO t2 (id, name) VALUES ($1, $2);", 2, "some")
b.Queue("COMMIT;")

// Example 2
b.Queue("INSERT INTO t1 (id, name) VALUES ($1, $2);", 1, "some")
b.Queue("INSERT INTO t2 (id, name) VALUES ($1, $2);", 2, "some")

err := conn.SendBatch(ctx, b).Close()
if err != nil {
	return err
}
```

jackc:
In your case with an error in the middle will be entirely rolled back.

And there is no effective difference between your cases with explicit `BEGIN` and `END` statements.

These behaviors are due to the PostgreSQL extended protocol treating all queries between `Sync` messages as implicitly transactional.

nar10z:
https://github.com/jackc/pgx/issues/1066

PR #2269, opened 2025-03-04
Fix args memory leak in rows.Close()

hrubymar10:
Hello,
I have found the memory leak in rows.Close() method. Please check it out.

The code for reproduction is available here: https://github.com/hrubymar10/pgx-memory-leak-test/tree/master

heap pprof before fix:
Screenshot 2025-03-04 at 11 28 20

heap pprof after fix:
Screenshot 2025-03-04 at 11 36 50

pprof files:
[heap.pb.zip](https://github.com/user-attachments/files/19068991/heap.pb.zip)
[heap_fixed.pb.zip](https://github.com/user-attachments/files/19068989/heap_fixed.pb.zip)

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -188,6 +188,8 @@ func (rows *baseRows) Close() {
 	} else if rows.queryTracer != nil {
 		rows.queryTracer.TraceQueryEnd(rows.ctx, rows.conn, TraceQueryEndData{rows.commandTag, rows.err})
 	}
+
+	rows.args = nil
 }
 
 func (rows *baseRows) CommandTag() pgconn.CommandTag {


jackc:
If setting rows.args to nil allows it to be GCed, then that implies that rows *cannot* be GCed for some reason. That would seem to be the ultimate issue.

It looks like the preallocated buffer of poolRows in pgxpool.connResource. That saves a alloc per query, but it pins the underlying row until the preallocated buffer is GCed.

hrubymar10:
Hi @jackc , any update on this? I understand that there is maybe bigger issue behind it but this solution is currently good enough. This issue really blocks us as we are often using params with slices of huge capacity.

jackc:
I was considering whether the actual fix should be made in pgxpool, but after reviewing it further, I decided to expand on your suggestion. In particular, I also clear row.values as otherwise the last row read could also be pinned in memory longer than desirable. See 07871c0a346cdcabfa0e39996b00557665a3b56c.

hrubymar10:
Hi @jackc, thanks for fixing this issue in the master codebase. How the release process works for pgx? When do you expect to create new tag containing those changes? Thx 🙏

jackc:
@hrubymar10 No formal process. Just whenever there are a critical fix or a reasonable quantity of fixes. As it turns out, I fixed a long-standing bug today, so that triggered v5.7.5.

PR #2270, opened 2025-03-05
CVE-2025-22869 golang.org/x/crypto bump version

irlndts:
Hi, there is a vulnerability [CVE-2025-22869](https://pkg.go.dev/vuln/GO-2025-3487)

I know pgx actually doesn't use the vulnerable functionality in golang.org/x/crypto but scanner alert still works.

I've done followings:

- Update go.mod to bump crypto version to v0.35.0
- Update go version to 1.24 since crypto v0.35.0 requires it
- Then go mod tidy
- Then go test ./... (everything is fine)


jackc:
Are you sure about the Go 1.24 requirement? It looks like 1.23 is sufficient to me. https://cs.opensource.google/go/x/crypto/+/master:go.mod

I'd like to avoid that change as that would break pgx's Go compatible versions policy.

nrvnrvn:
> Before Go 1.21, Go toolchains treated the go line as an advisory requirement: if builds succeeded the toolchain assumed everything worked, and if not it printed a note about the potential version mismatch. Go 1.21 changed the go line to be a mandatory requirement instead. 

> A module’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in require statements.

Source: https://go.dev/doc/toolchain


Don't change the `go` directive in the `go.mod` file manually. Do this instead:

```sh
$ go get golang.org/x/crypto
go: upgraded go 1.21 => 1.23.0
go: added toolchain go1.24.2
go: upgraded golang.org/x/crypto v0.31.0 => v0.37.0
go: upgraded golang.org/x/sync v0.10.0 => v0.13.0
go: upgraded golang.org/x/text v0.21.0 => v0.24.0
```

`go get` will happily bump the go version in the go.mod file to the highest possible based on the module dependencies.

jackc:
I updated this dependency with the proper Go version changes in c92d0a90458bd137ac40e45316919d9d61576472.

markus-gerstel-sp:
Just as a heads up: As of golang 1.24 [`golang.org/x/crypto/pbkdf2`](https://pkg.go.dev/golang.org/x/crypto@v0.37.0/pbkdf2) has migrated into the standard library (-> [`crypto/pbkdf2`](https://pkg.go.dev/crypto/pbkdf2@go1.24.0)), so if/when/once it is time to move to 1.24 as minimum supported version you can get rid of the `golang.org/x/crypto` dependency entirely.

PR #2279, opened 2025-03-15
Add MinIdleConns

### LINKED ISSUE #2278: Feature request: MinIdleConns

Right now pgxpool provides MinConns, which addresses the situation where you want your application to be able to always have at least some connections ready for when a new request comes in, if you care about tail latencies and don't want to sometimes introduces a DB connection in the middle of trying to serve the request (which matters when you have a remote DB and TLS is involved, which can sometimes be hundreds of milliseconds).

However, this does not address the situation that you have a fairly active server, so may have a good number of connections open at peak time, so MinConns does not kick in. But it may not have many idle connections when handling its traffic. This exposes one to the situation of potentially not having connections available to handle new requests again.

Of course one could set MinConns so high that it is above both your off-peak and on-peak usage, but that would be a waste of resources.

I believe something like `MinIdleConns` would solve this. The pool could always try to maintain at least `MinIdleConns` extra connections on top of the current amount of in-use connections. This would mean the application is always prepared to handle a burst of `MinIdleConns` simultaneous requests on top of whatever load it already has. Since this would be configurable, it would allow application owners to adjust it for whatever amount of burst traffic they want to be ready for.

**Additional context**
I'm happy to try and implement this if I can get an OK from a maintainer that it seems like a reasonable architecture and I'm not missing something.

djahandarie:
Resolves #2278. (It was a quick implementation so I just went ahead and did it while my attention was here.)

### DIFF
--- a/pgxpool/common_test.go
+++ b/pgxpool/common_test.go
@@ -147,6 +147,7 @@ func assertConfigsEqual(t *testing.T, expected, actual *pgxpool.Config, testName
 	assert.Equalf(t, expected.MaxConnIdleTime, actual.MaxConnIdleTime, "%s - MaxConnIdleTime", testName)
 	assert.Equalf(t, expected.MaxConns, actual.MaxConns, "%s - MaxConns", testName)
 	assert.Equalf(t, expected.MinConns, actual.MinConns, "%s - MinConns", testName)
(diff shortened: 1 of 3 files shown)


jackc:
LGTM. Maybe we can get another set of eyes on it before merging though.

hmh:
The logic in the patch seems sound, but I didn't try to compile or test the change.

jackc:
Thanks!

PR #2293, opened 2025-03-30
feat: add support for direct sslnegotiation

divyam234:
Add support for direct ssl negotiation which was introduced in postgres 17.

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION


divyam234:
@jackc Added pgmock tests see if its fine

PR #2303, opened 2025-04-20
feat: better collect rows generic parameters

ConsoleTVs:
Adds a better signature for the `CollectRows` generic parameters, on par with `AppendRows` that allows overriding the actual return slice. Useful for custom slice types, such as:

```go
type User struct{}
type Users []User

// ...

// old:
// CollectRows[User] -> []User

// new:
// CollectRows[User, Users] -> Users
```

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -447,8 +447,8 @@ func AppendRows[T any, S ~[]T](slice S, rows Rows, fn RowToFunc[T]) (S, error) {
 // CollectRows iterates through rows, calling fn for each row, and collecting the results into a slice of T.
 //
 // This function closes the rows automatically on return.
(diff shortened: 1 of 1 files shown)


jackc:
Do I understand correctly that this is to save a type cast or is it doing more? Without tests its a bit hard to follow the intent.

ConsoleTVs:
Oh yes, it's pure type gimmicks, there's no new logic involved.

jackc:
As far as I can tell you can do the same by assigning the result of `CollectRows` with `=` to a variable of your desired type rather than using `:=` to create it. It saves one line of code (or two if err needs to be declared).

I'm somewhat inclined against adding a public function to save a line of code on the basis of avoiding the increase in API surface unless it was going to be very frequently used. And anyone that is doing that type operation frequently can add this helper function to their project.

PR #2313, opened 2025-05-08
Implement AcquireTracer and ReleaseTracer for TraceLog

stampy88:
- `TraceLog` now implements the `pgxpool.AcquireTracer` and `pgxpool.ReleaseTracer` interfaces to log connection pool interactions.

### DIFF
--- a/tracelog/tracelog.go
+++ b/tracelog/tracelog.go
@@ -375,16 +375,16 @@ func (tl *TraceLog) TraceAcquireEnd(ctx context.Context, _ *pgxpool.Pool, data p
 	}
 
 	if data.Conn != nil {
-		if tl.shouldLog(LogLevelInfo) {
-			tl.log(ctx, data.Conn, LogLevelInfo, "Acquire", map[string]any{tl.Config.TimeKey: interval})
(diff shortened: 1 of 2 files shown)


jackc:
The code seems fine.

But this change could have a side effect of greatly increasing the amount of log output. If someone is mostly using `Query`, `QueryRow`, and `Exec` directly against a pool then the amount of output on the `info` level would triple. Instead of getting a single log entry for the query, they would get acquire, query, and release log entries.

I'm not sure this is a desirable outcome.

stampy88:
well, I think the additional logging would be useful if you don't have other visibility into your app. Knowing how long it took to acquire a connection is definitely helpful. Also, these log statements are pretty light; they are not like the `prepare`, `query`, or even `connect` in that it is really just timing/error information versus outputting SQL statements and args. One option would be to set this at the `debug` level instead of `info`.

jackc:
> Also, these log statements are pretty light; they are not like the prepare, query, or even connect in that it is really just timing/error information versus outputting SQL statements and args.

I'm just thinking about how the volume of messages could be greatly increased.

> One option would be to set this at the debug level instead of info.

Let's do that.

PR #2335, opened 2025-06-03
Attempt to acquire a connection MaxConns + 1 times before aborting

### LINKED ISSUE #2330: Acquire: Bail out after a certain number of failures
Related to #2328, while working on #2329 I noticed that the `Acquire` method uses an infinite loop when attempting to connect.

 I wonder if it would be acceptable to bail out after an (un)reasonable number of connection attempts (maybe 100? 1000?), with a descriptive error.  This could help users notice infinite loops more easily.

The immediate problem I encountered, which lead to an infinite loop, is already being addressed in #2328 and #2329. 

**Additional context**
N/A

flimzy:
Fixes #2330

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -2,6 +2,7 @@ package pgxpool
 
 import (
 	"context"
+	"errors"
 	"math/rand"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jackc:
Thanks!

PR #2338, opened 2025-06-13
readme: Fix typo in features

bersace:


jackc:
Maybe that could be worded better, but it's not a typo. The feature in question lets you handle NULLs by scanning into a **T rather than *T.

bersace:
@jackc thanks for the explanation. Changing the typography may help. Or just close until someone else confirm this need more clarity. I prefer brevity than long wording, even at the cost of some misinterpretation. It's just a feature list.

PR #2349, opened 2025-07-04
Make fewer allocations in QueryContext().

dominiquelefevre:
For a simple query like `SELECT * FROM t WHERE id = $1`, QueryContext() used to make 6 allocations on average. Reduce that to 4 by avoiding two allocations that
are immediately discarded:

  // the variadic arguments of c.conn.Query() escape,
  // so args is heap-allocated
  args := []any{databaseSQLResultFormats}
  // append() reallocates args to add enough space +
  // an array returned by namedValueToInterface()
  // is immediately discarded
  args = append(args, namedValueToInterface(argsV)...)


jackc:
LGTM

PR #2350, opened 2025-07-07
Add AcquireTimeout to Pool

typokign:
Adds a timeout for acquiring a free connection from the connection pool.

This existed in previous versions of PGX: https://github.com/jackc/pgx/blob/v2/conn_pool.go#L29.

This is useful for cases where queries may be slow, but acquiring a connection from the pool is expected to be fast, and it is desirable for requests to fail fast if the connection pool is starved. This is possible today by manually calling `Acquire()` for every query and passing a context with a shorter timeout, but there is no way to configure a separate timeout for `Acquire()` when using the `Query()`, `QueryRow()`, `Exec()`, etc. helper functions on Pool.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -94,6 +94,7 @@ type Pool struct {
 	maxConnLifetimeJitter time.Duration
 	maxConnIdleTime       time.Duration
 	healthCheckPeriod     time.Duration
+	acquireTimeout        time.Duration
 
 	healthCheckChan chan struct{}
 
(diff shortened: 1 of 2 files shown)


jackc:
I'm not opposed to this feature overall, but I have a few concerns about the implementation.

1. It appears that when the acquire timeout is triggered, the error returned will be a `context.DeadlineExceeded`. That means a caller can get a `context.DeadlineExceeded` error when the context they passed in is still valid. The ambiguity is unfortunate.
2. There's no way to override on a per call basis. Should be a rare occurrence, but the lack of symmetry is again unfortunate.
3. I could imagine that someone may want something similar like a default query timeout for the query execution itself. Something like a pgx side `statement_timeout`. Whatever approach is used it would be nice to have a pattern that works for all similar features.

Issue #2360, opened 2025-07-31
connection local statement cache does not correctly invalidate for batches


Jk1484:
**Describe the bug**
The issue is exactly what is described here: https://github.com/jackc/pgx/issues/841
But it is for batches. 

Closing rows returned by batch results wouldn't fix the issue because returned rows are already closed. Calling Close() on closed rows does not reach reach the part where it invalidates the cache for query.

This PR should fix the issue #2361 

**To Reproduce**

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
)

// run `docker run --name postgres-pgx-test -e POSTGRES_PASSWORD=q123 -e POSTGRES_DB=pgx -p 5435:5432 -d postgres:latest`
func main() {
	ctx := context.Background()

	db, err := pgx.Connect(ctx, "postgres://postgres:q123@localhost:5435/pgx?sslmode=disable")
	if err != nil {
		fmt.Fprintf(os.Stderr, "Unable to connect to database: %v\n", err)
		os.Exit(1)
	}
	defer db.Close(ctx)

	// Grab a single connection on which to perform the SELECT statements so we can be
	// sure to reproduce the bug in this toy example. The bug will be intermittent if we
	// instead just call the QueryContext method on the `*sql.DB`.
	_, err = db.Exec(ctx, `
        DROP TABLE IF EXISTS drop_cols;
        CREATE TABLE drop_cols (
            id SERIAL PRIMARY KEY NOT NULL,
            f1 int NOT NULL,
            f2 int NOT NULL
        );
    `)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("connected to db")

	// insert some data (probably not needed for replication)
	_, err = db.Exec(ctx, "INSERT INTO drop_cols (f1, f2) VALUES (1, 2)")
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("inserted values")

	getSQL := "SELECT * FROM drop_cols WHERE id = ANY($1)"

	// populate pgx's internal prepared statement cache
	rows, err := db.Query(ctx, getSQL, []int{1})
	if err != nil {
		log.Fatal(err)
	}
	rows.Close()
	fmt.Println("populated stmtcache")

	// change the schema (in a more realistic example, this would mean running a
	// migration while the application server is still running).
	_, err = db.Exec(ctx, "ALTER TABLE drop_cols DROP COLUMN f1")
	if err != nil {
[...]

PR #2361, opened 2025-07-31
invalidate cache on batch error for batchResults and pipelineBatchResults

Jk1484:
This PR should fix this issue #2360 

All the heavy lifting was done here #841 
But the fix for batches was missed.


batch.go
```diff
@@ -441,3 +445,20 @@ func (br *pipelineBatchResults) nextQueryAndArgs() (query string, args []any, er
 	br.qqIdx++
 	return bi.SQL, bi.Arguments, nil
 }
+
+// invalidates statement and description caches on batch results error
+func invalidateCachesOnBatchResultsError(conn *Conn, b *Batch, err *error) {
```
jackc, line 450: I don't understand why `err` is a `*error` instead of `error`.

Jk1484: If we used error, we would pass a copy of the error to the function, which would be nil and always be nil. We want to pass it as a pointer so that when the error is populated, we can access it, since we are deferring the function call. If we used error instead of *error, we would need to call the function before each return to make it work.

Issue #1738, opened 2023-09-17
Is it possible to disable the Ping that occurs when acquiring a connection in pgxpool?


db-instacart:
https://github.com/jackc/pgx/blob/b301530a5fd50fe8ee46149fec58162a52bed8c6/pgxpool/pool.go#L510-L516

I see it was added for:

```
Efficiently check if a connection has been closed before writing.
This reduces the cases where the application doesn't know if a query
that does a INSERT/UPDATE/DELETE was actually sent to the server or
not.
```

But, if it's only needed for non-reads then it seems we're doing extra pings for no reason.
Possibly noticeable for us as we have many pools configured, so likely connections can go idle for > 1 second.
Perhaps some config driven way to disable it?

jackc:
At the moment there is no way to configure or disable this.

I'm not opposed to changing this, but if something is to be done it should be more flexible than a simple switch. Probably a user supplied function that returns true or false to determine if the connection should be considered fresh or needs to ping.

db-instacart:
Sure, is it something you see living on the `Config` in `pool.go`?

jackc:
> Sure, is it something you see living on the Config in pool.go?

Yes.

db-instacart:
Hi Jack,

Was taking a look this morning at the code and had a question. What would you find useful to provide to the user to make the decision on whether or not they should do this liveness check? I'm not aware of everything that exists for use in the package, but right now it's using the `IdleDuration` from the `puddle.Resource`. Do you want to just provide the `puddle.Resource` similar to below?

```go

// On the Config
CheckLiveness func(*puddle.Resource[*ConnResource]) bool

// Default func to use in the ParseConfig
func checkLiveness(res *puddle.Resource[*ConnResource]) bool {
	return res.IdleDuration() > time.Second
}

// And the snippet in Acquire
cr := res.Value()
if p.checkLiveness(res) {
    err := cr.conn.Ping(ctx)
    if err != nil {
        res.Destroy()
        continue
   }
}
```

This would involve making `ConnResource` public so it can be used in the function parameter.

Or do you want to pass something else to their user provided function?

jackc:
Hmm... Good question. I'm not super eager to make `ConnResource` public. Though as you say, there isn't anything public that would give you the idle duration.

Maybe something like:

```go
type ShouldPingParams struct {
	Conn *pgx.Conn
	IdleDuration time.Duration
}

func ShouldPing(ctx context.Context, params ShouldPingParams) bool {
[...]

PR #2362, opened 2025-08-06
Support configurable ShouldPing func when acquiring conns

christiankiely:
Implemented the suggested approach in https://github.com/jackc/pgx/issues/1738.

Allowing ShouldPing to be configurable will allow us to change the threshold for Pings on acquiring a conn, or disable it entirely.


jackc:
Thanks!

ilyam8:
@jackc, hey, 👋 Could you please create a release that includes this change? This is a requirement for us to migrate from v4 to v5, as it resolves the PgBouncer compatibility issue that's currently blocking our upgrade.

jackc:
@ilyam8 v5.7.6 was just released.

PR #2367, opened 2025-08-15
Ensure zeronull int types implement Int64Scanner

zeghong:
Currently the `Int2`/`Int4`/`Int8` type in `zeronull` package does not implement `pgtype.Int64Scanner` interface correctly. 

This PR fixes the implementations but does not change the scanning behavior, since the `sql.Scanner` implementations are already in place.

Also correct error string when comparing minimum value.


jackc:
LGTM

PR #2372, opened 2025-08-20
Add note about QueuedQuery.Exec and BatchResults.Close

WGH-:
When reviewing my own old code, I noticed several monstrosities like this:

    batchResult := tx.SendBatch(ctx, batch)
    defer batchResult.Close()
    for range batch.Len() {
            if _, err := batchResult.Exec(); err != nil {
                    return err
            }
    }
    return nil

All of them can replaced with just this:

    return tx.SendBatch(ctx, batch).Close()

So I thought it might be a good idea to give a hint in the docs explicitly. This trick is not so apparent, after all.


jackc:
LGTM

Issue #1271, opened 2022-08-03
Sensitive connection string information part of parse error


Poweranimal:
The error `pgconn.parseConfigError` returned from `pgconn.ParseConfig` includes the full connection string in plain text (including the provided password).
Sensitive information, like the password, should not be visible in plaintext.

## Example:
```go
connString := "postgresql://postgres:Hello121\\@127.0.0.1:5432/postgres"
_, err := pgconn.ParseConfig(connString)
fmt.Println(err.Error())
```

Result:
```text
cannot parse `postgresql://postgres:xxxxxx@127.0.0.1:5432/postgres`: failed to parse as URL (parse "postgresql://postgres:Hello121\\@127.0.0.1:5432/postgres": net/url: invalid userinfo)
```

The above described incident can easily happen, if one passes the connection string via an environment variable with a password that is not url encoded.
e.g.:
```shell
POSTGRES_PASSWORD="not_\_url_/encoded_|password"
/my-service --conn-string "postgresql://postgres:${POSTGRES_PASSWORD}@127.0.0.1:5432/postgres"
```

jackc:
pgx already makes an effort but given that the problem is failing to parse the string it can't know for sure where the password is.

https://github.com/jackc/pgconn/blob/fe0fb3b24dca7c395038dc81bdb84e6706e50979/errors.go#L198-L221

The only way to be sure would be to not include any information from the conn string at all. We couldn't even expose the underlying error as the underlying parser may return the input string. e.g. `url.Parse()` errors do.

But I'd rather not do that, it's nice to have more than "parse failed" to go on when troubleshooting or debugging.

Maybe this should be the responsibility of whatever is calling `Connect()` or `ParseConfig()`.

Poweranimal:
What about making the `pgconn.parseConfigError` public with additional fields.
One field might be a detailed description of the failure and another one just a rough reason for the parse error.
Doing so, the caller of `pgconn.ParseConfig` can cast to the error type and only expose the rough error description via stdout or others to the public.

jackc:
> What about making the pgconn.parseConfigError public with additional fields.

That, or maybe a `RedactedError` that can wrap anything. `errors.Is` and `errors.As` would make it transparent to anything that cared to unwrap it.

> Regarding not knowing where the password is:
> Doesn't pgx already recognize this properly? It seems that the first part of the error message properly hides the password:

It *tries*. But it is error prone.

a-mesin:
[...]

PR #2376, opened 2025-08-26
ParseConfigError - redact password in connection string included in error message

phameratl:
Fixes https://github.com/jackc/pgx/issues/1271

Since `redactPW` can redact the password even if URL parse fails, we can remove the password also from the error string.


phameratl:
I did not see https://github.com/jackc/pgx/commit/f007d84675d68b9ce0177cce7700e5d6d097ee1e so I suppose this is not needed now.

PR #2378, opened 2025-09-08
fix: add check for pool is not nil

eGGshke:
If pool is nil, can return fatal error.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -577,6 +577,10 @@ func (p *Pool) createIdleResources(parentCtx context.Context, targetResources in
 
 // Acquire returns a connection (*Conn) from the Pool
 func (p *Pool) Acquire(ctx context.Context) (c *Conn, err error) {
+	if p == nil {
+		return nil, errors.New("pgxpool: Acquire called on nil Pool")
+	}
+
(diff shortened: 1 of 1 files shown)


jackc:
I don't think this is idiomatic Go. I have not seen explicit nil checks on objects that do not have specific nil behavior. I would expect a panic rather than a returned error as calling Acquire on a nil *Pool is a bug.

eGGshke:
This method is explicitly called in the Ping method (https://github.com/jackc/pgx/blob/master/pgxpool/pool.go#L797).
It seems to me that it is not quite right to get fatal. Alternatively, I can add a check directly in the ping method.

jackc:
Except in the rare case that there is some meaningful behavior when calling a method on `nil` such as `String()` in some cases, I *would* expect a a fatal error / panic when calling a method on `nil`.

eGGshke:
> Except in the rare case that there is some meaningful behavior when calling a method on `nil` such as `String()` in some cases, I _would_ expect a a fatal error / panic when calling a method on `nil`.

Yes, but it seems to me that this is unacceptable behavior, because when I call Ping, I want to check the connection status, get an error if it is not healthy and process it. But okay, thanks, I get it.

PR #2386, opened 2025-09-17
Always wrap arguments in parentheses in the SQL sanitizer

wull14:
My project discovered two security vulnerabilities during vulnerability scanning when referencing pgx/v3: CVE-2024-27289, CVE-2024-27304. This commit comes from v4: f94eb0e2. But I can't upgrade to v4, so I brought up this MR and hope to agree.


wull14:
By adding this commit to v3, we won't be able to scan for security vulnerabilities. I'm really looking forward to agreeing to this MR.

jackc:
Sorry, `v5` is the only supported version. The last release of `v3` was over 5 years ago.

PR #2389, opened 2025-09-24
[CHORE] Update dependencies

smithcoin:
bumped dependencies via dependabot.


jackc:
The golang.org/x/crypto dependency has been removed and golang.org/x/text has already been updated.

PR #2391, opened 2025-09-26
Close batch pipeline after a query error

bonnefoa:
Currently, if a batch ends with a statement_timeout on a br.Query(), the next br.Close() will only return the query error and leave the underlying pipeline opened. This prevents the connection from being reused and a second br.Close() is required to really close the pipeline.

This patch changes br.Close() to fallthrough and close the pipeline, even with a lastRow error present.


jackc:
Thanks!

PR #2393, opened 2025-09-29
feat(pgxpool): acquire ping timeout

amirsalarsafaei:
this commit adds ping timeout in acquire loop for when the connection ping results in timeout instead of an error

### DIFF
--- a/pgxpool/helper_test.go
+++ b/pgxpool/helper_test.go
@@ -1,0 +1,39 @@
+package pgxpool_test
+
+import (
+	"context"
+	"net"
+	"time"
+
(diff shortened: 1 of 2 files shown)


amirsalarsafaei:
@jackc the PR is final could you please review this.

jackc:
The code looks reasonable but it seems the test is flickering.

amirsalarsafaei:
@jackc  thanks for the review, I think I fixed the flickering, I added a proxy interface simulating real delay (using sleep) to ensure `PingTimeout` never flickers.

jackc:
Thanks!

PR #2409, opened 2025-10-20
zero alloc UUID parse

janisz:
This PR improves UUID parsing.
1. Use bytes instead of string
2. Get raw string data without copy
3. Use array
4. Copy and adapt code from hex package to prevent allocations

```
            │    old.txt    │               new.txt                │
            │    sec/op     │    sec/op     vs base                │
ParseUUID-8   101.00n ± 40%   36.74n ± 12%  -63.62% (p=0.000 n=10)

            │  old.txt   │              new.txt               │
            │    B/op    │   B/op     vs base                 │
ParseUUID-8   48.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

            │  old.txt   │               new.txt               │
            │ allocs/op  │ allocs/op   vs base                 │
ParseUUID-8   2.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

```go
func BenchmarkParseUUID(b *testing.B) {
	val := "bb20d081-9ca7-46e5-a60f-f331775adf5b"
	x, err := parseUUID(val)
	assert.NoError(b, err)
	assert.Equal(b, [16]uint8{0xbb, 0x20, 0xd0, 0x81, 0x9c, 0xa7, 0x46, 0xe5, 0xa6, 0xf, 0xf3, 0x31, 0x77, 0x5a, 0xdf, 0x5b}, x)
	for b.Loop() {
		x, err = parseUUID(val)
	}
	assert.NoError(b, err)
	assert.Equal(b, [16]uint8{0xbb, 0x20, 0xd0, 0x81, 0x9c, 0xa7, 0x46, 0xe5, 0xa6, 0xf, 0xf3, 0x31, 0x77, 0x5a, 0xdf, 0x5b}, x)
}
```

### DIFF
--- a/pgtype/uuid.go
+++ b/pgtype/uuid.go
@@ -5,6 +5,7 @@ import (
 	"database/sql/driver"
 	"encoding/hex"
 	"fmt"
+	"unsafe"
 )
 
 type UUIDScanner interface {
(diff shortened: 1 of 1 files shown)


jackc:
How much difference will this make in real world applications? Parsing text UUIDs is a fallback path. The pgx default is to use the binary format.

Ordinarily, I would still be inclined to accept an optimization even if it is in a less used path, but pgx doesn't use unsafe and unless there is a very strong reason I would prefer to keep it that way.

janisz:
> How much difference will this make in real world applications? Parsing text UUIDs is a fallback path.

Good question. I only saw it in benchmark test and not in production
https://github.com/stackrox/stackrox/blob/a53d91112017f483e8c1397dcb34e863e1e70c90/central/alert/datastore/internal/store/postgres/bench_test.go#L41

image

> but pgx doesn't use unsafe and unless there is a very strong reason I would prefer to keep it that way.

I agree, if it's fallback then probably it's not worth it.

Issue #2223, opened 2025-01-08
Occasional prepared statement already exist errors when using statement timeout


oschwald:
**Describe the bug**

We have a web service where we set a `statement_timeout` on our connections to ensure that queries do not run for an extended period of time. As part of this web service, we have a large, generated query. We occasionally see errors such as the following:

```
ERROR: prepared statement "stmtcache_2b7087dc2ab6296b87fa5baa7f329541c23d17376faa03b7" already exists (SQLSTATE 42P05)
```

My hypothesis is that there is a statement timeout in the `PREPARE` but after Postgres has actually created the prepared statement. This causes `pgx` to end up in a state where the prepared statement isn't added to the statement cache but does exist for the connection in Postgres. I was able to reproduce this with the code below.

**To Reproduce**

The following sample code reliably produces the error with a local Postgres server for me. This probably could be simplified a bit further.

```go
package main

import (
	"context"
	"database/sql"
	"errors"
	"fmt"
	"log"
	"math/rand"
	"os"

	"github.com/jackc/pgx/v5/pgconn"
	_ "github.com/jackc/pgx/v5/stdlib"
)

func main() {
	db, err := sql.Open("pgx", os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}

	ctx := context.Background()

	conn, err := db.Conn(ctx)
	if err != nil {
		log.Fatal(err)
	}

	_, err = conn.ExecContext(ctx, "SET statement_timeout TO 1")
	if err != nil {
		log.Fatal(err)
	}

	p := "s"
	for {
		var v uint

		q := fmt.Sprintf("SELECT 1 AS %s", p)
		err := conn.QueryRowContext(
			ctx,
			q,
		).Scan(&v)
		if err != nil && !errors.Is(err, context.Canceled) {
			var pgErr *pgconn.PgError
			if errors.As(err, &pgErr) && pgErr.Code == "57014" {
				continue
			}
			log.Fatal(err)
		}
		p = randString(100000)
	}
}

var letterRunes = []rune("abAB")

func randString(n int) string {
	b := make([]rune, n)
	for i := range b {
		b[i] = letterRunes[rand.Intn(len(letterRunes))]
	}
	return string(b)
}
```

**Expected behavior**

No prepared statement already exists error.

**Actual behavior**

```
[...]

PR #2411, opened 2025-10-21
Prepare overwrites existing prepared statement

jackc:
Prepare now sends a Close message as part of preparing a statement. This allows changes Prepare behavior in that it is no longer an error to prepare a statement with a name that already exists. In addition, it resolves issues where the connection gets confused and doesn't realize a statement is already prepared. This could happen if a statement_timout triggered between the Parse and Describe messages.

https://github.com/jackc/pgx/issues/2223

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1139,6 +1139,7 @@ func (c *Conn) sendBatchExtendedWithDescription(ctx context.Context, b *Batch, d
 	if len(distinctNewQueries) > 0 {
 		err := func() (err error) {
 			for _, sd := range distinctNewQueries {
+				pipeline.SendDeallocate(sd.Name)
 				pipeline.SendPrepare(sd.Name, sd.SQL, nil)
 			}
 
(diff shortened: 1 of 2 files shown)


jackc:
If this approach works, the batch system will also need to be updated to send the Close message when preparing statements.

oschwald:
This seems promising. It seems to resolve the issue for the easy-to-test reproducers that I have.

jackc:
I just added this logic into batches. It works, but I want to think a bit more before committing to this approach.

It feels slightly inelegant to always send a `Close` even when there shouldn't be an existing prepared statement.

This latest commit also introduces a slight difference in behavior between `Pipeline.SendPrepare` and `PgConn.SendPrepare`. The pipeline version doesn't send the `Close` because then different results would be returned to existing direct users of pipeline mode. I might be able to figure out a way to preserve existing behavior, but again, it starts to feel inelegant.

PR #2412, opened 2025-10-22
Add interface for Pool to make tests with mocks

gonnafaraway:
- add independent interface, that can be used by dev for mocking Pool with mockgen

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -74,6 +74,25 @@ func (cr *connResource) getPoolRows(c *Conn, r pgx.Rows) *poolRows {
 	return pr
 }
 
+// PoolModel describes main methods for reuse.
+type PoolModel interface {
+	Close()
+	Acquire(ctx context.Context) (c *Conn, err error)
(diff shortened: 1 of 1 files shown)


gonnafaraway:
@jackc hi, plz help :)

jackc:
See https://go.dev/wiki/CodeReviewComments#interfaces

> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

> Do not define interfaces on the implementor side of an API “for mocking”; instead, design the API so that it can be tested using the public API of the real implementation.

gonnafaraway:
i'm sure that it's not good practice to use interface in this situation as value
you say about "implementing" package, but pgx is not
it's abstract library, that provides driver to make requests to db
and also, this means, that you have interface and implementation of it, not only implementation.

i know many devs, that break mind about pgx about it doesn't contain main interface, all of them use the same mocks

best regards!

PR #2418, opened 2025-11-01
Prepare handles describe failure

### LINKED ISSUE #2223: Occasional prepared statement already exist errors when using statement timeout

jackc:
This handles a `Prepare()` that fails after `ParseComplete` but before receiving the complete statement description. Previously, this could leave a prepared statement on the server without pgx knowing about the statement. This could cause repeated failures as pgx continued to try to prepare the statement and failed.

Solved by detecting that error specifically and deallocating the statement during the next Prepare.

The same error in batches was already handled via invalidating the newly prepared statements in the statement cache.

Also, introduces faultyconn package which allows introducing difficult to reproduce errors into the PostgreSQL connection. This allowed creating a delay between the `Parse` and `Describe` messages which reliably reproduced this error in tests. 

faultyconn required adding `AfterNetConnect` hook to `pgconn.Config`. While this is mostly useful for testing, it could also be useful for logging or production diagnostics.

Fixes #2223
Supersedes #2411.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -1091,6 +1091,8 @@ func TestSendBatchHandlesTimeoutBetweenParseAndDescribe(t *testing.T) {
 	}
 
 	pgxtest.RunWithQueryExecModes(ctx, t, faultyConnTestRunner, extendedQueryModes, func(ctx context.Context, t testing.TB, conn *pgx.Conn) {
+		pgxtest.SkipCockroachDB(t, conn, "Induced error does not occur on CockroachDB")
+
 		_, err := conn.Exec(ctx, "set statement_timeout = '100ms'")
 		require.NoError(t, err)
 
--- a/conn_test.go
+++ b/conn_test.go
@@ -489,6 +489,8 @@ func TestPrepareHandlesTimeoutBetweenParseAndDescribe(t *testing.T) {
 	defer closeConn(t, conn)
 	require.NotNil(t, faultyConn)
 
+	pgxtest.SkipCockroachDB(t, conn, "Induced error does not occur on CockroachDB")
+
 	_, err = conn.Exec(ctx, "set statement_timeout = '100ms'")
 	require.NoError(t, err)
 


jackc:
@oschwald This approach is a bit more elegant than my previous PR. The new `faultyconn` test facility this prompted me to write should also aid in reproducing other rare cases in tests.

oschwald:
Thank you! We will try it out.

PR #2432, opened 2025-11-15
perf(iobufpool): optimize pool index calculations with O(1) bit manipulation

analytically:
Replace O(n) loops in getPoolIdx and putPoolIdx with O(1) direct calculation using bits.Len and bits.TrailingZeros. Pool sizes are powers of 2, so we can derive indices directly from buffer size/capacity exponents.

getPoolIdx:  Use bits.Len to calculate ceil(log2(size)) in constant time
putPoolIdx:  Use bits.TrailingZeros to calculate log2(cap) in constant time

Benchmarks (Apple M1 Pro):
  - getPoolIdx:      16.97ns → 3.52ns   (4.8x faster)
  - putPoolIdx:      24.75ns → 4.16ns   (5.9x faster)
  - Get/Put cycle:   89.15ns → 76.57ns  (14% faster)
  - Get(32K):        15.38ns → 10.52ns  (32% faster)

### DIFF
--- a/internal/iobufpool/iobufpool.go
+++ b/internal/iobufpool/iobufpool.go
@@ -4,7 +4,10 @@
 // an allocation is purposely not documented. https://github.com/golang/go/issues/16323
 package iobufpool
 
-import "sync"
+import (
+	"math/bits"
+	"sync"
+)
 
 const minPoolExpOf2 = 8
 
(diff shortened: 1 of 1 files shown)


analytically:
Failing CI doesn't seem related?

PR #2433, opened 2025-11-16
perf(stmtcache): replace O(n) invalidation scan with O(1) map lookup

analytically:
Replace linear slice scan in LRUCache.Put() with hash map lookup for invalidated statement tracking. Maintains both invalidStmts slice and invalidSet map to optimize both Put() and GetInvalidated() operations.

Performance impact (Apple M1 Pro):
- Baseline: 1.41x faster (178ns → 126ns)
- 100 invalidations: 1.56x faster (203ns → 131ns)
- 500 invalidations: 2.21x faster (287ns → 130ns)
- Worst case (re-adding invalidated): 3.85x faster (570ns → 148ns)

Original implementation degrades from 178ns to 570ns as invalidations accumulate. Optimized version maintains constant ~130ns performance regardless of invalidation count.

Memory overhead: ~8 bytes per invalidated statement. No API changes.

### DIFF
--- a/internal/stmtcache/lru_cache.go
+++ b/internal/stmtcache/lru_cache.go
@@ -12,14 +12,16 @@ type LRUCache struct {
 	m            map[string]*list.Element
 	l            *list.List
 	invalidStmts []*pgconn.StatementDescription
+	invalidSet   map[string]struct{}
 }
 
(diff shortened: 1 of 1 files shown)


jackc:
The code seems sound, though it does seem to be a code path that should be very rarely traversed.

PR #2437, opened 2025-11-19
Add pgvector Vector type support

MaanasSathaye:
Implements Vector type for pgvector extension compatibility. Includes binary and text format codecs, Scanner/Valuer interfaces, JSON marshaling, and comprehensive tests.

prior art for this exists in https://github.com/pgvector/pgvector-go/tree/master/pgx. I don't see a need just yet for halfvecs or sparsevecs though.


jackc:
Is there something wrong with the support provided by the official pgvector project you linked to? It appears to be maintained. I'd prefer to use that unless there is a good reason not to.

SirHomeboy:
> Is there something wrong with the support provided by the official pgvector project you linked to? It appears to be maintained. I'd prefer to use that unless there is a good reason not to.

@jackc sorry, missed this in my inbox. It's a bit of a russian doll problem. A few of my projects use [genieql](https://github.com/james-lawrence/genieql), which relies on the pgx driver for postgres. We've added a workaround solution to custom define a vector type in a supplemental driver, but are looking for something more native. The linked repo from @ankane contains the necessary vector types, but I can't read pgx types from two sources in genieql currently. Other solution would be for me to fork pgx + pgvector-go and maintain my own, but that becomes a bit of dependency mess. 

I'm happy to close this MR and let Andrew open his own too.

jackc:
>> Is there something wrong with the support provided by the official pgvector project you linked to? It appears to be maintained. I'd prefer to use that unless there is a good reason not to.

> @jackc sorry, missed this in my inbox. It's a bit of a russian doll problem. A few of my projects use [genieql](https://github.com/james-lawrence/genieql), which relies on the pgx driver for postgres. We've added a workaround solution to custom define a vector type in a supplemental driver, but are looking for something more native. The linked repo from @ankane contains the necessary vector types, but I can't read pgx types from two sources in genieql currently. Other solution would be for me to fork pgx + pgvector-go and maintain my own, but that becomes a bit of dependency mess.

> I'm happy to close this MR and let Andrew open his own too.

Okay. I'm going to close this. I think that support for extensions should be outside of pgx in almost all cases. Especially in this case where the extension is providing Go support officially.

Also, extensions may have different release cycles and backwards compatibility goals than PostgreSQL and pgx. Even if we wanted wanted more or less official support in the pgx project, it would still end up being a separate package like what has been done with https://github.com/jackc/pgx-shopspring-decimal.

Issue #2447, opened 2025-12-09
ColumnTypeLength method in stdlib/sql.go can not return the type length for 'varbit' type


DengChan:
Hi, I'm using pgx as a database/sql driver  to build a pg client. 
I try to use ColumnTypeLength function to get the type length for varbit type field, but it doesn't work.
I think it would be helpful to me if the code could be rewritten as below:

```
func (r *Rows) ColumnTypeLength(index int) (int64, bool) {
	fd := r.rows.FieldDescriptions()[index]

	switch fd.DataTypeOID {
	case pgtype.TextOID, pgtype.ByteaOID:
		return math.MaxInt64, true
	case pgtype.VarcharOID, pgtype.BPCharArrayOID:
		return int64(fd.TypeModifier - varHeaderSize), true
	case pgtype.VarbitOID:
		return int64(fd.TypeModifier), true
	default:
		return 0, false
	}
}
```

If you don't mind, I could submit a PR~

DengChan:
Here is my PR, #2448 
If you feel this PR is not very helpful, you can close it directly.
Thank you~

jackc:
#2448 has been merged.

PR #2448, opened 2025-12-09
ColumnTypeLength method return the type length for varbit type.

DengChan:
#2447


jackc:
Seems reasonable. But can you add some tests showing that it works?

DengChan:
> Seems reasonable. But can you add some tests showing that it works?

Sure! 
I add test case `TestRowsColumnTypeLength` in `stdlib/sql_test.go`, and the case passed as expected.

jackc:
Thanks!

PR #2461, opened 2026-01-01
pgio: optimize append/set functions with direct byte shifts

analytically:
Replace zero-fill + overwrite pattern with direct byte appends. Use array pointer cast in SetInt32 to eliminate bounds checks.

Benchmarks show 12-19% improvement in message building:
  BindMessage5Params:  164 ns → 132 ns  (19% faster)
  BindMessage20Params: 922 ns → 780 ns  (15% faster)
  Simulate10kQueries: 1.37ms → 1.21ms  (12% faster)

Assembly comparison for SetInt32:
  Before: 134 bytes, 4 bounds checks, 4 panicIndex calls
  After:   63 bytes, 1 bounds check,  1 panicSliceConvert call

No functional changes. All existing tests pass.

### DIFF
--- a/internal/pgio/write.go
+++ b/internal/pgio/write.go
@@ -1,26 +1,18 @@
 package pgio
 
-import "encoding/binary"
-
 func AppendUint16(buf []byte, n uint16) []byte {
-	wp := len(buf)
-	buf = append(buf, 0, 0)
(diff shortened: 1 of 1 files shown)


jackc:
Thanks

PR #2463, opened 2026-01-01
perf(pgtype): pre-allocate error sentinels for NULL scans

analytically:
~320x faster NULL scanning with zero allocations.

Benchmark results:
- NULL scan: 97 ns → 0.3 ns
- Mixed (2 NULL / 5 cols): 199 ns → 5.5 ns
- Memory: 48 B/op → 0 B/op per NULL

Adds 15 sentinel error variables (~800 bytes total, allocated once at startup) for common NULL-into-non-pointer scan errors. Non-NULL error paths unchanged.

Also fixes uintWrapper.ScanInt64 error message (*uint64 → *uint).

### DIFF
--- a/pgtype/builtin_wrappers.go
+++ b/pgtype/builtin_wrappers.go
@@ -11,13 +11,32 @@ import (
 	"time"
 )
 
+// Pre-allocated error sentinels for NULL scans (hot path optimization)
+var (
+	errNullInt8    = errors.New("cannot scan NULL into *int8")
+	errNullInt16   = errors.New("cannot scan NULL into *int16")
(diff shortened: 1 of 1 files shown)


jackc:
Is there a case where this has a real world impact? It's optimizing an error path that should only be triggered by application bugs.

analytically:
Impacts:
- Nullable columns, scanned into non-pointers, many NULLs
- LEFT JOINs scanned into non-pointers

It's a low-risk change with occasional high payoff.

jackc:
If you are scanning something that can be `null` into a Go value that can't accept it, that is an application bug. That is not a path to optimize that is behavior that needs to be fixed in the application.

PR #2466, opened 2026-01-03
perf(trace): optimize traceSingleQuotedString

analytically:
Replace fmt.Fprintf with direct hex encoding using lookup table. Add Grow() pre-allocation to eliminate builder reallocations.

Benchmarks show 1.7-8x speedup depending on input:
- all non-printable: 743ns -> 92ns (8x)
- mixed content:     422ns -> 108ns (4x)
- printable only:    242ns -> 144ns (1.7x)

Allocations reduced from 2-6 to 1 in all cases.

### DIFF
--- a/pgproto3/trace.go
+++ b/pgproto3/trace.go
@@ -397,15 +397,20 @@ func traceDoubleQuotedString(buf []byte) string {
 	return `"` + string(buf) + `"`
 }
 
+const hexChars = "0123456789abcdef"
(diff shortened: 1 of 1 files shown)


jackc:
Take a look at #2457. I'm still reluctant to merge optimizations of paths that shouldn't be triggered very often, but if it is to be done then I think that #2457 is a superset of this.

PR #2476, opened 2026-01-12
Fix golangci-lint configuration file

alexandear:
This PR fixes wrong elements in the `.golangci.yml`:

```console
$ golangci-lint config verify
jsonschema: "version" does not validate with "/properties/version/type": got number, want string
jsonschema: "formatters" does not validate with "/properties/formatters/additionalProperties": additional properties 'default' not allowed
The command is terminated due to an error: the configuration contains invalid elements
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 # See for configurations: https://golangci-lint.run/usage/configuration/
-version: 2
+version: "2"
 
 linters:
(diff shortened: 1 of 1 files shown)


dolmen:
Relevant golangci-lint documentation:  https://golangci-lint.run/docs/configuration/file/#formatters-configuration

PR #2480, opened 2026-01-14
docs(RegisterConnConfig): explain alternative usage

AldoFusterTurpin:
**Context**
The fact that RegisterConnConfig "holds" an internal map that can grow without bound is something that could be stated explicitly.

**PR changes**
I believe it would be a good idea to explain an alternative for clients with many short-lived connections to many DBs, to avoid memory leaks. Being a bit more explicit and vocal in the comments about how that works will help clients to choose if they really want to use RegisterConnConfig or the alternative shared.

I placed the comment both in
- RegisterConnConfig() (I believe this would be the perfect place as when you use a library you just check the docs from your editor and sometimes don't browse pkg.go.dev), 
- the top of the file in case people is browsing the documentation in https://pkg.go.dev/ so they can read through.

WDYT ? 

Thank you!

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -32,6 +32,12 @@
 //	connStr := stdlib.RegisterConnConfig(connConfig)
 //	db, _ := sql.Open("pgx", connStr)
 //
+// Consider using stdlib.OpenDB + sqlx.NewDb instead of RegisterConnConfig
+// if your client code perform many short-lived connections to multiple DBs (to avoid memory leaks).
+//
+// db := stdlib.OpenDB(*pgxConfig)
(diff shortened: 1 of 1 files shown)


AldoFusterTurpin:
@jackc (Sorry for the ping) I believe this is pretty straightforward and can help new people using the module 🙂

jackc:
So it would make sense to document and emphasize the usage of `stdlib.OpenDB`. Whenever possible that should be preferred over `RegisterConnConfig`.

However, the warnings in your PR aren't exactly correct. It does not maintain a map of connections, it maintains a map of configs. You should be able to make any number of connections from a single registered conn config. Also, you can unregister them as well.

PR #2482, opened 2026-01-18
Don't log pgx.QueryResultFormatsByOID

nkonev:
Hello!

In case using `stdlib.OpenDBFromPool()` there is senseless `map pgx.QueryResultFormatsByOID like [map[16:1 17:1 20:1 21:1 23:1 26:1 28:1 29:1 700:1 701:1 1082:1 1114:1 1184:1] 3]` as the first argument.

Reproducer is here https://github.com/nkonev/pgx-log-repro

Before (branch `master` of the reproducer)
Screenshot from 2026-01-19 01-22-15

After (branch `fix` of the reproducer)
Screenshot from 2026-01-19 01-22-00

### DIFF
--- a/tracelog/tracelog.go
+++ b/tracelog/tracelog.go
@@ -94,7 +94,20 @@ func LogLevelFromString(s string) (LogLevel, error) {
 func logQueryArgs(args []any) []any {
 	logArgs := make([]any, 0, len(args))
 
-	for _, a := range args {
+	for i, a := range args {
+		if i == 0 {
+			switch a.(type) {
+			case pgx.QueryResultFormats:
+				continue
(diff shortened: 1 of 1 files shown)


jackc:
I'm not sure about this. In particular, eliding `pgx.QueryRewriter` would remove `NamedArgs` from being logged. That seems undesirable. In an ideal world, there should be a place for a `QueryTracer` to get the SQL and args after all preprocessing. Perhaps another hook, perhaps final SQL and args get passed to `TraceQueryEnd`.

nkonev:
@jackc 
Generally you are right, but I don't feel so confident in the pgx internals =(

Let's leave this PR open, maybe somebody find a time to make it correctly.

Issue #872, opened 2020-11-15
Batch errors associated with wrong statement


phemmer:
When using `SendBatch()`, if one of the statements in the batch returns an error during prepared statement generation, the error is returned on the first result in the `BatchResult`. This is incorrect as the error may have been a result of latter statements.

For example:
```
batch := &pgx.Batch{}
batch.Queue("select 1")
batch.Queue("bad")
result := conn.SendBatch(context.Background(), batch)

_, err := result.Query()
fmt.Print(err)
```
```
ERROR: syntax error at or near "bad" (SQLSTATE 42601)
```

This makes it very confusing as it implies the error came from `select 1`.

I'm not sure the right way this should be solved. However I would argue that if a prepared statement doesn't already exist in the cache, it should not be making a round trip to create it, as this defeats the entire point of using batch statements.

Additionally, in my use case it's actually the usage of prepared statements that is causing the problem. The second statement is throwing an error because a table being referenced doesn't exist. However the first statement in the batch is the one creating it.

phemmer:
Another related issue, the documentation for [ConnConfig](https://pkg.go.dev/github.com/jackc/pgx/v4#ConnConfig) states:
```
	// BuildStatementCache creates the stmtcache.Cache implementation for connections created with this config. Set
	// to nil to disable automatic prepared statements.
```

For batch statements this is incorrect. `Conn.SendBatch` contains [the following code](https://github.com/jackc/pgx/blob/v4.9.2/conn.go#L706-L710):
```
		if c.stmtcache != nil && c.stmtcache.Cap() >= len(distinctUnpreparedQueries) {
			stmtCache = c.stmtcache
		} else {
			stmtCache = stmtcache.New(c.pgConn, stmtcache.ModeDescribe, len(distinctUnpreparedQueries))
		}
```
So basically it ignores the setting and creates a prepared statement cache anyway.

jackc:
pgx uses the PostgreSQL extended protocol by default. The extended protocol can only execute prepared statements -- in a sense -- there is a special nameless and ephemeral prepared statement that is used. See https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY for the details.

When using the extended protocol pgx needs to know the underlying data types of the parameters and results. Without a statement cache it does two round trips to the database, one to prepare and describe the SQL and another to execute it. This is necessary to safely use the binary format -- which can make a significant performance improvement.

[...]

PR #2490, opened 2026-01-29
Add more context to SendBatch error 

makalaaneesh:
Hi @jackc, this is a follow-up to https://github.com/jackc/pgx/issues/872#issuecomment-2422463155 

If there is an error while preparing statements in a batch (syntax error for example), 
1. the error is thrown as part of the first result regardless of which statement was responsible. 
2. the error  does not include the culprit SQL statement, which makes it hard to debug. 

As fixing 1 is tricky as pointed out in the #872, 
This PR just adds a bit of context to the error (similar to https://github.com/jackc/pgx/pull/1441 )

Code to test:
```
	ctx := context.Background()
	conn, err := pgx.Connect(ctx, "postgres://localhost:5432/postgres")
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(ctx)

	// Setup: create table outside batch
	conn.Exec(ctx, "DROP TABLE IF EXISTS batch_test")
	conn.Exec(ctx, "CREATE TABLE batch_test (id SERIAL PRIMARY KEY, name TEXT, value INT)")

	// Batch with 3 statements
	batch := &pgx.Batch{}
	batch.Queue("INSERT INTO batch_test (name, value) VALUES ($1, $2)", "item1", 100)
	batch.Queue("INSERT INTO batch_test (name, value) VALS ($1, $2)", "item2", 200) // SYNTAX ERROR
	batch.Queue("INSERT INTO batch_test (name, value) VALUES ($1, $2)", "item3", 300)

	br := conn.SendBatch(ctx, batch)
	for i := 0; i < 3; i++ {
		if _, err := br.Exec(); err != nil {
			panic(fmt.Sprintf("statement %d failed: %v", i+1, err))
		}
	}
	br.Close()

	fmt.Println("Batch executed successfully")
```

Before: 
`panic: statement 1 failed: ERROR: syntax error at or near "VALS" (SQLSTATE 42601)`

After: 
`panic: statement 1 failed: failed to prepare statement: INSERT INTO batch_test (name, value) VALS ($1, $2): ERROR: syntax error at or near "VALS" (SQLSTATE 42601)`

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -521,7 +521,8 @@ func newErrPreprocessingBatch(step, sql string, err error) ErrPreprocessingBatch
 
 func (e ErrPreprocessingBatch) Error() string {
 	// intentionally not including the SQL query in the error message
-	// to avoid leaking potentially sensitive information into logs
+	// to avoid leaking potentially sensitive information into logs.
+	// If the user wants the SQL, they can call SQL().
 	return fmt.Sprintf("error preprocessing batch (%s): %v", e.step, e.err)
 }
 


jackc:
I'm wary about including the entire SQL string in an error. We've had complaints before about error messages potentially leaking sensitive information into logs. I'd prefer to avoid adding any new leak vectors. Also, SQL text can be very large and I'm not sure we want to put the whole thing in error text.

I am still interested in improving this situation though. The "failed to prepare statement" part of the message is interesting. That, or something like it that made clear that the error occurred while preparing the batch is a good idea.

makalaaneesh:
That's fair. 

How about returning a specific error struct (something like `ErrPreprocessingBatch` )? 
- It would contain the exact SQL of the batch, more context on which step failed (prepare, in this case), etc
- Its string representation would NOT contain the SQL. 

That way, 
- by default, we improve the error reporting by making it clear that the error occurred while pre-processing the batch. No SQL in the string representation, so no leaking sensitive info by default. 
- if the user wants to get the specific problematic SQL, they can inspect the error struct.  

What do you think?

jackc:
That sounds like a reasonable approach.

makalaaneesh:
Hi @jackc , I've made the discussed changes.

PR #2491, opened 2026-01-29
feat: export PoolBatchResults, ErrBatchResults, ErrRows and ErrRow

grachevko:
Exports previously internal error wrapper types (ErrBatchResults, ErrRows, ErrRow) and adds PoolBatchResults constructor.

We implement custom connection pool logic combined with sqlc and currently we have to duplicate internal pgx error wrapper implementations.

Is this type of change acceptable?

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -505,3 +505,27 @@ func invalidateCachesOnBatchResultsError(conn *Conn, b *Batch, err error) {
 		}
 	}
 }
+
+type errBatchResults struct {
+	err error
+}
+
+func ErrBatchResults(err error) BatchResults {
+	return errBatchResults{err: err}
+}
+
+func (br errBatchResults) Exec() (pgconn.CommandTag, error) {
+	return pgconn.CommandTag{}, br.err
+}
+
+func (br errBatchResults) Query() (Rows, error) {
+	return ErrRows(br.err), br.err
+}
+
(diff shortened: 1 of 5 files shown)


jackc:
I'd prefer to keep these types as internal implementation details if possible. We don't currently expose the concrete types for non-error responses either.

Does this unblock functionality you need or is the goal to reduce some duplicated code?

grachevko:
> I'd prefer to keep these types as internal implementation details if possible. We don't currently expose the concrete types for non-error responses either.

This PR not expose any types, only constructors, which returns already exposed interfaces. Types keep internal. 

> Does this unblock functionality you need or is the goal to reduce some duplicated code?

Only reduce duplicates

jackc:
>> I'd prefer to keep these types as internal implementation details if possible. We don't currently expose the concrete types for non-error responses either.

> This PR not expose any types, only constructors, which returns already exposed interfaces. Types keep internal.

Ah. I see now. 

>>    Does this unblock functionality you need or is the goal to reduce some duplicated code?

> Only reduce duplicates

In general, I prefer to avoid increasing the API surface area unless the functionality cannot be built outside of pgx or the need is extremely common. In this case, since the functionality can be built outside and it is not an extremely common request, I'd rather not merge this.

PR #2496, opened 2026-02-01
Fix empty user in connection string defaulting to OS user

### LINKED ISSUE #2495: Empty user in connection string does not work
**Describe the bug**
When the connection string does not have a user specified, pgx will fail to connect due to 

```
server error: FATAL: no PostgreSQL user name specified in startup packet (SQLSTATE 28000)
```

**To Reproduce**

```go
package main

import (
        "context"
        "fmt"
        "log"

        "github.com/jackc/pgx/v5"
)

func main() {
        connStr := "host=/var/run/postgresql/ port=5432 dbname=maas_test user="

        conn, err := pgx.Connect(context.Background(), connStr)
        if err != nil {
                log.Fatal("Connection failed:", err)
        }
}
```

**Expected behavior**

The same connection string work with many other drivers (python-asyncpg for example) and with psql itself

```
psql "host=/var/run/postgresql/ port=5432 dbname=maasdb user="
```

this is because they automatically pick the current user executing the application.

```
ubuntu@maas-noble:/tmp/test$ psql "host=/var/run/postgresql/ port=5432 dbname=maasdb user="
psql (16.11 (Ubuntu 16.11-0ubuntu0.24.04.1))
Type "help" for help.

maasdb=> \conninfo
You are connected to database "maasdb" as user "ubuntu" via socket in "/var/run/postgresql/" at port "5432".
```

**Actual behavior**
Crashes with 
```
2026/01/30 22:20:56 Connection failed:failed to connect to `user= database=maas_test`: /var/run/postgresql/.s.PGSQL.5432 (/var/run/postgresql/): server error: FATAL: no PostgreSQL user name specified in startup packet (SQLSTATE 28000)
```

**Version**
 - Go: go version go1.24.4 linux/amd64
 - PostgreSQL: `$  PostgreSQL 16.11 (Ubuntu 16.11-0ubuntu0.24.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, 64-bit
 - pgx: v5.8.0

shv-ng:
Fixes #2495

Defaults the PostgreSQL user to the current OS user when the connection string explicitly specifies an empty `user=` parameter, matching libpq, psql, and asyncpg behavior.

### What
Default the PostgreSQL user to the current OS user when the connection string
explicitly specifies an empty `user=` parameter.

### Why
libpq and other drivers (e.g. psql, asyncpg) treat `user=` as a signal to fall
back to the current OS user. pgx currently leaves the user empty, which causes
the server to reject the startup packet with SQLSTATE 28000.

### How
- Detect when `user` is explicitly present but empty
- Default `Config.User` using `os/user.Current()`
- Ignore lookup errors to avoid introducing new failure modes

### Testing
- Added test covering explicit empty `user=` behavior

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -13,7 +13,6 @@ import (
 	"net"
 	"net/url"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -305,13 +304,6 @@ func ParseConfigWithOptions(connString string, options ParseConfigOptions) (*Con
 		},
 	}
 
-	if _, present := settings["user"]; present && config.User == "" {
-		currentUser, err := user.Current()
-		if err == nil {
-			config.User = currentUser.Username
-		}
-	}
-
 	if connectTimeoutSetting, present := settings["connect_timeout"]; present {
 		connectTimeout, err := parseConnectTimeoutSetting(connectTimeoutSetting)
 		if err != nil {
@@ -499,7 +491,9 @@ func parseURLSettings(connString string) (map[string]string, error) {
 	}
 
 	if parsedURL.User != nil {
-		settings["user"] = parsedURL.User.Username()
+		if u := parsedURL.User.Username(); u != "" {
+			settings["user"] = u
+		}
 		if password, present := parsedURL.User.Password(); present {
 			settings["password"] = password
 		}
@@ -626,6 +620,9 @@ func parseKeywordValueSettings(s string) (map[string]string, error) {
 			return nil, errors.New("invalid keyword/value")
 		}
 
+		if key == "user" && val == "" {
+			continue
+		}
 		settings[key] = val
 	}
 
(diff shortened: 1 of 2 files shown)


jackc:
We already get the OS user in `defaultSettings()`. The problem is it is being overridden by the explicitly set user. Instead of duplicating the logic to get the OS user, perhaps it would be better not to overwrite user with an empty string. This could be done in `mergeSettings()` or in `parseURLSettings()` and `parseKeywordValueSettings()`.

shv-ng:
Thanks, that makes sense.

I’ve updated the patch to avoid duplicating the OS user lookup. The fix now
ignores explicitly empty `user` values during parsing (both URL and
keyword/value), allowing the default user from `defaultSettings()` to remain.
I also added tests covering keyword and URL forms, with and without passwords.

Issue #2119, opened 2024-09-04
Query execution stuck until idle-in-transaction timeout


sergeydobrodey:
Query execution stuck until idle-in-transaction timeout is reached. Library stack trace:

```
goroutine 11794184 [IO wait, 4 minutes]:
internal/poll.runtime_pollWait(0x7f98fab5fda8, 0x72)
	runtime/netpoll.go:343 +0x85
internal/poll.(*pollDesc).wait(0xc00bca4d00?, 0xc004116000?, 0x0)
	internal/poll/fd_poll_runtime.go:84 +0x27
internal/poll.(*pollDesc).waitRead(...)
	internal/poll/fd_poll_runtime.go:89
internal/poll.(*FD).Read(0xc00bca4d00, {0xc004116000, 0x2000, 0x2000})
	internal/poll/fd_unix.go:164 +0x27a
net.(*netFD).Read(0xc00bca4d00, {0xc004116000?, 0xce20c48?, 0xc0043ef670?})
	net/fd_posix.go:55 +0x25
net.(*conn).Read(0xc002f02e40, {0xc004116000?, 0x300e060?, 0xc0043ef6b0?})
	net/net.go:179 +0x45
github.com/jackc/pgx/v5/pgconn/internal/bgreader.(*BGReader).Read(0xc007450880, {0xc004116000, 0x2000, 0x2000})
	github.com/jackc/pgx/v5@v5.5.5/pgconn/internal/bgreader/bgreader.go:100 +0xd7
io.ReadAtLeast({0x2120140, 0xc007450880}, {0xc004116000, 0x2000, 0x2000}, 0x5)
	io/io.go:335 +0x90
github.com/jackc/pgx/v5/pgproto3.(*chunkReader).Next(0xc0067c6ff0, 0x5)
	github.com/jackc/pgx/v5@v5.5.5/pgproto3/chunkreader.go:80 +0x291
github.com/jackc/pgx/v5/pgproto3.(*Frontend).Receive(0xc00bda2000)
	github.com/jackc/pgx/v5@v5.5.5/pgproto3/frontend.go:307 +0x3f
github.com/jackc/pgx/v5/pgconn.(*PgConn).peekMessage(0xc00bca8480)
	github.com/jackc/pgx/v5@v5.5.5/pgconn/pgconn.go:524 +0x14a
github.com/jackc/pgx/v5/pgconn.(*PgConn).receiveMessage(0xc00bca8480)
[...]

PR #2502, opened 2026-02-10
Fix BGReader deadlock when stopping background reads

sergeydobrodey:
BGReader.Read could block forever if Stop() was called while the background goroutine was stuck in a read.

Read now falls back to direct reading when status is StatusStopping, not just StatusStopped.
Adds a unit test that reproduces the deadlock and verifies the fix.

Tests
go test ./pgconn/internal/bgreader -run TestBGReaderReadDoesNotWaitWhenStopping

Possible it's a fix for https://github.com/jackc/pgx/issues/2119

### DIFF
--- a/pgconn/internal/bgreader/bgreader.go
+++ b/pgconn/internal/bgreader/bgreader.go
@@ -95,8 +95,8 @@ func (r *BGReader) Read(p []byte) (int, error) {
 		return r.readFromReadResults(p)
 	}
 
-	// There are no unread background read results and the background reader is stopped.
-	if r.status == StatusStopped {
+	// There are no unread background read results and the background reader is stopped or stopping.
+	if r.status == StatusStopped || r.status == StatusStopping {
 		return r.r.Read(p)
 	}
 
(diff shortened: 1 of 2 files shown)


jackc:
I think a race condition would be possible with this change. `BGReader` is defined as concurrency safe, but there is no guarantee the underlying `io.Reader` is. With this change the underlying `io.Reader` can be used concurrently.

1. `BGReader.Start()` is called. `bgRead()` is started in a goroutine and blocks on `Read()` from the underlying `io.Reader`.
2. `BGReader.Read()` is called and execution waits at `r.cond.Wait()`
3. `BGReader.Stop()` is called. This sets status to `StatusStopping`, but it does not stop the goroutine running `bgRead()`.
4. `BGReader.Read()` is called again. Because status is `StatusStopping` it also calls `Read()` from the underlying `io.Reader`.

In practice, this shouldn't happen, both because we don't call `BGReader.Read()` concurrently, and we only use it with `net.Conn` which is concurrency safe. However, I don't want to introduce a latent bug either.

sergeydobrodey:
Agree. I'll close this PR

PR #2504, opened 2026-02-14
Make LoadTypes query easier to support on "postgres-like" servers

JelteF:
I'm working with one of the many Postgres "compatible" servers, and making this small change to the query from the `LoadTypes` function makes it "just work". It was not needed to use the simple protocol for this query, it was only necessary for the results to use text encoding.


jackc:
See the CI test failures. It seems that some of the returned values must be in text format. That was probably the reason for using the simple protocol. You might be able to use `QueryResultFormats` to fix the issue.

JelteF:
Oops sorry, fixed now.

PR #2506, opened 2026-02-15
Skip standard_conforming_strings=off on PG19

JelteF:
For the same security reason why `standard_conforming_strings=off` is disallowed by pgx in combination with the simple query protocol, in PG19 `standard_conforming_strings=off` won't be allowed anymore at all.

See this for details: https://github.com/postgres/postgres/commit/45762084545ec14dbbe66ace1d69d7e89f8978ac


JelteF:
To clarify, I ran into this while running pgx tests against a local postgres server that was build from the postgres master branch.

PR #2509, opened 2026-02-21
fix returning wrong error in configTLS when DecryptPEMBlock fails

m-ocean-it:
Just discovered this seeming mistake with a linter. Probably, not critical at all.


jackc:
Thanks

PR #2511, opened 2026-03-04
Add Batch.QueueBulkInsert for multi-row INSERT batching

w00j00ng:
### Summary

Add a new `Batch.QueueBulkInsert` method that automatically rewrites a single-row INSERT template into a multi-row `INSERT ... VALUES (...), (...), (...)` statement and queues it as a batch query. This provides a convenient, high-performance way to bulk insert many rows using pgx's existing batch mechanism.

### Motivation

pgx's `Batch` already minimizes network round trips — `ExecBatch` sends all queued queries in a single round trip, and pipeline-based modes (CacheStatement, CacheDescribe, DescribeExec) pipeline everything through one or two flushes. **Round trips are not the problem.**

The problem is **per-statement overhead on the PostgreSQL server**. When N rows are queued as N individual `INSERT` statements within a batch, PostgreSQL must still parse, plan, and execute each statement separately, even though they all arrive in a single network write. Each statement also carries its own wire-protocol message sequence (Parse / Bind / Describe / Execute in extended protocol, or a separate semicolon-delimited SQL string in simple protocol). For large N this overhead dominates.

A multi-row `INSERT ... VALUES (...), (...), (...)` collapses N rows into a single statement that PostgreSQL parses once, plans once, and executes as one operation. This is significantly faster in practice.

pgx already provides `CopyFrom` for maximum throughput, but `COPY` cannot use `ON CONFLICT`, `RETURNING`, default expressions, or triggers on individual rows. Multi-row INSERT fills the gap: it's faster than N×1-row batched inserts while retaining full SQL flexibility.

Building the expanded SQL and renumbered placeholders by hand is tedious and error-prone. `QueueBulkInsert` automates this.

### Usage

```go
batch := &pgx.Batch{}
qqs, err := batch.QueueBulkInsert(
    "INSERT INTO ledger (description, amount) VALUES ($1, $2)",
    0,  // 0 = default chunk size of 100 rows per statement
    "a", 1, "b", 2, "c", 3,
)
// qqs contains one *QueuedQuery (3 rows < default 100)

br := conn.SendBatch(ctx, batch)
ct, err := br.Exec()
// ct.RowsAffected() == 3
err = br.Close()
```

For large inserts the arguments are automatically chunked:

```go
// 250 rows with batchSize=100 → 3 queued queries (100 + 100 + 50)
qqs, err := batch.QueueBulkInsert(
    "INSERT INTO ledger (description, amount) VALUES ($1, $2)",
    100,
    args...,
)
```

Trailing clauses like `ON CONFLICT` or `RETURNING` are preserved on every chunk.

### API

```go
func (b *Batch) QueueBulkInsert(query string, batchSize int, arguments ...any) ([]*QueuedQuery, error)
```

**Parameters:**
- `query` — A single-row INSERT template with positional placeholders (`$1`, `$2`, …).
- `batchSize` — Maximum number of rows per expanded INSERT. Pass `0` for the default (100).
- `arguments` — Flat list of arguments. Must be evenly divisible by the number of placeholders per row.

**Returns:**
- `[]*QueuedQuery` — The queued queries (one per chunk), or `nil` if no arguments are provided.
- `error` — Returned if the query cannot be parsed, arguments don't align, or `batchSize * paramsPerRow` exceeds PostgreSQL's 65535 bind parameter limit.

### Implementation details

| File | Description |
|---|---|
| `batch.go` | `QueueBulkInsert` method on `Batch` and `maxPostgresParameters` constant |
| `bulk_insert.go` | SQL parsing and rewriting helpers: `splitInsertTemplate`, `countInsertParams`, `renumberPlaceholders`, `buildBulkExpandedSQL`, and supporting scanner functions |
| `bulk_insert_test.go` | Integration and unit tests |

The SQL parser correctly handles:
- Single-quoted and double-quoted strings
- Line comments (`--`) and block comments (`/* */`)
- Nested parentheses in the VALUES tuple
- Case-insensitive `VALUES` keyword detection with identifier boundary checks

### Tests

| Test | What it covers |
|---|---|
| `TestBatchQueueBulkInsert` | Basic 3-row insert, verifies `RowsAffected` and row count |
| `TestBatchQueueBulkInsertEmptyArgs` | No arguments → returns `(nil, nil)`, batch unchanged |
| `TestBatchQueueBulkInsertBatchSize` | 250 rows / batchSize 100 → 3 chunks, verifies total count |
| `TestBatchQueueBulkInsertDefaultBatchSize` | 210 rows / default batchSize → 3 chunks |
| `TestBatchQueueBulkInsertExceedsParameterLimit` | batchSize exceeding 65535 params returns error |
| `TestBatchQueueBulkInsertArgMismatch` | Argument count not divisible by params-per-row returns error |
| `TestBatchQueueBulkInsertOnConflict` | `ON CONFLICT DO NOTHING` clause preserved and working |

All integration tests use `pgxtest.RunWithQueryExecModes` to verify behavior across all query execution modes.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -85,6 +85,70 @@ func (b *Batch) Len() int {
 	return len(b.QueuedQueries)
 }
 
+// maxPostgresParameters is the maximum number of bind parameters allowed in a single
+// PostgreSQL query ($1 through $65535).
+const maxPostgresParameters = 65535
+
+// QueueBulkInsert splits the provided arguments into chunks of batchSize rows and queues each
+// chunk as a single multi-value INSERT. The query must be a single-row INSERT template using
+// positional placeholders ($1, $2, ...):
+//
+//	INSERT INTO t (a, b) VALUES ($1, $2)
+//
+// QueueBulkInsert rewrites this into:
+//
+//	INSERT INTO t (a, b) VALUES ($1, $2), ($3, $4), ($5, $6), ...
+//
+// Any clause after the VALUES tuple (e.g. ON CONFLICT, RETURNING) is preserved on every chunk.
+//
+// batchSize is the maximum number of rows per INSERT statement. Pass 0 to use the default of
+// 100. Returns (nil, nil) if no arguments are provided.
+//
+// Returns an error if batchSize * params-per-row exceeds the PostgreSQL bind parameter limit of
+// 65535, if the argument count is not divisible by params-per-row, or if the query cannot be
+// parsed.
+func (b *Batch) QueueBulkInsert(query string, batchSize int, arguments ...any) ([]*QueuedQuery, error) {
+	if len(arguments) == 0 {
+		return nil, nil
+	}
+
+	effectiveBatchSize := 100
+	if batchSize > 0 {
+		effectiveBatchSize = batchSize
+	}
+
+	prefix, inner, suffix, err := splitInsertTemplate(query)
+	if err != nil {
+		return nil, fmt.Errorf("QueueBulkInsert: %w", err)
+	}
+
+	paramsPerRow := countInsertParams(inner)
+	if paramsPerRow == 0 {
+		return nil, fmt.Errorf("QueueBulkInsert: no bind parameters found in VALUES clause")
+	}
+	if len(arguments)%paramsPerRow != 0 {
+		return nil, fmt.Errorf("QueueBulkInsert: argument count (%d) is not divisible by params per row (%d)",
+			len(arguments), paramsPerRow)
+	}
+
+	rowCount := len(arguments) / paramsPerRow
+
+	if effectiveBatchSize*paramsPerRow > maxPostgresParameters {
+		return nil, fmt.Errorf("QueueBulkInsert: batchSize (%d) * params per row (%d) = %d exceeds PostgreSQL bind parameter limit of %d",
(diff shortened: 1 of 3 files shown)


jackc:
> Inserting many rows one at a time — even within a batch — results in one round-trip per row. 

This is not true.

w00j00ng:
> > Inserting many rows one at a time — even within a batch — results in one round-trip per row.
> 
> This is not true.

Sorry, I misunderstood how pgx batching works.

You're right that `Batch` does not cause one round-trip per row — all queries are sent together. The performance difference comes from PostgreSQL server-side overhead (parse/plan/execute per statement), not network round trips.

I've updated the PR description to correct this explanation.

jackc:
Even if this feature were to be accepted, it is implemented at the wrong layer. The fundamental new feature is building a multi-row insert statement. It has nothing to do with batches. You would expect it to be usable for single queries as well.

But that said, I don't think that this is driver layer functionality. It belongs in a layer above that. Which is what I actually did for my personal use, see pgxutil https://pkg.go.dev/github.com/jackc/pgxutil#Insert and https://pkg.go.dev/github.com/jackc/pgxutil#QueueInsert.

w00j00ng:
Thank you

Issue #1999, opened 2024-05-02
Ever Increasing number of Go routines with Pgx v5.5.5 and CockroachDB


victor-ferrer-form3:
**Describe the bug**

Recently we upgraded our software to pgx v5.5.5 and inmediately noticed that the number of Go routines our pods use is ever increasing.
This is a screenshot of our monitoring tool, depicting the `go_goroutines`metric:

![image](https://github.com/jackc/pgx/assets/82882005/4e3247e4-3591-412d-b185-80ffd79085f4)

The point where it starts to crawl up, matches our update to `v5.5.4` to `v5.5.5`.

If we enable Pprof after the pod has spent some hours running we can see this:

```
goroutine profile: total 318
[...]
74 @ 0x43e32e 0x4099ad 0x4095b2 0x9529ec 0x471501
#	0x9529eb	github.com/jackc/pgx/v5/pgconn/internal/ctxwatch.(*ContextWatcher).Watch.func1+0x8b	/build/vendor/github.com/jackc/pgx/v5/pgconn/internal/ctxwatch/context_watcher.go:51
```

**To Reproduce**

If possible, please provide runnable example such as:

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
)

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(context.Background())

	// Your code here...
}
```

Please run your example with the race detector enabled. For example, `go run -race main.go` or `go test -race`.

**Expected behavior**
We would expect those watches to be cancelled/closed and the Go Routines ended.

**Actual behavior**
Context watchers seem not to be finished properly.

**Version**
 - Go: `$ go version` -> 1.22
 - Cockroach DB:
 
```
cockroach version details:
Build Tag:        v23.1.12
Build Time:       2023/11/09 06:15:38
Distribution:     CCL
Platform:         linux amd64 (x86_64-pc-linux-gnu)
Go Version:       go1.19.13
C Compiler:       gcc 6.5.0
Build Commit ID:  d7e9824b4cd6ebf7a8548156f2a772ae6648257d
Build Type:       release
Enabled Assertions: false
(use 'cockroach version --build-tag' to display only the build tag)

```
 
 - pgx: `5.5.5`

victor-ferrer-form3:
Note: Reverting to `v5.5.2` solves the issue.

[...]

PR #2515, opened 2026-03-10
ctxwatch: ContextWatcher goroutine leak

hdonnay:
This fixes a goroutine leak that could occur if Unwatch was deferred a long time or was not called.

The mechanism now uses `context.AfterFunc` and a channel close, rather than a channel send+receive and boolean flags.

Closes: jackc/pgx#1999


jackc:
Thanks. LGTM.

I was concerned about this simply masking a missing Unwatch call so I did a little investigating and found a couple potential paths which I fixed with a subsequent commit.

Issue #2525, opened 2026-03-24
Parsing error when defining `min_protocol_version=3.2`


amotl:
Hi. First things first: Thanks a stack for your excellent work on pgx. It is a true gem.

### Problem
With the new support for the PostgreSQL protocol version 3.2 added with GH-2498 (thank you, @dbussink!), and the corresponding logic to handle erroneous value assignments that balance between both new options `min_protocol_version` vs. `max_protocol_version`, and their corresponding default values, currently, when defining
```
min_protocol_version=3.2
```
... this error is emitted:
```
2026/03/24 01:46:21 cannot parse `postgres://crate@localhost:5432/testdrive?min_protocol_version=3.2`:
min_protocol_version cannot be greater than max_protocol_version
```
The error message does not give the average user any indication about why that happens, while they were innocently trying to connect to the server using a specific minimum PostgreSQL protocol version. 

### Wish
When there are no objections, we think the optimal parsing and validation logic should be adjusted to be more [DWIM](https://en.wikipedia.org/wiki/DWIM), so that it's sufficient to provide `min_protocol_version=3.2` without further ado.

abrightwell:
Seems simple and straight forward enough, I'll put together something for review/discussion later this evening or tomorrow morning. 👍

amotl:
Thank you for resolving this swiftly. 🙇

abrightwell:
🫡

PR #2526, opened 2026-03-26
Update `max_protocol_version` and `min_protocol_version` defaults

abrightwell:
When `max_protocol_version` is not explicitly set, the default value is set to "3.0". This is reasonable and is inline with the same behavior as libpq[1]. Though, what libpq attempts to do that was not being done here is to reasonably upgrade or auto-raise the value of `max_protocol_version` when it has not been explicitly set.

Here, we're providing the same functionality. In libpq, this determination is based on the parsed config values. We do the same here, parsing the values in ParseConfig to validate and determine the appropriate default for max_protocol_version. The string fields on Config are then set accordingly, and re-parsed at connection time as before.

After parsing, when `max_protocol_version` has not been explicitly set, if `min_protocol_version` is greater than 3.0, `max_protocol_version` defaults to `latest`; otherwise it defaults to 3.0 for compatibility with older servers/poolers that don't support NegotiateProtocolVersion.

[1] https://github.com/postgres/postgres/commit/285613c60a7aff5daaf281c67002483b0d26e715

Fixes: #2525


jackc:
LGTM!

PR #2528, opened 2026-04-03
Fix stdlib to return `time.Time` for PostgreSQL `time` columns

### LINKED ISSUE #2508: Can persist an entity, but get am error, when reading using sqlx
**Describe the bug**
I am using sqlx and pgx.
I can persist an entity , but when I try to read it, I get an error
sql: Scan error on column index 2, name "db_timeofday": unsupported Scan, storing driver.Value type string into type *time.Time

**To Reproduce**
Execute the attached main. It will create the db table, insert an entity and receive an error when reading. 

```go
package main

import (
	"fmt"
	"log"
	"os"
	"time"

	"github.com/google/uuid"
	_ "github.com/jackc/pgx/v5/stdlib"
	"github.com/jmoiron/sqlx"
)

const (
	dropTable   = `DROP TABLE csv_table;`
	createTable = `CREATE TABLE csv_table (
               db_uuid           uuid,
               db_number          numeric,
             db_timeOfDay          time,
             CONSTRAINT csv_table_pk PRIMARY KEY (db_uuid)
              );`
)


type Entity struct {
	Uuid      uuid.UUID `db:"db_uuid"`
	Number    int       `db:"db_number"`
	TimeOfDay time.Time `db:"db_timeofday"`
}

func main() {
	dsn := os.Getenv("DATABASE_URL")
	conn, err := sqlx.Connect("pgx", dsn)

	log.SetFlags(log.Lshortfile)
	if err != nil {
		log.Fatal(err)
	}

	dropAndCreateTable(conn)
	persistEntity(conn)
	readEntity(conn)
}

func dropAndCreateTable(db *sqlx.DB) {
	_, err := db.Exec(dropTable)
	_, err = db.Exec(createTable)
	if err != nil {
		log.Fatal(err)
	}
}
func persistEntity(db *sqlx.DB) {
	entities := []Entity{Entity{Uuid: uuid.UUID{}, Number: 4711, TimeOfDay: time.Now()}}

	query := `INSERT INTO csv_table (db_uuid, db_number, db_timeofday) VALUES (:db_uuid, :db_number, :db_timeofday)`
	named, args, err := sqlx.Named(query, entities)
	if err != nil {
		log.Fatal(err)
	}
	rebind := db.Rebind(named)
	_, err = db.Exec(rebind, args...)
	if err != nil {
		log.Fatal(err)
	}
}

func readEntity(db *sqlx.DB) {
	var loaded Entity
	err := db.Get(&loaded, `SELECT db_uuid, db_number,db_timeofday FROM csv_table WHERE db_number = $1 `, 4711)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("Loaded:", loaded)
}

```

Please run the main`.

**Expected behavior**
I expect to read the persisted entity without an error.

**Actual behavior**
The db.Get in readEntity returns an error
main.go:72: sql: Scan error on column index 2, name "db_timeofday": unsupported Scan, storing driver.Value type string into type *time.Time

**Version**
- go version go1.26.0 linux/amd64
- PostgreSQL 17.4 (Debian 17.4-1.pgdg120+2) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit  
 - pgx: github.com/jackc/pgx/v5 v5.8.0

**Additional context**
 My go.mod:
module pgxexample

go 1.26
require (
	github.com/google/uuid v1.6.0
	github.com/jackc/pgx/v5 v5.8.0
	github.com/jmoiron/sqlx v1.4.0
)
require (
	github.com/jackc/pgpassfile v1.0.0 // indirect
	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
	github.com/jackc/puddle/v2 v2.2.2 // indirect
	golang.org/x/sync v0.17.0 // indirect
	golang.org/x/text v0.29.0 // indirect
)

abrightwell:
The `TimeOID` case was missing from the stdlib `Rows.Next()` switch, causing `time` columns to fall through to the `default` case. The `default` case scans into a `string`, which is not a valid `driver.Value` for temporal types. When `database/sql` attempts to assign the resulting `string` to a `*time.Time` scan destination, it fails with:

```
sql: Scan error on column index 0, name "t": unsupported Scan, storing
driver.Value type string into type *time.Time
```

To fix this, a `TimeOID` case is added that scans into `pgtype.Time` and converts microseconds since midnight to `time.Time`, matching the existing conversion in `timeWrapper.ScanTime`
(`pgtype/builtin_wrappers.go`).

Fixes: #2508

### DIFF
--- a/stdlib/sql.go
+++ b/stdlib/sql.go
@@ -712,7 +712,7 @@ func (r *Rows) ColumnTypeScanType(index int) reflect.Type {
 		return reflect.TypeFor[bool]()
 	case pgtype.NumericOID:
 		return reflect.TypeFor[float64]()
-	case pgtype.DateOID, pgtype.TimestampOID, pgtype.TimestamptzOID:
+	case pgtype.DateOID, pgtype.TimeOID, pgtype.TimestampOID, pgtype.TimestamptzOID:
 		return reflect.TypeFor[time.Time]()
 	case pgtype.ByteaOID:
 		return reflect.TypeFor[[]byte]()
@@ -816,6 +816,38 @@ func (r *Rows) Next(dest []driver.Value) error {
 						return nil, err
 					}
 					return d, nil
+				}
+			case pgtype.TimeOID:
+				var d pgtype.Time
+				scanPlan := m.PlanScan(dataTypeOID, format, &d)
+				r.valueFuncs[i] = func(src []byte) (driver.Value, error) {
+					err := scanPlan.Scan(src, &d)
+					if err != nil {
+						return nil, err
+					}
+					if !d.Valid {
+						return nil, nil
+					}
+
+					// The microseconds-to-time.Time conversion here is duplicated from
+					// timeWrapper.ScanTime in pgtype/builtin_wrappers.go. timeWrapper is
+					// unexported, so we inline the conversion.
+
+					var maxRepresentableByTime int64 = 24*60*60*1000000 - 1
+					if d.Microseconds > maxRepresentableByTime {
+						return nil, fmt.Errorf("%d microseconds cannot be represented as time.Time", d.Microseconds)
+					}
+
+					usec := d.Microseconds
+					hours := usec / (60 * 60 * 1000000)
+					usec -= hours * (60 * 60 * 1000000)
+					minutes := usec / (60 * 1000000)
+					usec -= minutes * (60 * 1000000)
+					seconds := usec / 1000000
+					usec -= seconds * 1000000
+					ns := usec * 1000
+
+					return time.Date(2000, 1, 1, int(hours), int(minutes), int(seconds), int(ns), time.UTC), nil
 				}
 			case pgtype.TimestampOID:
 				var d pgtype.Timestamp
--- a/stdlib/sql_test.go
+++ b/stdlib/sql_test.go
@@ -567,6 +567,17 @@ func TestConnQueryRowUnknownType(t *testing.T) {
 	})
 }
 
+// https://github.com/jackc/pgx/issues/2508
+func TestConnQueryRowTimeScanIntoTimeTime(t *testing.T) {
+	testWithAllQueryExecModes(t, func(t *testing.T, db *sql.DB) {
+		var actual time.Time
+		err := db.QueryRow("select '22:45:00'::time").Scan(&actual)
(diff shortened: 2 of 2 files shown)


abrightwell:
As mentioned in #2508, the only potential concern is that those who are currently scanning into a string would see different values. The time would be correct, but it would not be the same string.

For instance:

Before: "22:45:00"
After: "2000-01-01T22:45:00Z"

jackc:
> As mentioned in #2508, the only potential concern is that those who are currently scanning into a string would see different values. The time would be correct, but it would not be the same string.
> 
> For instance:
> 
> Before: "22:45:00" After: "2000-01-01T22:45:00Z"

I know you mentioned a `string` issue before, but it didn't quite register with me. That is a bigger incompatibility than I realized. If I understand correctly, there would no longer be anyway to get the actual text aside from casting to text in your SQL query. That would be unfortunate. Worse, a value like `24:00:00` might not work at all. This isn't a problem with the native pgx interface because it knows the source and destination types. With database/sql everything has to go through driver.Value without knowledge of where it will end up.

We already have a few of these edge cases, like for `Infinity`, but I'm not sure that we should add more.

It's unfortunate that https://github.com/golang/go/issues/67546 is still pending. Whenever it is finally resolved that will solve all these issues.

abrightwell:
> We already have a few of these edge cases, like for Infinity, but I'm not sure that we should add more.

Yeah, I'm not so sure adding such a case would make sense. Especially if it would/could be overcome by changes to the language that are on the horizon as mentioned.

> It's unfortunate that https://github.com/golang/go/issues/67546 is still pending. Whenever it is finally resolved that will solve all these issues.

Unfortunate indeed, that would be really nice to have for sure. I'm not terribly familiar with the CL process for golang, but I've been spending some time walking through what it would take to possibly help in getting this one unstuck.

abrightwell:
Closing given the above mentioned incompatibility concerns.

As well, it seems like https://github.com/golang/go/issues/67546 has been [accepted](https://github.com/golang/go/issues/67546#issuecomment-4347540757) 🥳 , so hopefully we'll see that land soon and this will no longer be a need or concern. 😁

PR #2532, opened 2026-04-13
pgconn: fix parseKeywordValueSettings rejecting trailing whitespace

### LINKED ISSUE #2284: Incorrect whitespace handling when parsing key-value connection strings
**Describe the bug**
The function [`parseKeywordValueSettings`](https://github.com/jackc/pgx/blob/master/pgconn/config.go#L540-L610) in the `pgconn` package incorrectly handles whitespace when parsing keyword/value settings. Specifically:  

1. It fails if the string has more than one trailing whitespace, as it expects another key-value pair after it.  
2. For **quoted values**, the behavior is inconsistent. The function does not consume a trailing space after a quoted value, instead relying on the next iteration's trimming. In other words, quoted values cannot be followed by trailing whitespaces at all.

This results in unexpected parsing failures for valid connection strings (according to the [documentation](https://www.postgresql.org/docs/9.5/libpq-connect.html#AEN43131) and to `psql`'s behaviour). The logic should be adjusted to allow trailing spaces without requiring another key-value pair.  

I'd be happy to provide a fix for this if we agree on a strategy.

**To Reproduce**
Here's a minimal example to reproduce the issue:

```go
package main

import (
	"context"
	"log"

)

func main() {
	_, err := pgxpool.New(context.Background(), "dbname=foo  ")
	log.Fatal(err)
}
```

You can also find a more detailed breakdown of the different ways to trigger the error [here](https://go.dev/play/p/FSzCxKiIuI6), isolating the root cause to the aforementioned function.

**Expected behavior**
```go
parseKeywordValueSettings(`port=5432 host=foo  `) // map[string]string{"host":"foo", "port":"5432"}
```

**Actual behavior**
```go
parseKeywordValueSettings(`port=5432 host=foo  `) // invalid keyword/value
```

**Version**
 - Go: `$ go version` -> `go version go1.24.0 darwin/arm64`
 - PostgreSQL: `$ psql --no-psqlrc --tuples-only -c 'select version()'` -> `PostgreSQL 14.17 (Ubuntu 14.17-1.pgdg22.04+1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit`
 - pgx: `$ grep 'github.com/jackc/pgx/v[0-9]' go.mod` -> `github.com/jackc/pgx/v5 v5.7.0`

alliasgher:
## Summary

Fixes #2284.

`parseKeywordValueSettings` rejected valid keyword/value connection strings that had trailing whitespace or multiple spaces between values.

### Root cause

After consuming each value, the parser advanced by one byte (`s[end+1:]` for unquoted values, `s[end+1:]` after the closing quote for quoted values). This skipped a single whitespace character but left any additional leading spaces in `s`. On the next iteration, the loop found no `=` in `"  "` and returned `"invalid keyword/value"`.

The same issue occurred with quoted values followed by **any** trailing whitespace, because `s[end+1:]` only moved past the closing `'` and left the spaces.

### Fix

- Trim leading whitespace from `s` once before entering the loop so that an entirely-whitespace input exits cleanly.
- After consuming each value (unquoted and quoted), use `strings.TrimLeft(…)` instead of `[end+1:]` to consume all subsequent whitespace. This matches `libpq`'s behavior and the PostgreSQL documentation, which describes the connection string syntax as allowing spaces between key-value pairs.

### Tests added

`TestParseConfigKVTrailingWhitespace` covers:
- single / multiple trailing spaces
- trailing tab
- quoted value followed by trailing spaces
- two key-value pairs with trailing spaces

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -631,6 +631,9 @@ func parseKeywordValueSettings(s string) (map[string]string, error) {
 		"dbname": "database",
 	}
 
+	// Trim any leading whitespace so that the loop exits cleanly when only
+	// spaces remain (e.g. trailing spaces after the last value).
+	s = strings.TrimLeft(s, " \t\n\r\v\f")
 	for len(s) > 0 {
 		var key, val string
 		eqIdx := strings.IndexRune(s, '=')
@@ -655,11 +658,9 @@ func parseKeywordValueSettings(s string) (map[string]string, error) {
 				}
 			}
 			val = strings.Replace(strings.Replace(s[:end], "\\\\", "\\", -1), "\\'", "'", -1)
-			if end == len(s) {
-				s = ""
-			} else {
-				s = s[end+1:]
-			}
+			// Consume the value and trim any subsequent whitespace so that
+			// multiple trailing spaces don't cause a spurious parse failure.
+			s = strings.TrimLeft(s[end:], " \t\n\r\v\f")
 		} else { // quoted string
 			s = s[1:]
 			end := 0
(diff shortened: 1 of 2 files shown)


jackc:
Thanks!

PR #2535, opened 2026-04-15
doc: docker minimal setup

Tubelight30:
I added a minimal PostgreSQL-in-Docker setup to CONTRIBUTING.md so someone can run `go test ./...` without installing PostgreSQL locally or using the devcontainer.
When I tried to set up the repo with a simple Docker Postgres container, I ran into two things that can be confusing for first-time contributors:
- go test can show (cached) even if the external DB environment changed, so I documented using `go test -count=1 ./...` when validating DB-related tests.
- Docker Postgres images often default to TimeZone=UTC while the Go process uses the host timezone (time.Local), which can cause timestamptz round-trip comparisons to fail. I documented running tests in UTC with TZ=UTC as a simple, deterministic workaround.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,35 @@ simplify environment variable handling.
 The easiest way to start development is with the included devcontainer. It includes containers for each supported
 PostgreSQL version as well as CockroachDB. `./test.sh all` will run the tests against all database types.
 
+### Using PostgreSQL in Docker (Minimal)
+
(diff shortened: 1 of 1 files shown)


404salad (approved):

jackc:
> Docker Postgres images often default to TimeZone=UTC while the Go process uses the host timezone (time.Local), which can cause timestamptz round-trip comparisons to fail. I documented running tests in UTC with TZ=UTC as a simple, deterministic workaround.

That sounds like a test bug that should be fixed rather than documenting a workaround.

PR #2536, opened 2026-04-17
feat: Add StructArgs and StrictStructArgs for named query arguments

### LINKED ISSUE #2406: Use a struct for named args

When querying data with pgx, I can quite easily load the data into structs using the various RowTo functions.
However, when inserting data, I have to repeat field names a lot:

```
_, err := c.db.Exec(ctx, `
	INSERT INTO foo (col)
	VALUES (@col)
`, pgx.StrictNamedArgs{
	"col":                 m.Col,
})
```

A feature similar to (Strict)NamedArgs, but instead of passing a map, passing a struct with the appropriate struct tags:

```
type Foo struct {
	Col string `db:"col"`
}

_, err := c.db.Exec(ctx, `
	INSERT INTO foo (col)
	VALUES (@col)
`, pgx.StructArgs(m))
```

None.

**Additional context**

None.

Tubelight30:
Hi
Currently, `NamedArgs` and `StrictNamedArgs` require manually building a map of argument names to values. With this change, callers can pass a struct or pointer to a struct and pgx will derive named arguments from exported fields.
Implementation Features:
- `StructArgs` converts exported struct fields into `NamedArgs`
- `StrictStructArgs` converts exported struct fields into `StrictNamedArgs`
- Field names are taken from the `db` struct tag when present (has options like `db:"name,omitempty";` only the part before the comma is used).
- If no `db` tag is present, falls back to the Go field name.
- Ignores unexported fields.
- Both struct values and pointers to structs are supported.

Design Note:
- embedded structs are intentionally skipped rather than flattened. This keeps the behavior predictable and avoids implicit recursive field expansion.

Example:
```
type Foo struct {
    ID   int    `db:"id"`
    Name string `db:"name,omitempty"`
}

f := Foo{ID:   1,Name: "alice"}
_, err := conn.Exec(ctx,
    "insert into foo (id, name) values (@id, @name)",
    pgx.StructArgs(f),
)
```

Fixes: #2406

### DIFF
--- a/named_args.go
+++ b/named_args.go
@@ -118,20 +118,10 @@ func structArgs(sa any) (map[string]any, error) {
 	return out, nil
 }
 
+// dbTagKey derives the named-argument key for a struct field. Tag parsing matches
+// RowToStructByName* in rows.go (structTagKey, Lookup, comma options, db:"-").
+// Anonymous embedded structs are skipped without flattening (unlike row scanning).
 func dbTagKey(sf reflect.StructField) (key string, ok bool, err error) {
-	tag := sf.Tag.Get("db")
-	if tag == "-" {
-		return "", false, nil
-	}
-
-	if tag != "" {
-		tag, _, _ = strings.Cut(tag, ",")
-		if tag == "" {
-			return "", false, fmt.Errorf("field %s has empty `db` tag", sf.Name)
-		}
-		return tag, true, nil
-	}
-
 	if sf.Anonymous {
 		ft := sf.Type
 		if ft.Kind() == reflect.Pointer {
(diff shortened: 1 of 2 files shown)


named_args.go
```diff
@@ -34,6 +36,115 @@ func (sna StrictNamedArgs) RewriteQuery(ctx context.Context, conn *Conn, sql str
...
+			return nil, fmt.Errorf("duplicate StructArgs key %q", key)
+		}
+
+		out[key] = v.Field(i).Interface()
+	}
+
+	return out, nil
+}
+
+func dbTagKey(sf reflect.StructField) (key string, ok bool, err error) {
```
mxey, line 124: Is there no existing code in the library that does this, since there’s already other uses of struct tags?

Tubelight30: I’m not fully sure yet. I’ll check the existing code for any similar struct-tag handling and get back to you.

Tubelight30: > Is there no existing code in the library that does this, since there’s already other uses of struct tags?

Good catch — I found related logic in `rows.go` and updated my code to follow that style.
I didn’t reuse the `RowToStructByName` logic directly since that logic flattens anonymous embedded structs recursively, while `StructArgs` is meant to stay flat for @ placeholders. Any thoughts on this approach?

mxey: I don’t have an opinion on which behavior is better but I think both should behave the same so the library is consistent.

Tubelight30: That makes sense. My initial thought was that for a simple “named args” helper, recursively flattening might be too complex. I’ll look into whether I should make it behave consistently with the existing logic.

PR #2542, opened 2026-04-25
stdlib: implement driver.RowsColumnScanner for Go 1.27

jackc:
Go 1.27 adds the driver.RowsColumnScanner interface, which lets a driver scan a column directly into the user's destination. This means stdlib users can now scan PostgreSQL arrays, ranges, multiranges, and other pgtype-aware values directly into Go slices and pgtype types without the *pgtype.Map.SQLScanner adapter.

ScanColumn first tries m.Scan for pgx-native handling, then falls back to the newly public sql.ConvertAssign on a driver.Value produced by the column codec. The fallback preserves database/sql conversion semantics (e.g., scanning a binary bool into *string yields "true"/"false") for destinations pgx does not recognize.


jackc:
I've updated this PR for what finally landed in Go 1.27 RC1. Should be ready to go, but extra eyes and testers would be great.

xrn:
Hey @jackc first of all as always thank you very much for contribution. Go 1.27 is released, do you think you would be merging this change?

PR #2543, opened 2026-04-27
Enable unconvert linter

alexandear:
Simplify code with `unconvert` linter:

```
go install github.com/mdempsky/unconvert@latest
unconvert -apply ./...
```


jackc:
LGTM

PR #2546, opened 2026-04-30
fix(pgxpool): pass base context to BeforeConnect from background healthcheck

### LINKED ISSUE #2545: `BeforeConnect` receives `context.Background()` (instead of the pool context) from the healthcheck path
**Describe the bug**
Long story short, we use `BeforeConnect` to implement IAM Authentication with AWS / Aurora, and I traced back some panics to a missing logger in the context passed to BeforeConnect. I would expect the hook to receive a useful context (for logging, tracing, fetching credentials, ...), but while that happens for `Pool.Acquire`, the background goroutine refilling the pool to satisfy the minimum size calls it with a bare `context.Background()`.

**To Reproduce**

```go
package main

import (
	"context"
	"fmt"
	"time"

	"github.com/jackc/pgx/v5"
)

type key int

const myKey key = 0

func main() {
	ctx := context.WithValue(context.Background(), myKey, "value")

	poolConfig, _ := pgxpool.ParseConfig("postgres://user:password@127.0.0.1:1/mydb")

	// Fail fast
	poolConfig.ConnConfig.ConnectTimeout = 200 * time.Millisecond

	// Prepare connections in the background
	poolConfig.MinConns = 1
	poolConfig.HealthCheckPeriod = 300 * time.Millisecond

	var calls int
	poolConfig.BeforeConnect = func(bcCtx context.Context, cc *pgx.ConnConfig) error {
		calls++
		if v := bcCtx.Value(myKey); v == nil {
			fmt.Printf("call %d: BeforeConnect ctx MISSING value\n", calls)
		} else {
			fmt.Printf("call %d: BeforeConnect ctx has value=%q\n", calls, v)
		}
		return nil
	}

	pool, err := pgxpool.NewWithConfig(ctx, poolConfig)
	if err != nil {
		panic(err)
	}
	defer pool.Close()

	// Let the background healthcheck fire a few times.
	time.Sleep(2 * time.Second)
}
```

**Expected behavior**
```
call 1: ctx has value="value"
call 2: ctx has value="value"
call 3: ctx has value="value"
```

**Actual behavior**
```
call 1: ctx has value="value"
call 2: ctx MISSING value
call 3: ctx MISSING value
```

Call 1 comes from the initial idle creation goroutine in `NewWithConfig`. The subsequent ones come from `backgroundHealthCheck` + `checkMinConns`.

**Impact**
Anyone relying on `context.Value` propagation in `BeforeConnect` (e.g. injecting a request-scoped logger or AWS credential cache) sees inconsistent errors: the hook works on `Acquire`, it works on initial pool warm-up, but it randomly breaks on background refills. In our case it caused a nil-pointer panic while trying to log a message about the authentication.

Either re-use the context passed to `NewWithConfig`, expose a new `Config.HealthCheckContext`, or at the very least document the behavior to avoid unpleasant surprises. I can have a go at the first option if there's agreement.

**Version**
 - Go: go1.26.2 darwin/arm64
 - pgx: v5.9.1
 - puddle: v2.2.2

blackwell-systems:
## Description

`checkMinConns` passes `context.Background()` to `createIdleResources`, which means the `Constructor` function (and therefore `BeforeConnect`) receives a bare context with no values. Any hook that depends on context values from the original `NewWithConfig` call (loggers, AWS IAM credential caches, tracing spans) breaks silently on background refills while working correctly on `Acquire` and initial pool warm-up.

## Fix

Store the context passed to `NewWithConfig` on the `Pool` struct (`baseCtx`) and use it in `checkMinConns` instead of `context.Background()`. This matches the behavior of the initial idle resource creation at line 337, which already uses the constructor context.

The change is 3 lines: one new field on `Pool`, one assignment in `NewWithConfig`, and replacing `context.Background()` with `p.baseCtx` in `checkMinConns`.

## Impact

Any pgxpool user with `MinConns > 0` or `MinIdleConns > 0` who uses `BeforeConnect` with context values is affected. The most common case is AWS Aurora IAM authentication, where the credential refresh logic needs a logger from the context.

Fixes #2545

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -251,7 +251,7 @@ func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
 		healthCheckPeriod:     config.HealthCheckPeriod,
 		healthCheckChan:       make(chan struct{}, 1),
 		closeChan:             make(chan struct{}),
-		baseCtx:               ctx,
+		baseCtx:               context.WithoutCancel(ctx),
 	}
 
 	if t, ok := config.ConnConfig.Tracer.(AcquireTracer); ok {
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -103,6 +103,7 @@ type Pool struct {
 	healthCheckTimer *time.Timer
 
 	healthCheckChan chan struct{}
+	baseCtx         context.Context
 
 	acquireTracer AcquireTracer
 	releaseTracer ReleaseTracer
@@ -250,6 +251,7 @@ func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
 		healthCheckPeriod:     config.HealthCheckPeriod,
 		healthCheckChan:       make(chan struct{}, 1),
 		closeChan:             make(chan struct{}),
+		baseCtx:               ctx,
 	}
 
 	if t, ok := config.ConnConfig.Tracer.(AcquireTracer); ok {
@@ -561,7 +563,7 @@ func (p *Pool) checkMinConns() error {
 	stat := p.Stat()
 	toCreate := max(p.minConns-stat.TotalConns(), p.minIdleConns-stat.IdleConns())
 	if toCreate > 0 {
-		return p.createIdleResources(context.Background(), int(toCreate))
+		return p.createIdleResources(p.baseCtx, int(toCreate))
 	}
 	return nil
 }


pgxpool/pool.go
```diff
@@ -250,6 +251,7 @@ func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
 		healthCheckPeriod:     config.HealthCheckPeriod,
 		healthCheckChan:       make(chan struct{}, 1),
 		closeChan:             make(chan struct{}),
+		baseCtx:               ctx,
```
abrightwell: Storing a context for reuse doesn't feel right. One concern is if it could be cancelled or carries a deadline, as all future refills would break. Removing those prior to storing would be safer, IMO:

```suggestion
		baseCtx:               context.WithoutCancel(ctx),
```

https://pkg.go.dev/context#WithoutCancel

That said, I wonder if the underlying issue is an incorrect use of `context`, ie., request-scoped vs. application-scoped. `BeforeConnect` seems like the right place for per-connection `context` enrichment, since it runs on every connection creation, whether for warmup or refill of the pool.

blackwell-systems: Good catch, updated to `context.WithoutCancel(ctx)`. That prevents a request-scoped context from breaking future refills while still preserving the values that `BeforeConnect` callbacks depend on (e.g., AWS IAM auth tokens).

On the context scope point: agreed, `BeforeConnect` is the right place for per-connection enrichment. The tricky part is that background healthchecks pass `context.Background()`, which drops all values from the original context. So users who store auth credentials in the context passed to `NewWithConfig` see it work on first connect but silently break on pool refill. `WithoutCancel` preserves the values without carrying over cancellation or deadlines.

abrightwell: > So users who store auth credentials in the context passed to NewWithConfig see it work on first connect but silently break on pool refill.

Credentials passed through the `context` seems like anti-pattern. I'd argue it points to the wrong injection point entirely. If credentials are fetched in `BeforeConnect`, they're available on every connection attempt regardless of what triggered it.

blackwell-systems: Fair point. Fetching credentials directly in `BeforeConnect` is cleaner and avoids the context-propagation question entirely. That said, the current behavior (silently dropping to `context.Background()` on refill) is surprising regardless of how credentials are injected. Happy to adjust the approach if jackc has a preference on the design direction.

Issue #2544, opened 2026-04-29
LoadType during AfterConnect fails with context timeout leading to queries failing


jaltavilla:
**Describe the bug**
I haven't been able to fully figure out what is happening here, so this will probably be a bit unclear. Sorry. There might even be two issues interacting here.
We setup our pgxpool to have an AfterConnect function. This function calls LoadTypes on the connection to prepare it for use. Frequently LoadTypes is resulting in a context timeout and disconnecting from the postgres server. However, it seems like this doesn't always surface to caller and then leads to queries failing with "failed to deallocate cached statement(s): conn closed".

**To Reproduce**
Unfortunately, I can't really provide a repro case as it's not deterministic. I was able to get this to trigger while running with the race detector enabled and it didn't detect any issues. I setup pgx with a `tracelog.TraceLog` at TraceLevel and also enabled tracing of the front end with `conn.PgConn().Frontend().Trace`. Attached is a log from when this occurred. In prior debugging attempts I examined the context passed to the AfterConnect function. It had ~1m10s left till it timed out and the actual call only took ~63ms before the context timeout occurred.

[pgx-trace.log](https://github.com/user-attachments/files/27213493/pgx-trace.log)

You can see the first error on line 14659 or by searching for "time="2026-04-29T17:33:47Z" level=error". 

**Expected behavior**
LoadTypes succeeds when there is a connection to the database.
Failures in AfterConnect / LoadType don't result in top level pgx calls such as `err := pool.SendBatch(ctx, batch).Close()` connection closed error when there is no underlying problems preventing connection.

**Actual behavior**
Queries return `failed to deallocate cached statement(s): conn closed` sometimes and some connection attempts fail.

**Version**
 - Go: go version go1.26.2 linux/amd64
 - PostgreSQL: PostgreSQL 17.9 (Debian 17.9-1.pgdg13+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
 - pgx: github.com/jackc/pgx/v5 v5.9.2

**Additional context**
The behavior occurs even when run locally. On a single machine there is a primary database and an asynchronous streaming replica. There are two pgbouncer instances running, one for each database. All connections are made through pgbouncer in transaction mode.

The behavior seems somewhat timing dependent. Running go test with `-v` reduces the frequency with which problems occur. I was only able to get it to occur with pgx logs enabled by logging to memory and then writing it to a file at the end. However, the race detector found no issues.

[...]

PR #2549, opened 2026-05-05
Fix LoadTypes error handling

jaltavilla:
LoadTypes was missing a rows.Err() check at the end. Fixed by replacing manual loop with ForEachRow.

This fixes one of the issues in #2544


jackc:
Can we just add the error check rather than changing the loop type?

jaltavilla:
If that's how you would prefer it to be fixed, I can make that change. 

It seems like using the helper that was created to remove the need for manual looping is safer. It would have prevented this bug as well as prevents future changes from introducing similar ones. Is there a downside to it?

jackc:
> If that's how you would prefer it to be fixed, I can make that change.

> It seems like using the helper that was created to remove the need for manual looping is safer. It would have prevented this bug as well as prevents future changes from introducing similar ones. Is there a downside to it?

I merged the other one. No downside in new code. Just wanted to go for the smaller diff.

PR #2554, opened 2026-05-10
pgconn: use fresh context for fallback connection in connectPreferred

### LINKED ISSUE #2172: Connection timeout wrongly applied to non-preferred hosts
**Describe the bug**
(Related to the testing I mentioned #2171 but this appears to be a straightforward bug.)

When connecting with a `target_session_attrs` preference that cannot be satisfied and the final host times out, the connection to the non-preferred host is not attempted and falsely times out.

**To Reproduce**

1. Create a connection string with two hosts and a preference for the second; e.g. a primary and standby with `target_session_attrs=prefer-standby`
2. Block network traffic from the client to the standby.
3. Configure a connection timeout for e.g. 1 second.
4. Execute a query.

**Expected behavior**
After 1 second the connection to the standby times out, the primary is chosen instead, and the query executes.

**Actual behavior**
After 1 second the query returns a timeout.

**Version**
 - go version go1.23.2 linux/amd64
 - PostgreSQL: PostgreSQL 14.13 (Ubuntu 14.13-1.pgdg22.04+1) on aarch64-unknown-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
 - pgx: v5.7.1

**Additional context**
From my understanding of the code, `connectPreferred` resets the timeout for each new host as it iterates the possible targets. However, once it's decided to fall back to `fallbackConnectOneConfig`, it reuses whatever the last `ctx` it had from the initial pass was. If this had timed out, the new connection will start in a timed-out state. It should instead probably iterate the entire again with fresh contexts based on `octx`. Or, as an optimization, the first non-preferred connection could be returned rather than the config and returned directly if no other preferred one is found / closed if a preferred one is found.

c-tonneslan:
When `ConnectTimeout` is set, `connectPreferred` creates a per-host timeout context (derived from `octx`) for each host in the loop. After the loop finishes, `ctx` holds whatever context was used for the last host, which may have already expired by the time we get to the fallback.

The fallback path (for `prefer-standby` / `prefer-primary`) was passing that stale `ctx` to `connectOne`, so it would immediately get a deadline exceeded error even though the caller's original context still had time left.

The fix creates a fresh `ConnectTimeout`-bounded context from `octx` for the fallback, same pattern as the loop uses for each host.

Fixes #2172

### DIFF
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -292,7 +292,13 @@ func connectPreferred(ctx context.Context, config *Config, connectOneConfigs []*
 	}
 
 	if fallbackConnectOneConfig != nil {
-		pgConn, err := connectOne(ctx, config, fallbackConnectOneConfig, true)
+		fallbackCtx := octx
+		if config.ConnectTimeout != 0 {
+			var cancel context.CancelFunc
+			fallbackCtx, cancel = context.WithTimeout(octx, config.ConnectTimeout)
+			defer cancel()
+		}
+		pgConn, err := connectOne(fallbackCtx, config, fallbackConnectOneConfig, true)
 		if err == nil {
 			return pgConn, nil
 		}


jackc:
LGTM

PR #2556, opened 2026-05-12
pgconn: preserve full error chain in normalizeTimeoutError

### LINKED ISSUE #2383: `normalizeTimeoutError` is overwriting dial error
**Describe the bug**
I'm using a proxying driver here for Google CloudSQL. I'm trying to handle some errors in a way that provides a clear reason to the user.

When it attempts to dial here:
https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/pgconn/pgconn.go#L320

I get a `DialError` that wraps an `net.OpError` that wraps a `poll.DeadlineExceededError`: `i/o timeout`. And that error has useful information about what is going on behind the scenes. It looks like this:

Image

However, this is passed to `newPerDialConnectError` and then `normalizeTimeoutError`. And eventually hands here:
https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/pgconn/errors.go#L145

Which drops the useful error and loses all useful information.

Is there a way to wrap this error so that useful information isn't lost downstream?

Interestingly, its not hitting my context timeout - in fact my context is dropped and that at least partially seems to be due to how `database/sql` drops the context here:
https://github.com/golang/go/blob/ac803b5949f6dbc5bfa559afe506d35f9e1b3195/src/database/sql/sql.go#L809
And then pgx generates a new context here:
https://github.com/jackc/pgx/blob/61d3c965ad442cc14d6b0e39e0ab3821f3684c03/stdlib/sql.go#L325

So ultimately 

**To Reproduce**
I don't have a full, runnable example, but this is the entrypoint back to pgx

```go
	db, err := sql.Open(sqlDriverPostgresIam, dsn)
	if err != nil {
		return fmt.Errorf("failed to open database connection: %w", err)
	}

	var result int
	err = db.QueryRowContext(ctx, "SELECT 1").Scan(&result)
```

**Expected behavior**
I'd like to return information from the upstream error, which has information on the IP that it cannot connect to, etc.

**Actual behavior**
I'm getting a less useful and even confusing error about failing to connect to localhost, which is just the proxy part of the driver.

**Version**
 - Go: `go version go1.24.7 darwin/arm64`
 - PostgreSQL: 17
 - pgx: `v5.7.6`

**Additional context**
Driver: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/postgres/pgxv5/postgres.go

c-tonneslan:
Fixes #2383.

When a network timeout isn't caused by a context cancellation or deadline exceeded, `normalizeTimeoutError` was wrapping only the `net.Error` found via `errors.As`, discarding any outer context. For example, a `DialError` wrapping a `net.OpError` wrapping `poll.DeadlineExceededError` would have the `DialError` (including its address info) stripped off.

The fix is to wrap `err` (the full original error) rather than `netErr` (the inner `net.Error` extracted by `errors.As`). This keeps the entire chain intact so callers can still inspect it. The `net.Error` is still reachable via `errors.As` since it's embedded in the chain.

Added a test that verifies a `wrappedDialError{addr}` wrapping a `timeoutNetError` is still unwrappable after going through `normalizeTimeoutError`.

### DIFF
--- a/pgconn/errors.go
+++ b/pgconn/errors.go
@@ -144,7 +144,7 @@ func normalizeTimeoutError(ctx context.Context, err error) error {
 		} else if ctx.Err() == context.DeadlineExceeded {
 			return &errTimeout{err: ctx.Err()}
 		} else {
-			return &errTimeout{err: netErr}
+			return &errTimeout{err: err}
 		}
 	}
 	return err
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -292,7 +292,13 @@ func connectPreferred(ctx context.Context, config *Config, connectOneConfigs []*
 	}
 
 	if fallbackConnectOneConfig != nil {
-		pgConn, err := connectOne(ctx, config, fallbackConnectOneConfig, true)
+		fallbackCtx := octx
+		if config.ConnectTimeout != 0 {
+			var cancel context.CancelFunc
+			fallbackCtx, cancel = context.WithTimeout(octx, config.ConnectTimeout)
+			defer cancel()
+		}
+		pgConn, err := connectOne(fallbackCtx, config, fallbackConnectOneConfig, true)
 		if err == nil {
 			return pgConn, nil
 		}
--- a/pgconn/pgconn_private_test.go
+++ b/pgconn/pgconn_private_test.go
@@ -1,9 +1,14 @@
 package pgconn
 
 import (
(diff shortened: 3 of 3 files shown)


jackc:
LGTM

PR #2558, opened 2026-05-15
pgconn: add configurable max message body length

carter-ya:
## Summary

- Add `pgconn.Config.MaxBodyLen` and apply it to `pgproto3.Frontend` during `ConnectConfig` and `Construct`.
- Add regression coverage for configured body length limits.

## Rationale

`pgproto3.Frontend` already supports `SetMaxBodyLen`, but `pgconn` users currently need to reach through to the protocol frontend to configure it. Exposing the setting on `pgconn.Config` makes it possible to reject unreasonable message bodies before `chunkReader` allocates the buffer.

The default remains `0`, so no maximum is enforced unless callers opt in. This preserves the existing behavior while allowing applications to configure a defensive limit.

## Testing

- `go test ./pgproto3 ./pgconn -run 'TestFrontendReceiveExceededMaxBodyLen|TestBackendReceiveExceededMaxBodyLen|TestConnectConfigMaxBodyLen|TestConfigCopyReturnsEqualConfig'`

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -44,6 +44,11 @@ type Config struct {
 	LookupFunc     LookupFunc // e.g. net.Resolver.LookupHost
 	BuildFrontend  BuildFrontendFunc
 
+	// MaxBodyLen is the maximum length of a PostgreSQL wire protocol message body in octets. If a message body exceeds
+	// this length, reading the message will fail with pgproto3.ExceededMaxBodyLenErr. The default value is 0, which
+	// means no maximum is enforced.
+	MaxBodyLen int
+
 	// BuildContextWatcherHandler is called to create a ContextWatcherHandler for a connection. The handler is called
 	// when a context passed to a PgConn method is canceled.
 	BuildContextWatcherHandler func(*PgConn) ctxwatch.Handler
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -44,10 +44,10 @@ type Config struct {
 	LookupFunc     LookupFunc // e.g. net.Resolver.LookupHost
 	BuildFrontend  BuildFrontendFunc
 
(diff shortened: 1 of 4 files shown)


jackc:
Is `MaxBodyLen` the right name for this at the `pgconn.Config` layer? It has a clear meaning at the `pgproto3.Frontend` layer. But at the connection / config layer I'm not sure I would immediately understand what it meant. 🤷

carter-ya:
> Is `MaxBodyLen` the right name for this at the `pgconn.Config` layer? It has a clear meaning at the `pgproto3.Frontend` layer. But at the connection / config layer I'm not sure I would immediately understand what it meant. 🤷

  Good point. `MaxBodyLen` mirrors the `pgproto3.Frontend` API, but I agree it may be too protocol-internal for `pgconn.Config`.

  Would you prefer one of these directions?

  1. Keep it as a top-level `pgconn.Config` field, but use a more explicit name, e.g. `MaxProtocolMessageBodyLen`.
  2. Add a small protocol namespace, e.g. `Config.Protocol.MaxMessageBodyLen`, so future protocol-level frontend options have a clearer home.

  My current leaning is option 1 as the smallest API change, but option 2 may be cleaner if you expect more protocol-level knobs to be exposed later.

jackc:
One additional bit of information: this already can be configured on connection with `Config.BuildFrontend`. This feature doesn't enable new functionality, it makes it a bit easier to use. Depending on how frequently this is used, it may or may not be necessary to add this helper. That said, I'm willing to accept it.

I would go for the top-level `pgconn.Config` field. That matches what we already have with `MinProtocolVersion` and `MaxProtocolVersion`.

carter-ya:
Thanks, that makes sense. I’ll keep it as a top-level `pgconn.Config` field and rename it to `MaxProtocolMessageBodyLen` to make the protocol-level meaning clearer while matching the existing `MinProtocolVersion` / `MaxProtocolVersion` style.

Issue #2560, opened 2026-05-16
pgxpool: MaxConnLifeTime not enforced at Acquire time, potentially allowing connections to live indefinitely on hot pools


arthurdotwork:
**Describe the bug**

`MaxConnLifetime` is documented as "duration since creation after which a connection will be automatically closed." The expiry check (`isExpired`) only runs inside `checkConnsHealth`, which only inspects idle connections via `AcquireAllIdle`. It is never checked during `Acquire`.

On a hot pool where connections are rarely idle when the health check tick fires, a connection can live well past `MaxConnLifetime` indefinitely since it's constantly acquired, used, returned, and immediately re-acquired before the next health check has a chance to see it idle.

This matters in practice when running behind a connection proxy with a finite grace period (e.g. PgBouncer sidecar in Kubernetes). If `MaxConnLifetime` is set below the proxy's `terminationGracePeriodSeconds` to ensure connections are recycled before `SIGKILL`, the guarantee doesn't hold on hot pools which are exactly the services where it matters most.

**To Reproduce**

- `MaxConnLifetime` = 90s, `HealthCheckPeriod` = 60s
- Pool is under sustained load (e.g. a high-QPS service)
- A connection is created at t=0
- Health check at t=60: conn is currently acquired → skipped
- Conn expires at t=90: nobody checks, it's in use
- Conn is returned at t=90.001, re-acquired at t=90.002 → ShouldPing sees IdleDuration < 1s → no ping
- Health check at t=120: conn is acquired again → skipped
- Connection is now 120s+ old and still in use

**Expected behavior**

We should check against the `maxAge` when acquiring a connection. If we use `isExpired`, it is a single time comparison operation with no I/O cost.

```
// in Acquire loop, after ShouldPing block:
if p.isExpired(res) {
    res.Destroy()
    continue
}
```

This would make `MaxConnLifetime` a hard guarantee regardless of pool utilization.

**Actual behavior**

`MaxConnLifetime` is not a hard guarantee against a hot pool.

**Version**

- pgx v5.8.0
- Go 1.24
- Kubernetes with PgBouncer sidecar (120s terminationGracePeriodSeconds)

**Additional context**

N/A.

arthurdotwork:
Kudos to @jbourdale for the investigation 🙏

arthurdotwork:
[...]

PR #2561, opened 2026-05-16
feat: check if conn is expired before acquire

arthurdotwork:
This should take care of #2560, all informations are in the issue.


jackc:
The tests didn't pass, but it was a small fix so I just fixed and merged it.

Issue #2522, opened 2026-03-20
Discovered DoS vulnerability in pgproto3/v2: panic on negative field length (GHSA-jqcq-xjh3-6g23)


aishubin:
A high-severity vulnerability has been discovered in the dependency github.com/jackc/pgproto3/v2.

Issue Summary:
The DataRow.Decode function improperly validates field lengths. A malicious or compromised PostgreSQL server can send a DataRow message with a negative field length, causing a slice bounds out of range panic.

Source:
https://github.com/advisories/GHSA-jqcq-xjh3-6g23

Required Action:
The dependency must be updated to a version that includes the fix for this vulnerability.

jackc:
That package is not used in v5 of pgx. And that library itself is EOL and archived. Also see: https://github.com/jackc/pgx/issues/2507.

Issue #2520, opened 2026-03-19
functionCall.Decode memory-safety vulnerability CWE-697


athuljayaram:
pgproto3/function_call.go

The argument length field is read as: argumentLength := int(binary.BigEndian.Uint32(src[rp:]))

Note the absence of the intermediate int32() cast present in every other length-field read in the package. On a 64-bit platform, int is 64 bits wide. int(uint32(0xFFFFFFFF)) evaluates to 4,294,967,295, not -1. The null sentinel check: if argumentLength == -1 { // treat as SQL NULL }

is dead code on every 64-bit machine. When a server sends a null argument using the standard 0xFFFFFFFF encoding, the null path is never taken. Execution falls through to: argumentValue := src[rp : rp+argumentLength]

Attempting to slice a small buffer with an upper bound of 4,294,967,295 causes an immediate runtime panic: panic: runtime error: slice bounds out of range [:4294967307] with capacity 14

The fix is restoring the missing cast: argumentLength := int(int32(binary.BigEndian.Uint32(src[rp:])))

This bug crashes not only on crafted input but on any legitimate PostgreSQL server that sends a null function argument, making it a correctness issue as well as a security one.

Reference

https://securityinfinity.com/research/pgx-v5-functioncall-decode-broken-null-sentinel-64bit

jackc:
Thanks for the report. Fixed now.

One clarification though, FunctionCall is a Frontend message. That is, it is a message sent by the client not the server. FunctionCall.Decode is unreachable when using pgx as a driver as pgx sends them but never receives them. Where it is reachable is when using the `pgproto3` package directly for purposes such as simulating a server or acting as a proxy.

Issue #2519, opened 2026-03-19
Security: Bind.Decode negative parameter length bypass (CWE-129)


athuljayaram:
## Summary

`Bind.Decode` in `pgproto3/bind.go` contains an incomplete bounds check that allows a server-controlled negative parameter length (other than the null sentinel `-1`) to bypass validation and cause an immediate client-side panic.

- **File:** `pgproto3/bind.go`
- **Affected version:** v5.8.0
- **CWE:** [CWE-129 — Improper Validation of Array Index](https://cwe.mitre.org/data/definitions/129.html)
- **CVSS 3.1:** 7.5 High — `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`

## Vulnerable Code

The parameter length is correctly read as a signed integer:

```go
msgSize := int(int32(binary.BigEndian.Uint32(src[rp:])))
```

The null sentinel check (`msgSize == -1`) is correct. However, the bounds check that follows is:

```go
if len(src[rp:]) < msgSize {
    return &invalidMessageFormatErr{...}
}
```

`len()` always returns a non-negative value. When `msgSize` is **any negative value other than `-1`**, the comparison is always `false` and the guard is bypassed entirely. Execution reaches:

```go
dst.Parameters[i] = src[rp : rp+msgSize]
```

A negative slice upper bound causes an immediate, unrecoverable runtime panic:

```
panic: runtime error: slice bounds out of range [:-13619142]
```

## Root Cause

The identical guard is already present in `DataRow.Decode` in the same package but was not applied to `Bind.Decode`:

```go
// DataRow.Decode — correct:
} else if valueLen < 0 {
    return &invalidMessageFormatErr{messageType: "DataRow"}
}
```

## Fix

Add a single additional condition to the existing bounds check:

```diff
- if len(src[rp:]) < msgSize {
+ if msgSize < 0 || len(src[rp:]) < msgSize {
      return &invalidMessageFormatErr{messageType: "Bind"}
  }
```

## Impact

A malicious or compromised PostgreSQL server can crash any pgx/v5 client process by sending a `Bind` message with a parameter length field set to any negative value other than `-1` (e.g. `0xFFFFEB7A` → `int32 = -5254`). No authentication is required — only the ability to route the client to an attacker-controlled endpoint.

## Reference

https://securityinfinity.com/research/pgx-v5-bind-decode-negative-parameter-length-bypass

jackc:
[...]

Issue #2507, opened 2026-02-19
Negative field length panics in DataRow.Decode


athuljayaram:
**Package:** `github.com/jackc/pgproto3` v2.3.3
  **File:** `data_row.go:48–59`
  **Severity:** 7.5 High
  **CWE:** CWE-129 - Improper Validation of Array Index

  ---

  ## Summary

  `DataRow.Decode` panics with `slice bounds out of range` when a server
  sends a field length that is a negative int32 value (any uint32 in the
  range `0x80000000`–`0xFFFFFFFE`).

  ---

  ## Root Cause

  Line 48 reads the field length correctly as a signed int32:

go
  msgSize := int(int32(binary.BigEndian.Uint32(src[rp:])))

  The null sentinel check on the following line correctly handles -1.
  However, the subsequent bounds check at line 55:

  if len(src[rp:]) < msgSize {
      return &invalidMessageFormatErr{messageType: "DataRow"}
  }

  is vacuously false whenever msgSize is negative. In Go, len() always
  returns a non-negative integer, so len(...) < -13619152 can never be
  true. The guard is bypassed entirely, and execution reaches line 59:

  dst.Values[i] = src[rp : rp+msgSize : rp+msgSize]
  // panic: runtime error: slice bounds out of range [:-13619152]

  The process terminates unconditionally.

  ---
##  Impact

  A malicious or compromised PostgreSQL server can crash any Go application
  using this library by sending a single crafted DataRow message. No
  special privilege is required — any server the client connects to can
  trigger this.

  CVSS 3.1: AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H - 7.5 High

  ---
 ##   Fix

  Add a msgSize < 0 guard between the null check and the bounds check:

  if msgSize == -1 {
      dst.Values[i] = nil
  } else if msgSize < 0 || len(src[rp:]) < msgSize {
      return &invalidMessageFormatErr{messageType: "DataRow"}
  } else {
      dst.Values[i] = src[rp : rp+msgSize : rp+msgSize]
      rp += msgSize
  }

## Reference
https://securityinfinity.com/research/memory-safety-vulnerabilities-in-go-postgresql-wire-protocol-parsers-pgproto3-pgx

jackc:
[...]

PR #2566, opened 2026-05-24
fix(pgtype): return error instead of panicking on malformed geometric text

ultramcu:
## Problem

The text scan plans for `circle`, `box`, `lseg`, `path` and `polygon` parse the value by calling `strings.IndexByte` to locate a `,` or `)` separator and then slicing the string — without checking for the `-1` "not found" result. Malformed text that is missing an expected separator therefore panics with `slice bounds out of range [:-1]` instead of returning an error.

This is reachable through the public `database/sql` `Scanner` API (e.g. `Circle.Scan`, or scanning a text-format column into one of these types), so malformed/corrupt input turns into a process-crashing panic rather than a normal scan error.

Repro (panics before this change):

```go
var c pgtype.Circle
_ = c.Scan("<(123456789>") // panic: slice bounds out of range [:-1]

var b pgtype.Box
_ = b.Scan("(1234567890")  // panic

var l pgtype.Lseg
_ = l.Scan("[(123456789")  // panic

var p pgtype.Path
_ = p.Scan("((1234567")    // panic

var g pgtype.Polygon
_ = g.Scan("((1234567")    // panic
```

`point.go` already guards this correctly (it uses `strings.Cut` and returns `"invalid format for point"`); these five types were missed. This is the same class of malformed-input panic recently fixed for other types (#2507, #2519, #2520, #2522).

## Fix

Guard every `strings.IndexByte` result with an `if end == -1 { return fmt.Errorf("invalid format for <type>") }` before slicing, matching the existing `point.go` behavior. Behavior for well-formed input is unchanged; this only turns a panic into a returned error (no public API change).

## Tests

Adds (in `pgtype/geometric_malformed_test.go`, no database required):

- `TestGeometricScanMalformedReturnsError` — each of the five types returns an error (no panic) on malformed input. Fails before this change (panics), passes after.
- `TestGeometricScanValid` — well-formed values still parse correctly.

`gofmt -l -s`, `go vet ./pgtype/`, and `go build ./...` are clean.

---

> Disclosure (per CONTRIBUTING.md): this is an AI-assisted contribution. The bug was surfaced by an automated audit; the fix and tests were written with AI assistance and reviewed and verified locally (fail-before / pass-after).

### DIFF
--- a/pgtype/box.go
+++ b/pgtype/box.go
@@ -189,6 +189,9 @@ func (scanPlanTextAnyToBoxScanner) Scan(src []byte, dst any) error {
 
 	var end int
 	end = strings.IndexByte(str, ',')
+	if end == -1 {
+		return fmt.Errorf("invalid format for Box")
+	}
 
 	x1, err := strconv.ParseFloat(str[:end], 64)
 	if err != nil {
@@ -197,6 +200,9 @@ func (scanPlanTextAnyToBoxScanner) Scan(src []byte, dst any) error {
 
 	str = str[end+1:]
 	end = strings.IndexByte(str, ')')
+	if end == -1 {
+		return fmt.Errorf("invalid format for Box")
+	}
 
 	y1, err := strconv.ParseFloat(str[:end], 64)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


jackc:
Merged with some additional fixes found during review.

Issue #2453, opened 2025-12-18
Panic encoding when non-nil interfaces with nil values occur


jaltavilla:
**Describe the bug**
We've encountered panics when encoding composite types and arrays due to nil checks not handling non-nil interfaces with nil values. Since the nil check returns a false negative later code goes on to call a value receiver method and panic.
Here are a few places where we've encountered them:
* https://github.com/jackc/pgx/blob/4c1308c14b2eabaaa365e8cd69120b1de8f953e9/pgtype/array_codec.go#L134
* https://github.com/jackc/pgx/blob/4c1308c14b2eabaaa365e8cd69120b1de8f953e9/pgtype/composite.go#L479
Replacing these checks with one that also uses reflection fixes these panics. It's likely other codecs have similar code paths.

**To Reproduce**
The below example shows cases that cause both the array and composite codecs to trigger a panic.

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/pgtype"
)

type Child struct {
	field1 string
	field2 int32
}

type Parent struct {
	field1 int32
	field2 *Child
}

type ParentWithSlice struct {
	field1 int32
	field2 []*Child
}

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(context.Background())

	_, err = conn.Exec(context.Background(),
		`DROP TYPE IF EXISTS parent_with_slice_t CASCADE;
		 DROP TYPE IF EXISTS parent_t CASCADE;
		 DROP TYPE IF EXISTS child_t CASCADE;
		 CREATE TYPE child_t AS ( field1 text, field2 int4 );
		 CREATE TYPE parent_t AS ( field1 int4, field2 child_t );
		 CREATE TYPE parent_with_slice_t AS ( field1 int4, field2 child_t[] );
		 CREATE OR REPLACE PROCEDURE do_something(parent parent_t)
		 LANGUAGE plpgsql
		 AS $$
		 BEGIN
		 END;
		 $$;
		 CREATE OR REPLACE PROCEDURE do_something_else(parent parent_with_slice_t)
		 LANGUAGE plpgsql
		 AS $$
		 BEGIN
		 END;
		 $$;`)
	if err != nil {
		log.Fatal(err)
	}

[...]

PR #2567, opened 2026-05-25
pgtype: handle typed-nil driver.Valuer in array/composite codecs

donners:
**Note:** This proposed fix was authored by Claude Code (Opus 4.7). I have a good understanding of the problem and the aim of the fix, but don't have a deep understanding of the surrounding code. Map.Encode already handles this issue, so the fix seemed fairly well scoped in terms of mirroring that pattern.

---

Regression introduced in v5.8.0 (1a5fa7f). The fix for #2453 routed typed-nil array elements and composite fields through the NULL path by discarding the second return value of isNilDriverValuer. For value-receiver Valuers on a typed-nil receiver the helper returns (true, true) meaning "nil but Value() is safe to call" — Map.Encode honours this; the array/composite encoders were dropping it on the floor, so a column declared NOT NULL JSONB[] that previously accepted []T{nil, nil} (because T.Value() returns []byte("[]")) now fails the insert because pgx writes {NULL,NULL}.

Mirrors the existing Map.Encode pattern at pgtype.go:1970: when isNil && callNilDriverValuer, dispatch through encodePlanDriverValuer instead of emitting NULL. The pointer-receiver-Valuer panic case from #2453 still goes through the NULL path because isNilDriverValuer returns (true, false) for it.

### DIFF
--- a/pgtype/array_codec.go
+++ b/pgtype/array_codec.go
@@ -131,7 +131,8 @@ func (p *encodePlanArrayCodecText) Encode(value any, buf []byte) (newBuf []byte,
 
 		elem := array.Index(i)
 		var elemBuf []byte
-		if isNil, _ := isNilDriverValuer(elem); !isNil {
+		isNil, callNilDriverValuer := isNilDriverValuer(elem)
+		if !isNil {
 			elemType := reflect.TypeOf(elem)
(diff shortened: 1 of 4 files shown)


jackc:
Thanks!

PR #2568, opened 2026-05-27
Return ErrTxClosed from Rollback when the connection is already closed (#2557)

### LINKED ISSUE #2557: Transaction rollback returns `conn closed` instead of `ErrTxDone` when the pgconn is closed
**Describe the bug**

I spend quite some time further debugging #2551, and think that I now found the actual issue I was running into.

When the context is canceled while a transaction query is executed, the low-level `pgconn.PgConn` is closed due to an `i/o timeout` at https://github.com/jackc/pgx/blob/82b212cb99c5e77bb54568a5ca7b9fa2b772dc0a/pgconn/pgconn.go#L1533-L1539
But the outer `pgx.dbTx` state is not set to "closed".

**To Reproduce**

```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"github.com/jackc/pgx/v5"
)

func main() {
	dsn := os.Getenv("TEST_DATABASE_POSTGRESQL")

	conn, err := pgx.Connect(context.Background(), dsn)
	if err != nil {
		fmt.Printf("unexpected error: %v\n", err)
		os.Exit(1)
	}

	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
	defer cancel()

	tx, err := conn.BeginTx(ctx, pgx.TxOptions{})
	if err != nil {
		fmt.Printf("unexpected error: %v\n", err)
		os.Exit(1)
	}

	// Sleep for longer than the context timeout to trigger a cancellation
	_, _ = tx.Exec(ctx, "SELECT pg_sleep(1)")
	err = tx.Rollback(context.Background())
	fmt.Printf("error on first rollback: %v\n", err)

	err = tx.Rollback(context.Background())
	fmt.Printf("error on second rollback: %v\n", err)
}

// outputs:
//   error on first rollback: conn closed
//   error on second rollback: tx is closed
```

**Expected behavior**

The transaction state should match the underlying connection state, or rather the error returned from `tx.Rollback` should be `pgx.ErrTxClosed` when the connection is closed.
It generally seems weird to me that the low-level connection would close itself without notifying the higher-level code that created the connection.

**Actual behavior**

An unexpected, non-assertable, low-level error is returned.

**Version**
 - Go: `$ go version` -> go version go1.26.2 darwin/arm64
- PostgreSQL: `$ psql --no-psqlrc --tuples-only -c 'select version()'` -> PostgreSQL 16.13 (Debian 16.13-1.pgdg13+1) on aarch64-unknown-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
- pgx: `$ grep 'github.com/jackc/pgx/v[0-9]' go.mod` -> v5.9.2

**Additional context**

Potentially related: #1145 #2100

luongs3:
Fixes #2557.

### Problem

When a transaction query's context is canceled mid-flight, the low-level `pgconn.PgConn` is torn down (i/o timeout), but the outer `pgx.dbTx` state is left open (`tx.closed` stays `false`). The next `Rollback` then issues `"rollback"` on the already-dead connection and surfaces the raw `"conn closed"` error instead of `pgx.ErrTxClosed`.

This contradicts `Rollback`'s documented contract:

> Rollback will return an error where `errors.Is(ErrTxClosed)` is true if the Tx is already closed […]

A closed underlying connection means the transaction is, for all practical purposes, already closed — so callers reasonably expect `ErrTxClosed` (which they can assert on) rather than an opaque low-level error.

### Reproduction (from the issue)

```go
ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
defer cancel()

tx, _ := conn.BeginTx(ctx, pgx.TxOptions{})
_, _ = tx.Exec(ctx, "SELECT pg_sleep(1)") // canceled; pgconn is closed
err := tx.Rollback(context.Background())
// before: err == "conn closed"
// after:  errors.Is(err, pgx.ErrTxClosed) == true
```

### Fix

In `Rollback`, check `tx.conn.IsClosed()` up front. If the connection is already closed, mark the tx closed and return `ErrTxClosed`. This mirrors how `Commit` already consults connection state (`tx.conn.PgConn().TxStatus()`), and leaves the normal rollback path — including the existing "rollback failed → `die()`" handling — untouched.

### Tests

Added `TestTransactionRollbackAfterCanceledQueryReturnsErrTxClosed`, which reproduces the issue's exact scenario (cancel a `pg_sleep` query, then roll back) and asserts `errors.Is(err, pgx.ErrTxClosed)`. It fails on `master` (`conn closed`) and passes with this change. The existing transaction test suite continues to pass.

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -211,6 +211,16 @@ func (tx *dbTx) Rollback(ctx context.Context) error {
 		return ErrTxClosed
 	}
 
+	// The underlying connection may have already been closed out from under the
+	// transaction (e.g. a query's context was canceled, triggering an i/o
+	// timeout that tore down the pgconn). In that case the transaction is
+	// effectively closed, so honor the documented contract and return
+	// ErrTxClosed rather than leaking the low-level "conn closed" error.
+	if tx.conn.IsClosed() {
+		tx.closed = true
+		return ErrTxClosed
+	}
+
 	_, err := tx.conn.Exec(ctx, "rollback")
 	tx.closed = true
 	if err != nil {
--- a/tx_rollback_closed_test.go
+++ b/tx_rollback_closed_test.go
@@ -1,0 +1,40 @@
+package pgx_test
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/stretchr/testify/require"
+)
+
+// When a transaction query's context is canceled mid-flight, the low-level
+// pgconn is torn down but the outer pgx.Tx state is left open. The next
+// Rollback should honor its documented contract and return an error matching
+// ErrTxClosed, rather than leaking the low-level "conn closed" error.
+//
+// Regression test for jackc/pgx#2557.
+func TestTransactionRollbackAfterCanceledQueryReturnsErrTxClosed(t *testing.T) {
+	t.Parallel()
+
+	conn := mustConnectString(t, os.Getenv("PGX_TEST_DATABASE"))
+	defer closeConn(t, conn)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	tx, err := conn.BeginTx(ctx, pgx.TxOptions{})
+	require.NoError(t, err)
+
+	// Sleep longer than the context timeout so the query is canceled and the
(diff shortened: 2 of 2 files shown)


jackc:
As discussed in the issue, if any change is to be made, this isn't the direction we will go.

PR #2570, opened 2026-05-28
fix(pgtype): JSON DecodeValue returns raw bytes for literal 'null' (#2430)

### LINKED ISSUE #2430: `pgx.Conn.Query()` behavior for JSON string value of "null" returns incorrect value
This query incorrectly returns a nil value when using the `pgx.Conn.Query()` method:

```sql
select 'null'::json;
```

This response is the JSON string literal value "null", not the JSON null literal. The JSON codec is incorrectly parsing this response into a nil value.

Here's a simple repro:

```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/jackc/pgx/v5"
)

func main() {
	// Adjust connection string as needed
	connStr := "postgres://postgres:postgres@localhost:5432/postgres"

	conn, err := pgx.Connect(context.Background(), connStr)
	if err != nil {
		log.Fatal("connect error:", err)
	}

	rows, err := conn.Query(context.Background(), `select 'null'::json;`)
	if err != nil {
		log.Fatal("query error:", err)
	}

	var rowSlice []any
	for rows.Next() {
		row, err := rows.Values()
		if err != nil {
			log.Fatal("query error:", err)
		}
		rowSlice = append(rowSlice, row)
	}

	fmt.Println("Result:", rowSlice)
}
```

Prints:

```
Result: [[<nil>]]
```

This works correctly using the QueryRow() and Scan() interfaces directly with a string scan target, like this:

```go
	var result string

	err = conn.QueryRow(context.Background(), `select 'null'::json;`).Scan(&result)
```

Also works correctly with the vanilla postgres connector:

```go
package main

import (
	"database/sql"
	"fmt"
	"log"

	_ "github.com/lib/pq"
)

func main() {
	connStr := "host=localhost port=5432 user=postgres password=postgres dbname=postgres sslmode=disable"

	db, err := sql.Open("postgres", connStr)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	var result string

	err = db.QueryRow(`select 'null'::json;`).Scan(&result)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Result:", result)
}
```

Prints:

```
Result: null
```

**Version**
 - Go: go version go1.25.3 darwin/arm64
 - PostgreSQL: PostgreSQL 18.0 (Homebrew) on aarch64-apple-darwin24.4.0, compiled by Apple clang version 17.0.0 (clang-1700.0.13.3), 64-bit
 - pgx: v5.6.1-0.20240826124046-97d20ccfadaa

luongs3:
Fixes #2430.

## Problem

`JSONCodec.DecodeValue` and `JSONBCodec.DecodeValue` unmarshal into a Go `interface{}`, which collapses the JSON document `null` to a Go `nil`. That makes a JSON `null` literal **indistinguishable** from SQL NULL — both come out of the `rows.Values()` / `Map.Scan(&any)` pipeline as `nil`.

The vanilla `database/sql` interface does not have this problem, because `DecodeDatabaseSQLValue` returns the raw response bytes — so `QueryRow().Scan(&str)` correctly yields `"null"` for `select 'null'::json;`.

## Repro

```go
rows, _ := conn.Query(ctx, `select 'null'::json;`)
for rows.Next() {
    row, _ := rows.Values()
    fmt.Println(row)
}
// Before: [<nil>]   (looks identical to SQL NULL)
// After:  [[110 117 108 108]]   ([]byte("null"))
```

## Fix

When `src` is non-nil but `c.Unmarshal(src, &dst)` yields `dst == nil` (i.e. the JSON document was literally `null`, or `null` with whitespace), return a copy of the raw `src` bytes instead of the Go nil. SQL NULL (`src == nil`) continues to return Go nil as before.

Applied to:

- `pgtype/json.go` — `JSONCodec.DecodeValue`
- `pgtype/jsonb.go` — `JSONBCodec.DecodeValue` (same change, applied after the binary-format version byte is stripped)

## What does NOT change

Other JSON values (`"hello"`, `42`, `[1,2,3]`, `{"k":"v"}`) continue to unmarshal into Go values exactly as before. The only behavior change is the JSON-null-but-not-SQL-NULL case, and only to make it distinguishable from SQL NULL.

## Tests

`pgtype/json_2430_test.go` adds `TestJSONCodecDecodeValueJSONNullLiteral` — 7 table-driven sub-cases exercising the codec directly (no DB required):

- `json/sql_null` — SQL NULL still returns Go nil
- `jsonb/sql_null/text` — same for JSONB
- `json/json_null_literal` — JSON `null` now returns `[]byte("null")`
- `jsonb/json_null_literal/text` — JSONB `null` (text format) returns `[]byte("null")`
- `jsonb/json_null_literal/binary` — JSONB `null` (binary format) returns `[]byte("null")` after stripping the v1 header byte
- `json/string` — `"hello"` still unmarshals to Go string `"hello"`
- `jsonb/number/text` — `42` still unmarshals to Go `float64(42)`

All pass with the fix; the existing JSON tests that require a live PostgreSQL (`TestJSONCodecScanNull`, etc.) are unchanged.

### DIFF
--- a/pgtype/json.go
+++ b/pgtype/json.go
@@ -239,5 +239,20 @@ func (c *JSONCodec) DecodeValue(m *Map, oid uint32, format int16, src []byte) (a
 
 	var dst any
 	err := c.Unmarshal(src, &dst)
-	return dst, err
+	if err != nil {
+		return nil, err
+	}
+	if dst == nil {
+		// The source was non-NULL (src != nil) but unmarshaled to a Go nil — this
+		// happens for the literal JSON document `null` (and equivalent whitespace
+		// padding). Returning nil here would be indistinguishable from SQL NULL,
+		// which is what DecodeValue returns when src is nil above. Hand back a
+		// copy of the raw bytes so callers (rows.Values(), Map.Scan(&any), …) can
+		// tell the two apart — matching what DecodeDatabaseSQLValue returns to
+		// the database/sql interface for the same input. See #2430.
+		dstBuf := make([]byte, len(src))
+		copy(dstBuf, src)
+		return dstBuf, nil
+	}
+	return dst, nil
 }
--- a/pgtype/json_2430_test.go
+++ b/pgtype/json_2430_test.go
@@ -1,0 +1,84 @@
+package pgtype_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgtype"
+)
+
+// TestJSONCodecDecodeValueJSONNullLiteral guards against regression of #2430:
+// the JSON document `null` is a valid non-NULL value distinct from SQL NULL.
+// JSONCodec.DecodeValue and JSONBCodec.DecodeValue must surface that distinction
+// to callers like rows.Values() (which uses DecodeValue), instead of collapsing
+// both to Go nil.
+func TestJSONCodecDecodeValueJSONNullLiteral(t *testing.T) {
+	t.Parallel()
+
+	jsonCodec := &pgtype.JSONCodec{Marshal: json.Marshal, Unmarshal: json.Unmarshal}
+	jsonbCodec := &pgtype.JSONBCodec{Marshal: json.Marshal, Unmarshal: json.Unmarshal}
+
+	cases := []struct {
+		name   string
+		decode func([]byte) (any, error)
+		src    []byte
+		want   any
+	}{
+		// SQL NULL — src nil — must continue to return Go nil.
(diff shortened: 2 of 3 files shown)


jackc:
As I mentioned in the original issue, I'm not convinced that the current behavior is wrong. But even if it is, having JSON `null` be special cased to `[]byte` doesn't seem right either.

luongs3:
Makes sense — you know the type semantics here better than I do, and I agree special-casing `null` to `[]byte` is an odd shape even if the asymmetry is real. I'll close this. Thanks for taking a look.

Issue #998, opened 2021-05-03
Feature Request: Remove Truncation from Args when logging


marksalpeter:
In `func logQueryArgs(args []interface{}` it looks like pgx deliberately truncates long args. We need to be able to use the full args to debug our queries in staging. 

Can we remove this truncation completely  or add a config param that would allow full args to be logger?

I'm happy to make the change in a pr. Just let me know which you prefer.

jackc:
I don't think we should remove it entirely. That would flood the logs of people relying on that behavior now. Making the truncation length configurable would be reasonable though.

PR #2588, opened 2026-06-22
fix: tracelog truncation makes output longer than original

### LINKED ISSUE #2095: Tracelog truncation can make output longer.
This is a pretty small nitpick, but I noticed:

https://github.com/jackc/pgx/blob/ea9610f6722e0c6c6073660f14ded3c9683a593e/tracelog/tracelog.go#L101

The message indicating truncation can take up more space than the number of bytes being removed - i.e. if you commonly have values of 65 bytes, it will truncate 1 byte but print out an extra ` (truncated 1 bytes)`, which is 20 more bytes. Maybe changing this can be rolled into a solution for https://github.com/jackc/pgx/issues/998

xbrxr03:
## Problem

`logQueryArgs` truncates long strings and byte slices, but the truncation suffix makes the output **longer** than the original value:

- A 65-byte string truncated to 64 bytes + `" (truncated 1 bytes)"` = **84 bytes** — 29 bytes *longer* than the original
- A 65-byte slice shows 128 hex chars + `" (truncated 1 bytes)"` = **150+ chars** — much longer than the 130-char full hex encoding

Reported in #2095.

## Fix

Replace the verbose `fmt.Sprintf` suffix with an ellipsis (`…`) and ensure the truncated result is always **strictly shorter** than the original:

- **`[]byte`**: Show first 60 bytes as hex (120 chars) + `…` = 123 chars, always shorter than the 130+ char full hex
- **`string`**: Truncate at a rune boundary so that `len(result) < len(original)`, using `maxCut = len(v) - 4` to account for the 3-byte ellipsis

Also fixed the off-by-one: `[]byte` threshold was `< 64` (excluding 64-byte slices) but `string` was `> 64` (excluding 65+ byte strings). Now both use `<= 64` / `> 64` consistently.

## Tests

Added `TestLogQueryArgsTruncation` with 5 subtests:
- Short strings and bytes pass through unchanged
- Long strings truncate with `…` and are strictly shorter than the original
- Long bytes truncate with `…` and are shorter than full hex
- UTF-8 strings truncate at rune boundaries

Fixes #2095

### DIFF
--- a/tracelog/tracelog.go
+++ b/tracelog/tracelog.go
@@ -97,19 +97,41 @@ func logQueryArgs(args []any) []any {
 	for _, a := range args {
 		switch v := a.(type) {
 		case []byte:
-			if len(v) < 64 {
+			if len(v) <= 64 {
 				a = hex.EncodeToString(v)
 			} else {
-				a = fmt.Sprintf("%x (truncated %d bytes)", v[:64], len(v)-64)
+				// Show first 60 bytes as hex (120 chars) + ellipsis.
+				// This ensures the result is always shorter than the
+				// full hex encoding: 120 + 3 < 130+ for any 65+ byte input.
+				a = hex.EncodeToString(v[:60]) + "…"
 			}
 		case string:
 			if len(v) > 64 {
+				// Find the rune boundary nearest to 64 bytes,
+				// then back off so the ellipsis (3 bytes) still
+				// makes the result shorter than the original.
 				l := 0
 				for w := 0; l < 64; l += w {
 					_, w = utf8.DecodeRuneInString(v[l:])
 				}
 				if len(v) > l {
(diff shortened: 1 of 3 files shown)


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

jackc:
> Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

🤷‍♂️ 

It's a non-commercial project. I maintain it on nights and weekends. So if it's not an emergency it may take a few weeks for me to take a look.

As far as this particular issue though, if anything is to be done, I would see it as two changes.

1. Add a field to `TraceLogConfig` that controls the maximum length of the value before truncation. This would also resolve #998.
2. Retain the number of bytes truncated notification. Knowing a query had 200 byte argument vs. a 200 MB argument is important information.

Issue #2583, opened 2026-06-15
Allow BeginTx to treat server-sent BEGIN errors as non-fatal instead of killing the connection


VashDD:
**Is your feature request related to a problem? Please describe.**

We are introducing load shedding to our Postgres connection pooler and upon testing we have realized that `Conn.BeginTx` unconditionally calls `c.die()` on any error from the BEGIN statement ([here](https://github.com/jackc/pgx/blob/7293fb11125be0373a92f716683f2d494f6fd4b0/tx.go#L105)), even if the error is a clean server‑side rejection (an `ErrorResponse` + `ReadyForQuery`, connection still usable), which causes it so under stress, connections on the pool are constantly destroyed and re-created, which puts even a bigger load on our connection pooler.

**Describe the solution you'd like**

Ideally, we should be able to define on the pgx config which error codes we consider non-fatal and not kill the connection upon receiving said errors. This way we don't churn connections to the connection pooler. Also possibly have a config to not kill the connection if the error code is not a fatal.

**Describe alternatives you've considered**

As of now, we have implemented our own custom `Tx` struct which basically checks the error code and avoids closing the connections at the begin if it is an insufficient resources error code (53000), which is the one we send if the connection pooler is rejecting queries due to load shedding.

**Additional context**

I believe other Postgres proxies like [Multigres](https://multigres.com/docs/architecture#resilience:~:text=Multigres%20will%20provide%20resilience%20against%20spikes%20and%20overloads%20by%20employing%20queuing%20and%20load%20shedding%20mechanisms.) and [PGKeeper](https://www.figma.com/blog/pgkeeper-building-the-bouncer-we-needed-for-postgres/#:~:text=Both%20support%20priority%20levels%2C%20so%20higher%2Dpriority%20requests%20are%20served%20first%20when%20capacity%20runs%20low%2C%20and%20both%20use%20CoDel%20to%20drop%20stale%20requests%20under%20sustained%20overload) implement or plan to implement load shedding capabilities, so this functionality wouldn't be scoped only to our case.

jackc:
Hmm... if I understand correctly, this is a case that can't occur with pgx directly talking to PostgreSQL. You would never get an error starting a transaction unless something was catastrophically wrong (e.g. application or pgx bug, protocol desync, or network error).

I am willing to relax this requirement, maybe allow any non-fatal PgError and die for everything else. That should still be safe.

VashDD:
[...]

PR #2589, opened 2026-06-23
Keep connections alive after BEGIN statement for non-FATAL errors

VashDD:
## Problem
For more context, see: #2583 

`Conn.BeginTx` calls `c.die()` on **any** error from `BEGIN`, unconditionally destroying the connection:

```go
_, err := c.Exec(ctx, txOptions.beginSQL())
if err != nil {
    c.die()
    return nil, err
}
```

The assumption ("begin should never fail unless the connection is broken") is true for connecting directly to postgres, but with a connection pooler in between, there are errors that should not kill the connection, as it is reusable, for instance errors due to load shedding.

## Change

Only `die()` when the connection is actually compromised due to:

- non-`PgError` (transport error / context cancellation) 
- `FATAL`/`PANIC` `PgError` (server tearing the session down) 

## Tests

- `TestBeginTxNonFatalErrorKeepsConnAlive` — a BEGIN failing with a non-FATAL error leaves `IsClosed() == false` and the connection still answers `select 1`.
- `TestBeginTxFatalErrorKillsConn` — a BEGIN that terminates its own backend (`pg_terminate_backend(pg_backend_pid())`) returns a `FATAL` error and the connection is closed.
- `pgxpool.TestConnReleaseWhenBeginFail` — updated: a recoverable BEGIN failure now releases the connection back to the pool **alive** (TotalConns stays 1, reusable) instead of destroying it.
- `pgxpool.TestConnDestroyedWhenBeginFailsFatally` — a BEGIN that terminates its own backend fails fatally, so the pool destroys the connection (TotalConns drops to 0) instead of reusing it.

### DIFF
--- a/pgxpool/pool_test.go
+++ b/pgxpool/pool_test.go
@@ -1060,11 +1060,8 @@ func TestConnReleaseWhenBeginFail(t *testing.T) {
 	tx, err := db.BeginTx(ctx, pgx.TxOptions{
 		IsoLevel: pgx.TxIsoLevel("foo"),
 	})
-	assert.Error(t, err)
-	if !assert.Zero(t, tx) {
-		err := tx.Rollback(ctx)
-		assert.NoError(t, err)
-	}
+	require.Error(t, err)
+	require.Zero(t, tx)
 
 	require.EqualValues(t, 1, db.Stat().TotalConns())
 
--- a/pgxpool/pool_test.go
+++ b/pgxpool/pool_test.go
@@ -1075,6 +1075,40 @@ func TestConnReleaseWhenBeginFail(t *testing.T) {
 	require.EqualValues(t, 1, n)
 }
 
+func TestConnDestroyedWhenBeginFailsFatally(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+
+	controllerConn, err := pgx.Connect(ctx, os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+	defer controllerConn.Close(ctx)
+	pgxtest.SkipCockroachDB(t, controllerConn, "Server does not support pg_terminate_backend() (https://github.com/cockroachdb/cockroach/issues/35897)")
+
+	db, err := pgxpool.New(ctx, os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+	defer db.Close()
+
+	// The BEGIN terminates its own backend, so it fails with a fatal error and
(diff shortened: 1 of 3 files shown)


pgxpool/pool_test.go
```diff
@@ -1066,6 +1066,35 @@ func TestConnReleaseWhenBeginFail(t *testing.T) {
 		assert.NoError(t, err)
 	}
 
+	assert.EqualValues(t, 1, db.Stat().TotalConns())
```
sean-: Why isn't this a `requre`?

pgxpool/pool_test.go
```diff
@@ -1066,6 +1066,35 @@ func TestConnReleaseWhenBeginFail(t *testing.T) {
...
+	require.NoError(t, err)
+	defer controllerConn.Close(ctx)
+	pgxtest.SkipCockroachDB(t, controllerConn, "Server does not support pg_terminate_backend() (https://github.com/cockroachdb/cockroach/issues/35897)")
+
+	db, err := pgxpool.New(ctx, os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+	defer db.Close()
+
+	tx, err := db.BeginTx(ctx, pgx.TxOptions{BeginQuery: "select pg_terminate_backend(pg_backend_pid())"})
+	assert.Error(t, err)
```
sean-: Why isn't this a `requre`?

pgxpool/pool_test.go
```diff
@@ -1066,6 +1066,35 @@ func TestConnReleaseWhenBeginFail(t *testing.T) {
...
+	if !assert.Zero(t, tx) {
+		err := tx.Rollback(ctx)
+		assert.NoError(t, err)
```
sean-: `require`?

tx.go
```diff
@@ -112,6 +115,14 @@ func (c *Conn) BeginTx(ctx context.Context, txOptions TxOptions) (Tx, error) {
 	}, nil
 }
 
+func isConnectionFatal(pgErr *pgconn.PgError) bool {
+	severity := pgErr.SeverityUnlocalized
+	if severity == "" {
+		severity = pgErr.Severity
+	}
+	return severity == "FATAL" || severity == "PANIC"
```
sean-, line 123: Probably unrelated to this PR, but I wish these magic strings were constants that could be traced through the source.

VashDD:
Thanks for the review, addressed comments, ready for another pass.

tx.go
VashDD: Agree, I don't mind doing a follow up for that, but should take part in a different PR.

sean- (approved):

PR #2602, opened 2026-07-17
Fix nil pointer panic in Numeric.Int64Value when Int is nil

vsaraikin:
Numeric.Int64Value panics with a nil pointer dereference when the Numeric is valid but its Int field is nil.

The other methods that read a Numeric already treat a nil Int as zero. Float64Value returns 0, Value returns "0", and MarshalJSON returns 0. Int64Value is the exception. It goes through toBigInt, which returns the nil Int unchanged when Exp is 0, and then IsInt64 dereferences it.

pgtype.Numeric{Valid: true} is an easy way to end up here, since that's a fairly normal way to build a zero value.

```go
n := pgtype.Numeric{Valid: true}
n.Float64Value() // 0, ok
n.MarshalJSON()  // "0", ok
n.Int64Value()   // panic: nil pointer dereference
```

I put the guard in toBigInt so it returns zero when Int is nil, which makes all four methods agree. That also covers uint64Wrapper and uintWrapper ScanNumeric, which reach toBigInt the same way and would panic on a valid-but-nil Numeric too.

The test sits next to the existing Float64Value one in numeric_test.go. It panics on master and passes with the change. No database needed.

### DIFF
--- a/pgtype/numeric.go
+++ b/pgtype/numeric.go
@@ -149,6 +149,10 @@ func (n *Numeric) ScanScientific(src string) error {
 }
 
 func (n *Numeric) toBigInt() (*big.Int, error) {
+	if n.Int == nil {
+		return big.NewInt(0), nil
+	}
+
 	if n.Exp == 0 {
 		return n.Int, nil
 	}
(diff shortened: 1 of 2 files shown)


jackc:
Merged, with an additional fix for `NaN` and `Inf`.

PR #2603, opened 2026-07-19
Update PgBouncer guidance and expand integration coverage

achanda:
Document prepared statement support in PgBouncer 1.21+ and test all compatible query execution modes, including parameterized queries, transactions, Exec, and batches. Also enable pgbouncer tests in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,33 @@ jobs:
           PGX_SSL_PASSWORD: ${{ matrix.pgx-ssl-password }}
           PGX_TEST_TLS_CLIENT_CONN_STRING: ${{ matrix.pgx-test-tls-client-conn-string }}
 
+  test-pgbouncer:
+    name: Test PgBouncer with PostgreSQL 18
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v5
+
+      - name: Set up Go 1.26
+        uses: actions/setup-go@v6
+        with:
+          go-version: "1.26"
+
+      - name: Set up PostgreSQL 18
+        run: ci/setup_test.bash
+        env:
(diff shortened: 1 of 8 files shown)


jackc:
There are a few suggestions from Claude that made sense to me:

---

1. The new CI job runs the entire suite for two tests. ci.yml:157 is go test -parallel=1 -race -count=1 -v ./..., which duplicates the existing PG18/Go1.26 matrix job in full (~2 min of pgconn + ~36 s of pgproto3 alone here) and -v dumps the whole suite's output. Only TestPgbouncerQueryExecModes and TestConnContextCanceledCancelsRunningQueryOnServer consume PGX_TEST_PGBOUNCER_CONN_STRING. Narrowing to -run 'TestPgbouncer|TestConnContextCanceledCancelsRunningQueryOnServer' ./... and dropping -v cuts the job to well under a minute.

2. The 39-line source build is unnecessary. PGDG publishes pgbouncer 1.25.2-1.pgdg22.04+1 for jammy — I confirmed it's in the jammy-pgdg package index — and ci/setup_test.bash has already added that repo by the time setup_pgbouncer.bash runs. So the download, sha256 pin, build-essential/libevent-dev/libssl-dev install, and ./configure && make collapse to sudo apt-get install -y pgbouncer, and there's no pinned hash to bump on every PgBouncer release.

5. Coverage gap the PR could cheaply close. With StatementCacheCapacity: 32 and only ~5 distinct SQL strings, the cache never evicts, so pgx's protocol-level Close path (conn.go:1457) is never exercised through PgBouncer — arguably the most likely thing to break now that the docs recommend the default mode with PgBouncer. I tested it separately (capacity 8, 200 distinct statements) and it passes, so this is a coverage suggestion, not a defect.

achanda:
Thanks for the review, makes sense to me. Pushed an update with all suggested fixes.

Issue #2609, opened 2026-07-22
RowScanner takes precedence over CompositeIndexScanner incorrectly


jaltavilla:
**Describe the bug**
When scanning into a single go type the RowScanner interface will be used in some cases where it shouldn't be. For example if the one column contains a composite the CompositeIndexScanner should be used, but the RowScanner will instead be used. This will result in an error because there aren't multiple columns to scan into individual fields.

**To Reproduce**

If possible, please provide runnable example such as:

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/pgtype"
)

type Result struct {
	ID   int64
	Data string
}

func main() {
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close(context.Background())

	_, err = conn.Exec(context.Background(),
		`DROP FUNCTION IF EXISTS get_result;
		 DROP TYPE IF EXISTS result_t;
		 CREATE TYPE result_t AS (m_id INTEGER, m_data TEXT);
		 CREATE OR REPLACE FUNCTION get_result(v_input result_t)
		 	RETURNS result_t
		 AS 
		 $$
		 BEGIN
		 	RETURN v_input;
		 END;
		 $$ LANGUAGE plpgsql;`)
	if err != nil {
		log.Fatal(err)
	}

	types, err := conn.LoadTypes(context.Background(), []string{"result_t"})
	if err != nil {
		log.Fatal(err)
	}
	conn.TypeMap().RegisterTypes(types)

	input := Result{
		ID:   1,
		Data: "data",
	}

	var output Result
	row := conn.QueryRow(context.Background(), "SELECT get_result FROM get_result( $1 )", input)
	err = row.Scan(&output)
	if err != nil {
		log.Fatal(err)
	}
}

// Comment this out and it works
func (dest *Result) ScanRow(rows pgx.Rows) error {
	return rows.Scan(
		&dest.ID,
		&dest.Data,
	)
}

var _ pgx.RowScanner = new(Result)

func (value *Result) SkipUnderlyingTypePlan() {}

var _ pgtype.SkipUnderlyingTypePlanner = new(Result)

func (value *Result) ScanNull() error {
	return fmt.Errorf("cannot scan NULL into %s", "Result")
}

func (value *Result) ScanIndex(i int) any {
	switch i {
[...]

PR #2610, opened 2026-07-22
Fix Scan's handling of type that implements RowScanner and CompositeIndexGetter

jaltavilla:
This fixes issue #2609

### DIFF
--- a/rows.go
+++ b/rows.go
@@ -243,19 +243,18 @@ func (rows *baseRows) Scan(dest ...any) error {
 		return err
 	}
 
-	if len(dest) == 1 {
-		if rc, ok := dest[0].(RowScanner); ok {
-			err := rc.ScanRow(rows)
-			if err != nil {
-				rows.fatal(err)
-			}
+	if len(fieldDescriptions) != len(dest) {
+		rc, hasRowScanner := dest[0].(RowScanner)
+		if !hasRowScanner || len(dest) != 1 {
+			err := fmt.Errorf("number of field descriptions must equal number of destinations, got %d and %d", len(fieldDescriptions), len(dest))
+			rows.fatal(err)
 			return err
 		}
-	}
 
-	if len(fieldDescriptions) != len(dest) {
-		err := fmt.Errorf("number of field descriptions must equal number of destinations, got %d and %d", len(fieldDescriptions), len(dest))
-		rows.fatal(err)
+		err := rc.ScanRow(rows)
+		if err != nil {
+			rows.fatal(err)
+		}
 		return err
 	}
 
--- a/rows.go
+++ b/rows.go
@@ -232,6 +232,23 @@ func (rows *baseRows) Next() bool {
 	}
 }
 
+func isRegisteredAsComposite(typeMap *pgtype.Map, oid uint32) bool {
+	// Types registered with an appropriate codec can be inferred to be a composite.
+	// In the case we don't know it's a composite then we don't scan it as one.
+	// That's consistent with the requirement that composites need to be registered in the type map.
+	dt, ok := typeMap.TypeForOID(oid)
+	if !ok {
+		return false
+	}
+
+	switch dt.Codec.(type) {
+	case pgtype.RecordCodec, *pgtype.CompositeCodec:
+		return true
+	default:
+		return false
+	}
+}
+
 func (rows *baseRows) Scan(dest ...any) error {
 	m := rows.typeMap
 	fieldDescriptions := rows.FieldDescriptions()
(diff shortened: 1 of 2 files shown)


jackc:
According to Claude:

1. RowToMap breaks on any single-column query. [rows.go:542](vscode-webview://0efsq4ucopn6c9hovmg9t77ij4bth6na85bfaeq4esqtp9r4jou4/rows.go#L542) scans through one dest that implements RowScanner (*mapRowScanner). With one column the counts now match, so the RowScanner branch is never reached and pgx tries to plan a scan of the column into the map type:


can't scan into dest[0] (col: a): cannot scan text (OID 25) in text format into *pgx.mapRowScanner
So pgx.CollectRows(rows, pgx.RowToMap) on select 'foo'::text as a fails. Passes on master, fails on the PR branch. Two columns still work — which is why the existing suite doesn't catch it.

2. Panic when Scan is called with zero destinations. [rows.go:247](vscode-webview://0efsq4ucopn6c9hovmg9t77ij4bth6na85bfaeq4esqtp9r4jou4/rows.go#L247) evaluates dest[0] before checking len(dest):


panic: runtime error: index out of range [0] with length 0
  github.com/jackc/pgx/v5.(*baseRows).Scan  rows.go:247
rows.Scan() against a 1-column result returns a clean "number of field descriptions must equal number of destinations" error on master.

3. Every single-column RowScanner stops working — the general form of #1, and a public API behavior break. A user type whose ScanRow does rows.Scan(&rs.field) no longer has ScanRow called for a one-column result; it gets cannot scan text (OID 25) in text format into *T instead. That's broader than #2609 asks for, and there's no opt-out.

jaltavilla:
I've re-implemented this to only override the use of RowScanner in the specific case that the destination is a composite and the field is known to be a composite. 

I also added a couple tests for cases the prior pr broke to help future contributors.

jackc:
I spent some more time reviewing this and there are still some edge cases.

1. Multi-column results whose first column is a composite
2. No format check — text-format modes hard-fail with no fallback

Both of those can be fixed, and I would have just added a patch on top of yours, but fundamentally there is a conflict when both `RowScanner` and `CompositeIndexScanner` are implemented on the same type with different bahavior. In particular, `RowScanner` can perform arbitrary logic. e.g.

```go
func (r *auditedResult) ScanRow(rows pgx.Rows) error {
    var inner svInner              // svInner implements CompositeIndexScanner
    if err := rows.Scan(&inner); err != nil {
        return err
    }
    r.name, r.age = inner.name, inner.age
    r.scanRowRan = true                        // derived state
    r.displayName = strings.ToUpper(r.name)    // only ScanRow produces this
    return nil
}
```

Now frankly, I never thought that people would implement both interfaces on the same type. I would use renamed type for one of the interfaces. But obviously, at least one person is. And while I also can't imagine why you would use different logic in `RowScanner` vs `CompositeIndexScanner`, presumably someone might.

So I'm thinking maybe there should be a different approach. What if we always use `RowScanner` if it is available? This is actually the current logic.

The implementation of `RowScanner` is responsible for checking the number of columns and it can dispatch to the proper scanning strategy.

For example:

```go
type Person struct {
	Name string
	Age  int32
}

func (p *Person) ScanRow(rows pgx.Rows) error {
	if fds := rows.FieldDescriptions(); len(fds) == 1 {
		return rows.TypeMap().Scan(fds[0].DataTypeOID, fds[0].Format, rows.RawValues()[0], p)
	}
	return rows.Scan(&p.Name, &p.Age)
}

func (p *Person) ScanNull() error { return fmt.Errorf("cannot scan NULL into Person") }

func (p *Person) ScanIndex(i int) any {
	switch i {
	case 0:
		return &p.Name
	case 1:
		return &p.Age
	}
	panic("out of range")
}
```

I had to make a tiny change to make this convenient -- I added `Rows.TypeMap()` to make this approach cleaner -- previously mostly possible through `Rows.Conn().TypeMap()`. I already pushed that to master, makes sense regardless of this issue.

---

Unless you see something wrong with this approach, I would close this PR and the underlying issue. It should be a `ScanRow` implementation responsibility instead of a pgx responsibility.

jaltavilla:
Thanks for the detailed review.

> I never thought that people would implement both interfaces on the same type

We have a go generate tool that generates the data layer code. A domain type frequently ends up with both types because it will be returned as a series of columns (RowScanner) from some sprocs and as a composite along with other columns (CompositeIndexScanner) in other sprocs. 

We hadn't considered renamed types; the docs describe RowScanner and CompositeIndexScanner separately, and implementing on the type being scanned into seemed like the straightforward mapping. In practice, split types seem like they will run into difficulty because []T vs []Renamed aren't compatible in go, This would work against reuse in nested/array fields.

> Unless you see something wrong with this approach, I would close this PR and the underlying issue. 

Our mental model had been that composite scanning occurs whenever postgres sends a composite value, and RowScanner as a (very useful) convenience for mapping all the columns into a single object. We were surprised that implementing both caused problems when scanning into a single destination. 

If you want to close this pr we'll update our code generator for RowScanner to add the dispatch you suggested. It might be worth documenting this precedence to help other users.

Thanks again for your time!

jackc:
> If you want to close this pr we'll update our code generator for RowScanner to add the dispatch you suggested. It might be worth documenting this precedence to help other users.

Done in b548b13471c8b808f45d8356b96f0ec3026290d6.

PR #2612, opened 2026-07-23
fix: return error from ConnectConfig when config is nil

sonnemusk:
## Summary
`ConnectConfig(ctx, nil)` paniced inside `connConfig.Copy()`.

Return `pgx: ConnectConfig requires a non-nil *ConnConfig` instead.

Note: a non-nil but manually constructed `ConnConfig` still panics as before (`config must be created by ParseConfig`).

## Test plan
- [x] Unit test for nil config


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #2613, opened 2026-07-23
fix: trim surrounding space in ParseConfig connection strings

sonnemusk:
## What

`ParseConfig` / `ParseConfigWithOptions` trim surrounding space on the connection string before parsing.

## Why

Env vars and config files often leave padded DSN values.

## Test

- `TestParseConfigTrimSpace`


sonnemusk:
Closing this PR. After review, this is too low-value / looks like bulk automated noise rather than a carefully chosen change. Sorry for the noise — will focus on real, high-signal fixes only.

PR #2615, opened 2026-07-29
stdlib: Add the GetTxFromDriverTx method

CyJaySong:
In the compatibility mode of the database/sql standard library, obtain pgx.Tx


jackc:
At first glance, the code seems fine. However, it should be tested. And is there some particular use case this enables?

CyJaySong:
> At first glance, the code seems fine. However, it should be tested. And is there some particular use case this enables?

use case for [river with bun](https://riverqueue.com/docs/bun)， `bun` uses `pgx/stdlib` to connect to the database, while `river` uses `pgx`. To ensure consistency, it is necessary to obtain `pgx.Tx` from the transaction of `bun.Tx`

I have already used it in a production environment. This code is simple enough, so I don't intend to write tests for it. It's unlikely to go wrong.

CyJaySong:
@jackc please

Issue #1889, opened 2024-01-30
pgx hangs when trying to acquire a connection from the pool


mpjunior92:
Hi,

I am using `pgxpool` to manage DB connections to a PostgreSQL instance and have observed a very strange behavior after running a simple query on the DB: `pgx` is not able to acquire a connection from the pool; it gets stuck while waiting for DB data to be sent (I left running for 2 hours and no connection was acquired). Here is the call stack collected using `pprof`:
```
#	0x477d5b	internal/poll.runtime_pollWait+0x3b											/usr/local/go/src/runtime/netpoll.go:343
#	0x53d27d	internal/poll.(*pollDesc).wait+0x7d											/usr/local/go/src/internal/poll/fd_poll_runtime.go:84
#	0x53d310	internal/poll.(*pollDesc).waitRead+0x30											/usr/local/go/src/internal/poll/fd_poll_runtime.go:89
#	0x53edce	internal/poll.(*FD).Read+0x40e												/usr/local/go/src/internal/poll/fd_unix.go:164
#	0x72be72	net.(*netFD).Read+0x72													/usr/local/go/src/net/fd_posix.go:55
#	0x74f6fc	net.(*conn).Read+0x9c													/usr/local/go/src/net/net.go:179
#	0x7e1721	crypto/tls.(*atLeastReader).Read+0xc1											/usr/local/go/src/crypto/tls/conn.go:805
#	0x57a41b	bytes.(*Buffer).ReadFrom+0x13b												/usr/local/go/src/bytes/buffer.go:211
#	0x7e1a48	crypto/tls.(*Conn).readFromUntil+0x168											/usr/local/go/src/crypto/tls/conn.go:827
#	0x7df8d8	crypto/tls.(*Conn).readRecordOrCCS+0x258										/usr/local/go/src/crypto/tls/conn.go:625
#	0x7df5e4	crypto/tls.(*Conn).readRecord+0x24											/usr/local/go/src/crypto/tls/conn.go:587
#	0x7e6629	crypto/tls.(*Conn).Read+0x1e9												/usr/local/go/src/crypto/tls/conn.go:1369
#	0xfd158a	github.com/jackc/pgx/v5/pgconn/internal/bgreader.(*BGReader).Read+0x22a							/home/eigenlabs/go/pkg/mod/github.com/jackc/pgx/v5@v5.5.2/pgconn/internal/bgreader/bgreader.go:100
[...]

PR #2619, opened 2026-08-05
pgxpool: accept pool_ping_timeout in the connection string

1991santhu:
Every pool setting on `pgxpool.Config` has a matching `pool_*` connection string key, except `PingTimeout`. `ParseConfig` didn't know the key, so `pgx.ParseConfig` left it in `RuntimeParams` and it got sent to the server as a startup parameter. Postgres then refused every connection in the pool:

```
failed to connect to `user=pgx_md5 database=pgx_test`:
  127.0.0.1:55432 (127.0.0.1): server error: FATAL: unrecognized configuration parameter "pool_ping_timeout" (SQLSTATE 42704)
```

So it's not that the key was ignored. Setting it broke the pool outright, and there was no way to bound the ping from a URL at all.

That matters for #1889. `Acquire` pings a connection that has been idle for a second or more, and with `PingTimeout` at 0 that ping inherits the caller's context. If the caller passes a context with no deadline and the network path has gone away without an RST, which is what the reporters see on Aurora and behind PrivateLink, the read blocks for as long as the socket stays open and `Acquire` never returns. `PingTimeout` is the existing mitigation for that. It was only reachable if you built the config in Go, and most people configure a pool from a URL.

What changed:

- `ParseConfig` consumes `pool_ping_timeout` as a duration string and rejects negative values.
- The default is unchanged. Zero still means no timeout, so nothing moves for existing users.
- Documented `pool_min_idle_conns` on `ParseConfig` too. It's been supported since `MinIdleConns` was added but was never in the list.

I stopped at making the knob reachable rather than picking a non-zero default. A default would change behaviour for everyone and that's your call, not mine. Happy to follow up with one if you want it.

Verified against Postgres 17 in Docker, macOS arm64, Go 1.26.5.

New and existing ping-timeout tests:

```
$ PGX_TEST_DATABASE="postgres://pgx_md5:secret@127.0.0.1:55432/pgx_test" \
    go test ./pgxpool/ -run 'TestParseConfig|TestPoolAcquirePingTimeout' -v
--- PASS: TestParseConfigExtractsPingTimeout (0.00s)
--- PASS: TestParseConfigPingTimeoutDefaultsToZero (0.00s)
--- PASS: TestParseConfigExtractsPoolArguments (0.00s)
--- PASS: TestParseConfigRejectsInvalidPingTimeout (0.00s)
--- PASS: TestPoolAcquirePingTimeout (6.02s)
PASS
ok  	github.com/jackc/pgx/v5/pgxpool	6.225s
```

Whole package:

```
$ PGX_TEST_DATABASE="postgres://pgx_md5:secret@127.0.0.1:55432/pgx_test" go test ./pgxpool/
ok  	github.com/jackc/pgx/v5/pgxpool	6.245s
```

The new tests fail on unpatched `pool.go`:

```
--- FAIL: TestParseConfigExtractsPingTimeout (0.00s)
    map[string]string{"pool_ping_timeout":"250ms"} should not contain "pool_ping_timeout"
--- FAIL: TestParseConfigRejectsInvalidPingTimeout (0.00s)
    pool_ping_timeout=abc should be rejected
    pool_ping_timeout=250 should be rejected
    pool_ping_timeout=-1s should be rejected
```

`go vet ./pgxpool/` and `gofmt -l pgxpool/` are both clean. Diff is two files, no lockfile or generated churn.

---

## Reviewer notes (not for the PR body)

- The `FATAL: unrecognized configuration parameter` output above is real. I reproduced it with a scratch module against the real server on unpatched code, then confirmed it goes away with the patch.
- I deliberately did not add an integration test that puts the key in `PGX_TEST_DATABASE`. CI sets that variable in keyword/value form in one job and as a URI in another, so appending a parameter is not format-safe. The parse-level assertion that the key leaves `RuntimeParams` is what actually prevents the failure, and `TestPoolAcquirePingTimeout` already covers the behaviour once `PingTimeout` is set.
- Governance check: the last 20 merged PRs on `jackc/pgx` are all from non-maintainers.
- No PR references issue #1889.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -344,10 +344,12 @@ func NewWithConfig(ctx context.Context, config *Config) (*Pool, error) {
 //
 //   - pool_max_conns: integer greater than 0 (default 4)
 //   - pool_min_conns: integer 0 or greater (default 0)
+//   - pool_min_idle_conns: integer 0 or greater (default 0)
 //   - pool_max_conn_lifetime: duration string (default 1 hour)
 //   - pool_max_conn_idle_time: duration string (default 30 minutes)
 //   - pool_health_check_period: duration string (default 1 minute)
 //   - pool_max_conn_lifetime_jitter: duration string (default 0)
+//   - pool_ping_timeout: duration string (default 0, meaning no timeout)
 //
 // See Config for definitions of these arguments.
 //
@@ -446,6 +448,18 @@ func ParseConfig(connString string) (*Config, error) {
 			return nil, pgconn.NewParseConfigError(connString, "cannot parse pool_max_conn_lifetime_jitter", err)
 		}
 		config.MaxConnLifetimeJitter = d
+	}
+
+	if s, ok := config.ConnConfig.Config.RuntimeParams["pool_ping_timeout"]; ok {
+		delete(connConfig.Config.RuntimeParams, "pool_ping_timeout")
+		d, err := time.ParseDuration(s)
+		if err != nil {
+			return nil, pgconn.NewParseConfigError(connString, "cannot parse pool_ping_timeout", err)
+		}
+		if d < 0 {
+			return nil, pgconn.NewParseConfigError(connString, "pool_ping_timeout must not be negative", nil)
+		}
+		config.PingTimeout = d
 	}
 
 	return config, nil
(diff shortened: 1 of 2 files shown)


jackc:
Merged with several adjustments.

PR #2620, opened 2026-08-05
fix: handle context cancellation race condition in pgxpool

parastejpal987-cmyk:
### Description
This PR addresses a critical race condition in `pgxpool.Acquire()`. 

Previously, if a connection was assigned to a waiting caller at the exact millisecond that the caller's context was canceled (e.g., via a timeout), the connection would be leaked because it was never returned to the idle pool.

This patch adds a safety check immediately after acquiring the resource. If `ctx.Err() != nil`, it safely calls `res.Release()` to return the connection back to the pool before bubbling the context error up to the caller.

### Changes Made
- Added context cancellation check in `Acquire()`
- Safely releases resources back to the pool to prevent connection leakage under heavy concurrency.

### DIFF
--- a/pgxpool/pool.go
+++ b/pgxpool/pool.go
@@ -619,6 +619,13 @@ func (p *Pool) Acquire(ctx context.Context) (c *Conn, err error) {
 			return nil, err
 		}
 
+		// Handle race condition: If the connection is assigned to this waiter at the exact
+		// moment the waiter's context is canceled, we must safely return it to the idle pool.
+		if ctx.Err() != nil {
+			res.Release()
+			return nil, ctx.Err()
+		}
+
 		cr := res.Value()
 
 		// Destroy expired connections before doing any further work (such as
--- a/pgxpool/repro_test.go
+++ b/pgxpool/repro_test.go
@@ -1,0 +1,55 @@
+package pgxpool_test
+
+import (
+	"context"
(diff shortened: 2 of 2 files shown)


jackc:
Do you have some sort of reproduction case? Based on my investigation, the described race condition does not exist.

parastejpal987-cmyk:
Hi @jackc, I've pushed a new commit that includes a dedicated reproduction case (repro_test.go). It runs a high-concurrency stress test designed specifically to trigger the race condition where a context cancels at the exact millisecond the connection is assigned. If you run the test without my patch, you should see the connection leak and the test fail. With the patch applied, it passes perfectly. Let me know what you think!

jackc:
Thanks for adding the test. I ran it 100 times against both the unpatched parent and current master, and it passed every time. The underlying semaphore and Puddle acquisition paths already return resources when cancellation wins the handoff. If acquisition wins concurrently, the caller receives and owns the connection, so there is no leak. Since this test does not fail without the patch, it does not reproduce the reported bug.

parastejpal987-cmyk:
"Thank you for taking the time to run the tests and review this so thoroughly! It makes perfect sense that puddle is already handling the handoff correctly. I appreciate the explanation, and I'll go ahead and close this PR. Have a great day!"

Issue #2623, opened 2026-08-11
Data races in pgproto3 Frontend when a connection is used from two goroutines during teardown


adisetiawanx:
## Summary

When a single PgConn is driven from two goroutines at once (a CopyFrom that runs on the raw connection obtained through `sql.Conn.Raw`, racing a rollback that database/sql triggers on a cancelled context), running with `-race` reports several data races inside the `pgproto3.Frontend` wire buffer and on `PgConn.txStatus`.

This shows up while testing the reproducer in #2622 once the panics there are fixed. With the fix in place the process no longer panics and completes, but the race detector still reports:

- `pgproto3.(*Frontend).Send()`, `.Flush()`, `.SendParse()`: concurrent writes to the frontend write buffer.
- `pgconn.(*PgConn).TxStatus()` vs `receiveMessage()`: a read/write of the `txStatus` field.

## Repro

Run the reproducer from #2622 under `go test -race` (or `go run -race`):

```
go run -race . -workers 16 -seconds 12
```

It prints `survived` and then reports roughly 7 data races in the locations above.

## Context

The panics that made this scenario crash (double close of cleanupDone, double-arm of the slow write timer, and a double unlock) are addressed separately. The remaining races are unsynchronized memory access rather than panics. They matter because under the race detector they are pointed out as real, and in theory concurrent writes to the frontend buffer could corrupt the wire protocol.

Resolving them means serializing the shared frontend write buffer, which is a larger change to the hot path than the panic fixes, so I am reporting it separately rather than bundling it with #2622.

jackc:
> When a single PgConn is driven from two goroutines at once 

`pgconn.PgConn` makes no claim to be concurrency safe. Using it concurrently is a bug in the caller.

adisetiawanx:
Closing this follow-up. Same reasoning as the PR: concurrent use of a single PgConn is a caller bug, not something the library needs to guard against.

Issue #2553, opened 2026-05-08
pgx v5.9.2 panic "slow write timer already active" on conn destruction after read returns "closed network connection"


RafOSS-br:
**Describe the bug**

`pgxpool` panics with `BUG: slow write timer already active` during conn destruction when a read on a pooled conn fails with `use of closed network connection` (TCP closed externally — `pg_terminate_backend`, network reset, broker kill, etc.) and the pool then tries to destroy that broken conn.

The slow-write timer set by the failed read is never cleared, so the subsequent `pgConn.Close()` from the pool's destruct callback hits `enterPotentialWriteReadDeadlock` with a non-nil `slowWriteTimer` and panics.

**To Reproduce**

The panic is timing-dependent and I haven't been able to deterministically reproduce it in a small standalone program despite ~1M operations with concurrent `pg_terminate_backend` against the workers. The workload pattern that triggers it in production is:

- A `pgxpool` shared by HTTP request handlers (UoW per request) AND a long-lived background ticker that runs `BeginTx` + multi-statement reads on a stream table every few seconds.
- Sustained ~2000 RPS for several minutes.
- The TCP connection used by the background ticker dies externally (TimescaleDB chunk maintenance, OS-level TCP reset, or — in our staging — a load test that triggers a transient connection issue).

A best-effort repro skeleton is below. It runs 16 workers doing `BeginTx` + slow read + `Commit` while a separate raw connection issues `pg_terminate_backend` on active backends every 20 ms. Run for at least 60 s; the panic does not always fire but the workload exercises the same code path. Run with `-race`.

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"sync"
	"sync/atomic"
	"time"

	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/pgxpool"
)

func main() {
	url := os.Getenv("DATABASE_URL")
	cfg, err := pgxpool.ParseConfig(url)
	if err != nil {
		log.Fatal(err)
	}
	cfg.MaxConns = 8
	cfg.MinConns = 2
	pool, err := pgxpool.NewWithConfig(context.Background(), cfg)
	if err != nil {
		log.Fatal(err)
	}
	defer pool.Close()

	killer, err := pgx.Connect(context.Background(), url)
	if err != nil {
		log.Fatal(err)
	}
	defer killer.Close(context.Background())

	const workers, duration = 16, 60 * time.Second
	stop := make(chan struct{})
[...]

Issue #2332, opened 2025-05-25
panic: BUG: slow write timer already active


kaber2:
**Describe the bug**

Occasionally during heavy insert workloads, I get the aforementioned panic.

**To Reproduce**

I can't reliably reproduce it, I have 5 instances of a program running, each one running a heavy insert-only workload with many 10k inserts/sec total, using multiple parallel connections from a pool with CopyFrom. The workload only runs once every couple of weeks and sometimes it works fine, sometimes I receive the panic.

I have the following backtrace:

```
  goroutine 1 [chan receive]:    
          /home/xxx/go/pkg/mod/github.com/jackc/puddle/v2@v2.2.2/pool.go:76 +0x90    
  created by github.com/jackc/puddle/v2.(*Resource[...]).Destroy in goroutine 3709477    
          /home/xxx/go/pkg/mod/github.com/jackc/puddle/v2@v2.2.2/pool.go:688 +0x36      
  github.com/jackc/puddle/v2.(*Pool[...]).destroyAcquiredResource(0xc5facffc70, 0xcdd01eee80?)    
          /home/xxx/go/pkg/mod/github.com/jackc/puddle/v2@v2.2.2/pool.go:708 +0x1f      
  github.com/jackc/puddle/v2.(*Pool[...]).destructResourceValue(0xcdfb311c20, 0xccca96c008?)    
          /home/xxx/go/pkg/mod/github.com/jackc/pgx/v5@v5.7.1/pgxpool/pool.go:259 +0x9a      
  github.com/jackc/pgx/v5/pgxpool.NewWithConfig.func2(0xcd7c8f2a10)    
          /home/xxx/go/pkg/mod/github.com/jackc/pgx/v5@v5.7.1/conn.go:303 +0x31      
  github.com/jackc/pgx/v5.(*Conn).Close(0x3ae5dd0?, {0x3ae5eb0?, 0xce0201cd20?})    
          /home/xxx/go/pkg/mod/github.com/jackc/pgx/v5@v5.7.1/pgconn/pgconn.go:663 +0x185      
  github.com/jackc/pgx/v5/pgconn.(*PgConn).Close(0xc8836bd208, {0x3ae5eb0, 0xce0201cd20})    
          /home/xxx/go/pkg/mod/github.com/jackc/pgx/v5@v5.7.1/pgconn/pgconn.go:1875 +0xa6      
[...]

PR #2624, opened 2026-08-11
pgconn: make connection close single-shot and concurrency-safe

### LINKED ISSUE #2622: panic: close of closed channel — the `status` guard around `cleanupDone` is an unsynchronized test-and-set (reproducer included)
## Summary

`pgConn.cleanupDone` is closed from four places, each guarded only by a plain
read/write of `pgConn.status`, an ordinary `byte` field with no mutex or atomic
(`pgconn/pgconn.go:92`):

| line (v5.9.2) | site |
|---|---|
| 633-635 | `receiveMessage`, FATAL `ErrorResponse` with `OnPgError` returning false |
| 689-694 | `Close` |
| 723-729 | `asyncClose` |
| 1443-1445 | `CopyFrom`, `bufferingReceiveErr` path |

Because the guard is a non-atomic test-and-set, two goroutines can both pass it
and both close the channel:

```go
func (pgConn *PgConn) Close(ctx context.Context) error {
	if pgConn.status == connStatusClosed {   // read
		return nil
	}
	pgConn.status = connStatusClosed         // write

	defer close(pgConn.cleanupDone)
	...
	return pgConn.conn.Close()               // panics here, on the deferred close
}
```

The window is wide: between the guard and the deferred close there is a full
`Terminate` + flush round trip.

`Close`, `asyncClose` and the `CopyFrom` path each already handle "somebody
else closed this" incorrectly, and #2470, #2364 and #1920 each repaired one
call path. The guard itself is still unsynchronized on master, so the next
combination of paths panics again — the reproducer below hits two different
ones on v5.9.2 and v5.10.0.

## Reproducer

Needs a reachable postgres, and nothing else — dropped into an empty module,
`go mod tidy` resolves it to v5.10.0. As published below it panics within 25 s
on darwin/arm64, in 4 runs out of 4. The `-fix` variant, which differs only in
the context the transaction is begun on, survives all 3 runs.

```go
// go run .        -> panics
// go run . -fix   -> does not
package main

import (
	"context"
	"database/sql"
	"flag"
	"fmt"
	"math/rand/v2"
	"os"
	"sync"
	"time"

	"github.com/jackc/pgx/v5"
	"github.com/jackc/pgx/v5/stdlib"
)

var (
	fix     = flag.Bool("fix", false, "begin the transaction on an uncancellable context")
	workers = flag.Int("workers", 32, "concurrent workers")
	seconds = flag.Int("seconds", 25, "run time")
	dsn     = flag.String("dsn", "host=localhost port=5432 dbname=postgres sslmode=disable", "")
)

func main() {
	flag.Parse()
	db, err := sql.Open("pgx", *dsn)
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(2)
	}
	defer db.Close()
	db.SetMaxOpenConns(*workers)
	db.SetMaxIdleConns(*workers)

	if err := db.PingContext(context.Background()); err != nil {
		fmt.Fprintln(os.Stderr, "no postgres:", err)
		os.Exit(2)
	}

	deadline := time.Now().Add(time.Duration(*seconds) * time.Second)
	var wg sync.WaitGroup
	for range *workers {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for time.Now().Before(deadline) {
				round(db)
			}
		}()
	}
	wg.Wait()
	fmt.Println("survived")
}

func round(db *sql.DB) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	conn, err := db.Conn(ctx)
	if err != nil {
		return
	}
	defer conn.Close()

	var pgxConn *pgx.Conn
	if err := conn.Raw(func(dc any) error {
		pgxConn = dc.(*stdlib.Conn).Conn()
		return nil
	}); err != nil {
		return
	}

	beginCtx := ctx
	if *fix {
		beginCtx = context.WithoutCancel(ctx)
	}
	tx, err := conn.BeginTx(beginCtx, nil)
	if err != nil {
		return
	}
	if _, err := tx.ExecContext(ctx,
		`CREATE TEMP TABLE IF NOT EXISTS bulk (id int, name text) ON COMMIT DROP`); err != nil {
		_ = tx.Rollback()
		return
	}

	// cancel while the COPY is streaming
	time.AfterFunc(time.Duration(rand.IntN(3000))*time.Microsecond, cancel)

	rows := make([][]any, 20000)
	for i := range rows {
		rows[i] = []any{i, "some padding value to make the copy stream take a while"}
	}
	_, _ = pgxConn.CopyFrom(ctx, pgx.Identifier{"bulk"}, []string{"id", "name"},
		pgx.CopyFromRows(rows))

	_ = tx.Rollback()
}
```

To be explicit about what the reproducer does, since it matters for the
diagnosis: the transaction is a `database/sql` `Tx`, and the bulk load runs on
the `*pgx.Conn` obtained through `sql.Conn.Raw` — the documented escape hatch,
and the only way to get COPY performance under `database/sql`. That COPY is
therefore outside the per-connection lock `database/sql` puts around every
other statement, which is what lets two closers meet.

## Stacks

The reproducer produces two distinct panics from the same defect.

`CopyFrom`'s inline close (3 of 4 runs) — the site of #2364:

```
panic: close of closed channel
github.com/jackc/pgx/v5/pgconn.(*PgConn).CopyFrom(...)
	pgx@v5.9.2/pgconn/pgconn.go:1445
github.com/jackc/pgx/v5.(*copyFrom).run(...)
	pgx@v5.9.2/copy_from.go:202
github.com/jackc/pgx/v5.(*Conn).CopyFrom(...)
	pgx@v5.9.2/copy_from.go:275
```

`Close` from the transaction rollback (1 of 4 runs) — this is the stack that
originally took a server down and prompted the investigation:

```
panic: close of closed channel
github.com/jackc/pgx/v5/pgconn.(*PgConn).Close(...)
	pgx@v5.9.2/pgconn/pgconn.go:717
github.com/jackc/pgx/v5.(*Conn).die(...)
	pgx@v5.9.2/conn.go:445
github.com/jackc/pgx/v5.(*dbTx).Rollback(...)
	pgx@v5.9.2/tx.go:218
github.com/jackc/pgx/v5/stdlib.wrapTx.Rollback(...)
	pgx@v5.9.2/stdlib/sql.go:909
database/sql.(*Tx).rollback.func1()
	/usr/local/go/src/database/sql/sql.go:2345
database/sql.(*Tx).awaitDone(...)
	/usr/local/go/src/database/sql/sql.go:2221
created by database/sql.(*DB).beginDC
```

`database/sql.beginDC` unconditionally spawns `go tx.awaitDone()` for every
`BeginTx(ctx, …)`, so on cancellation that goroutine rolls back; the `rollback`
`Exec` fails on the dead context, pgx takes the error branch at `tx.go:218` and
calls `conn.die()` → `PgConn.Close()`, concurrently with the COPY's own
teardown.

## Still present on master

v5.10.0 has the identical code — `Close` at 738-745, `asyncClose` at 772-780,
`IsClosed` at 811-813, `CopyFrom`'s inline close unchanged.
`diff v5.9.2 v5.10.0 pgconn/pgconn.go` shows no change touching `cleanupDone`.

The stacks above are from v5.9.2, but the reproducer was last run on v5.10.0,
where the same two sites panic at their shifted line numbers: `pgconn.go:767`
(the `return pgConn.conn.Close()` whose deferred `close(cleanupDone)` at 744
panics) in 2 of 4 runs, and `pgconn.go:1513` (`CopyFrom`) in the other 2.

## Suggested fix

Make the transition atomic and the close single-shot, so no combination of call
paths can double-close:

```go
type PgConn struct {
	...
	status      atomic.Uint32
	cleanupOnce sync.Once
}

func (pgConn *PgConn) markClosed() bool {
	return pgConn.status.Swap(connStatusClosed) != connStatusClosed
}

func (pgConn *PgConn) finishCleanup() {
	pgConn.cleanupOnce.Do(func() { close(pgConn.cleanupDone) })
}
```

`sync.Once` alone removes the panic; making `status` atomic additionally fixes
the data race and makes `IsClosed()` meaningful to callers.

Adjacent, same path: with `CancelRequestContextWatcherHandler`, `HandleCancel`
runs on the `ctxwatch` goroutine and its inner goroutine reads `pgConn.conn`,
`pgConn.pid` and `pgConn.secretKey` in `CancelRequest` while the owning
goroutine may be in `Close()` calling `pgConn.conn.Close()`. That is the same
shape as the connect race fixed in v5.10.0 ("Fix data race when context is
cancelled during connect"), but on the query path.

## Version

- pgx: v5.9.2 (verified unchanged on v5.10.0)
- Go: 1.26.0
- PostgreSQL: 18.3
- OS/arch: darwin/arm64

---

Reported by Claude Opus 5 on behalf of Martin Rode (Programmfabrik GmbH, fylr).

adisetiawanx:
Fixes #2622 and, along the way, the same "BUG:" panic family reported in #2332 and #2553. These let a connection torn down from two paths at once panic and kill the whole process instead of failing gracefully.

## The problem

`PgConn.cleanupDone` was closed from four places: `Close`, `asyncClose`, the `receiveMessage` OnPgError branch, and CopyFrom's buffering-receive error branch. Each was guarded only by a plain read-then-write of `PgConn.status`, a plain `byte` field with no synchronization. Two goroutines could both pass the guard and both close the channel, which panics with "close of closed channel". The reproducer in #2622 (a COPY streaming on a raw connection while the transaction context is cancelled) hits this reliably on master.

While testing that reproducer against a live PostgreSQL I found the same teardown scenario trips two more defensive panics in the same class:

- "slow write timer already active": the final COPY flush and the Close teardown flush can overlap within the 15ms window and both arm the single shared slow write timer. #2332 reports this panic under a heavy CopyFrom workload.
- "cannot unlock unlocked connection": `lock()` used a plain store for the busy transition, so two callers could both conclude they owned the connection and the second unlock would have nothing to unlock.

## The fix

Four changes, split into logical commits:

1. **Make closing single-shot.** `markClosed()` transitions the connection to closed with an atomic swap so exactly one goroutine is the closer, and `cleanupDone` is closed through a `sync.Once` (`finishCleanup`). All four close paths now go through these two helpers.
2. **Serialize the slow write timer.** `enter`/`exitPotentialWriteReadDeadlock` now hold a small mutex across the enter/write/exit lifecycle, so concurrent flushes queue instead of double-arming the timer and panicking. In normal use the mutex is uncontended since only one flush is in progress at a time.
3. **Make `lock()` a compare-and-swap** and reject locking a connection that is still connecting. Only one caller wins the busy transition, which makes the double-unlock unreachable.
4. **CopyFrom now closes `abortCopyChan` on the buffering-receive error path**, so the copy writer goroutine stops writing to the connection while teardown proceeds. That path was the only early-return path missing the close the others already did.

The `status` field is now an `atomic.Uint32`. This is not cosmetic: it removes the data race on the guard itself, makes the single-closer swap possible, and lets `IsClosed` and lock/unlock work while close paths run. Every read and write site was converted to Load/Store or CAS.

## Testing

- The reproducer from #2622: on master it panics with "close of closed channel". With this change it runs to completion, printing "survived", repeatedly and across 32, 48 and 64 workers against PostgreSQL 18.
- The full `pgconn` test suite passes against a live PostgreSQL 18.
- The full `pgconn` suite passes under `-race` (Go 1.26, Linux).
- New unit tests assert `markClosed` reports a single winner under concurrency and that `finishCleanup` never closes the channel twice.

## Known limitation

Under `-race` the #2622 reproducer still reports a handful of data races inside the `pgproto3` wire buffer (`Frontend.Send`, `Frontend.Flush`) and on `pgConn.txStatus`. These only appear because the reproducer drives one connection from two goroutines at once (a CopyFrom on the raw connection obtained via `sql.Conn.Raw`, racing the rollback database/sql spawns on a cancelled context). They do not panic and the run completes, but they are real unsynchronized accesses, so I documented them in #2623 with the details rather than silently bundling a larger frontend change into this PR.

### DIFF
--- a/pgconn/cleanup_single_shot_test.go
+++ b/pgconn/cleanup_single_shot_test.go
@@ -1,0 +1,72 @@
+package pgconn
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// https://github.com/jackc/pgx/issues/2622
+//
+// The status guard around cleanupDone used to be a non-atomic test-and-set,
+// so two close paths (Close, asyncClose, receiveMessage's OnPgError branch,
+// CopyFrom's buffering-receive error branch) could both pass it and both call
+// close(cleanupDone), panicking with "close of closed channel".
+//
+// These test the mechanism the fix relies on directly: exactly one goroutine
+// can win markClosed, and finishCleanup is safe to call from any number of
+// goroutines at once.
+func TestMarkClosedIsSingleShot(t *testing.T) {
+	t.Parallel()
+
+	pgConn := &PgConn{cleanupDone: make(chan struct{})}
+
+	const callers = 128
+	results := make(chan bool, callers)
+	var wg sync.WaitGroup
+	for i := 0; i < callers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			results <- pgConn.markClosed()
+		}()
+	}
+	wg.Wait()
+	close(results)
+
+	won := 0
+	for r := range results {
+		if r {
+			won++
+		}
+	}
+
+	require.Equal(t, 1, won, "exactly one goroutine must be reported as the closer")
+	require.False(t, pgConn.markClosed(), "markClosed must report false once already closed")
+	require.True(t, pgConn.IsClosed(), "connection must be reported closed after markClosed")
+}
+
+func TestFinishCleanupIsSingleShot(t *testing.T) {
+	t.Parallel()
+
+	pgConn := &PgConn{cleanupDone: make(chan struct{})}
+
+	const callers = 128
+	var wg sync.WaitGroup
+	for i := 0; i < callers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			// Must not panic by closing the channel more than once.
+			pgConn.finishCleanup()
+		}()
+	}
+	wg.Wait()
+
+	select {
+	case <-pgConn.cleanupDone:
+	default:
+		t.Fatal("cleanupDone must be closed after finishCleanup")
+	}
+}
--- a/pgconn/pgconn.go
+++ b/pgconn/pgconn.go
@@ -116,6 +116,13 @@ type PgConn struct {
 	// cleanupOnce guarantees cleanupDone is closed exactly once, no matter how
 	// many close paths run concurrently.
 	cleanupOnce sync.Once
+
+	// slowWriteTimerMux serializes the slow write timer lifecycle across
+	// enter/exitPotentialWriteReadDeadlock pairs. Concurrent flushes (for
+	// example CopyFrom completing its final flush while Close tears the
+	// connection down) would otherwise both arm the single shared timer and
+	// trip the "slow write timer already active" panic.
+	slowWriteTimerMux sync.Mutex
 }
 
 // Connect establishes a connection to a PostgreSQL server using the environment and connString (in URL or keyword/value
@@ -851,16 +858,24 @@ func (pgConn *PgConn) IsBusy() bool {
 
 // lock locks the connection.
 func (pgConn *PgConn) lock() error {
-	switch pgConn.status.Load() {
-	case connStatusBusy:
-		return &connLockError{status: "conn busy"} // This only should be possible in case of an application bug.
-	case connStatusClosed:
-		return &connLockError{status: "conn closed"}
-	case connStatusUninitialized:
-		return &connLockError{status: "conn uninitialized"}
+	for {
+		switch pgConn.status.Load() {
+		case connStatusBusy:
+			return &connLockError{status: "conn busy"} // This only should be possible in case of an application bug.
+		case connStatusClosed:
+			return &connLockError{status: "conn closed"}
+		case connStatusUninitialized:
+			return &connLockError{status: "conn uninitialized"}
+		}
+		// CAS so only one caller wins when the connection is acquired from
+		// multiple goroutines (e.g. a CopyFrom on the raw connection racing
+		// a rollback triggered by database/sql). A plain store here would let
+		// two lockers both think they own the connection and the second
+		// unlock would panic.
+		if pgConn.status.CompareAndSwap(uint32(connStatusIdle), uint32(connStatusBusy)) {
+			return nil
+		}
 	}
-	pgConn.status.Store(uint32(connStatusBusy))
-	return nil
 }
 
 func (pgConn *PgConn) unlock() {
@@ -1544,6 +1559,9 @@ func (pgConn *PgConn) CopyFrom(ctx context.Context, r io.Reader, sql string) (Co
 					pgConn.conn.Close()
 					pgConn.finishCleanup()
 				}
+				// Stop the copy writer goroutine so it does not keep writing
+				// to the connection while the teardown proceeds.
+				close(abortCopyChan)
 				return CommandTag{}, normalizeTimeoutError(ctx, err)
 			}
 			// peekMessage never returns err in the bufferingReceive mode - it only forwards the bufferingReceive variables.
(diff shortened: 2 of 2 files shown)


jackc:
`pgconn.PgConn` makes no claim to be concurrency safe. Using it concurrently is a bug in the caller.

adisetiawanx:
Thanks, that makes sense. You're right about the reproducer too. Carrying the connection out of the Raw closure is what creates the two-goroutine situation, and that's on the caller side, not something PgConn should have to handle.

Since every double-close path needs that concurrent use and PgConn doesn't claim to be safe for it, I'll close this. Appreciate you taking the time to look at it.

adisetiawanx:
Closing this. The underlying panics are real, but the only way to trigger them is concurrent use of a single PgConn, which is outside the documented contract. Thanks again to jackc for the clarification.

PR #2625, opened 2026-08-12
pgconn: use the default port for a URL host that omits its port

sueun-dev:
`ParseConfig` pairs the wrong port with a host in a multi-host connection URL when one of the hosts omits its port.

`parseURLSettings` splits `url.Host` on commas but only appends to its `ports` slice when a host carries an explicit port. As soon as an earlier host omits its port, that slice is no longer aligned with the host list, and the fallback builder in `ParseConfigWithOptions` then reads the wrong element for each host (falling back to `ports[0]` once it runs off the end).

So for `postgres://foo,bar:5433/mydb` pgx connects `foo` on 5433 instead of the default 5432, and for `postgres://foo:1,bar,baz:3/mydb` the explicit ports shift onto the wrong hosts. libpq keeps the two lists aligned and resolves an omitted per-host port to the default:

```
$ PGCONNECT_TIMEOUT=1 psql "postgresql://192.0.2.1,192.0.2.2:5433,192.0.2.3:9999/db"
psql: error: connection to server at "192.0.2.1", port 5432 failed: timeout expired
connection to server at "192.0.2.2", port 5433 failed: timeout expired
connection to server at "192.0.2.3", port 9999 failed: timeout expired
```

pgx before this change resolves that URL to `192.0.2.1:5433, 192.0.2.2:9999, 192.0.2.3:5433`. This bites HA connection strings that list, say, a primary and a replica on different ports — the primary silently ends up on the replica's port.

The fix keeps `ports` positionally aligned with `hosts` (a host that omits its port contributes an empty element) and resolves an empty element to the default port (PGPORT or the built-in default) when the fallbacks are built. That matches the documented libpq rule: "An empty string, or an empty item in a comma-separated list, specifies the default port number." A URL where no host specifies a port is unchanged, since the port is only recorded when at least one host set it.

Added two `TestParseConfig` cases (a leading and a middle omitted port); both fail before and pass after. `go test ./pgconn/` passes.

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -474,12 +474,23 @@ func ParseConfigWithOptions(connString string, options ParseConfigOptions) (*Con
 	hosts := strings.Split(settings["host"], ",")
 	ports := strings.Split(settings["port"], ",")
 
+	// A host whose port was omitted arrives here as an empty port element. libpq
+	// resolves it to the default port (PGPORT or the compiled-in default), not to
+	// another host's port.
+	defaultPort := envSettings["port"]
+	if defaultPort == "" {
+		defaultPort = defaultSettings["port"]
+	}
+
 	for i, host := range hosts {
 		var portStr string
 		if i < len(ports) {
 			portStr = ports[i]
 		} else {
 			portStr = ports[0]
+		}
+		if portStr == "" {
+			portStr = defaultPort
 		}
 
 		port, err := parsePort(portStr)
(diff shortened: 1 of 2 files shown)


jackc:
This issue was the last straw for keeping the net/url parser as the core. Instead of one more workaround to match libpq behavior, I decided on 023281e597e33a62e02feadcac0845a745b2b002. It introduces a new parser designed to match libpq. This resolves this issue as well as several other edge incompatibilities and makes it easier to match any libpq changes in the future.

PR #2633, opened 2026-08-20
Fix panic on unterminated quoted value ending in backslash

ProjectMutilation:
A backslash as the last byte of a quoted keyword/value value panicked with "slice bounds out of range [:2] with length 1". The quoted-string loop skips the byte after a backslash by incrementing end, and the for loop then increments it again, so `='\` (or `0='\`, reached via pgxpool.ParseConfig) drove end past len(s). The subsequent bounds check compared end == len(s), which no longer matched, and s[:end] sliced past the end of the string.

The unquoted branch has guarded this since be69c1c1 ("Fix parseDSNSettings with bad backslash"), but that fix missed the quoted branch, which has carried the same unguarded increment since the parser was ported from pgx v3 in 99f22ac8. 

Found by go native fuzzing.

The quoted branch now bounds-checks end after the backslash skip, matching the unquoted branch, and returns "unterminated quoted string in connection info string" instead of panicking. Regression test covers both crashing inputs and the value-prefixed variants.

### DIFF
--- a/pgconn/config.go
+++ b/pgconn/config.go
@@ -792,6 +792,9 @@ func parseKeywordValueSettings(s string) (map[string]string, error) {
 				}
 				if s[end] == '\\' {
 					end++
+					if end == len(s) {
+						return nil, errors.New("unterminated quoted string in connection info string")
+					}
 				}
 			}
 			if end == len(s) {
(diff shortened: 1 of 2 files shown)


jackc:
Merged after a rebase + some additional changes to match libpq more closely.

PR #2641, opened 2026-08-28
Replace error messages with predefined variables for Batch

mike-sb:
Added predifined errors for batch funcs to consume them with errors.Is.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -8,6 +8,10 @@ import (
 	"github.com/jackc/pgx/v5/pgconn"
 )
 
+var ErrBatchClosed = errors.New("batch already closed")
+var ErrBatchEmpty = errors.New("no more results in batch")
+var ErrBatchReference = errors.New("no reference to batch")
+
 // QueuedQuery is a query that has been queued for execution via a [Batch].
 type QueuedQuery struct {
 	SQL       string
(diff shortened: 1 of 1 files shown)


jackc:
I'm not necessarily opposed to this, but I do have one concern.

In general, I think of named errors as existing because the caller is going to do something based on the type of error. But as far as I can tell, all of these errors are actually caller bugs. In that case, I don't see why the caller would check for them. If they happen the caller should be changed so they never occur.

But perhaps I'm missing some use case...

mike-sb:
> I'm not necessarily opposed to this, but I do have one concern.
> 
> In general, I think of named errors as existing because the caller is going to do something based on the type of error. But as far as I can tell, all of these errors are actually caller bugs. In that case, I don't see why the caller would check for them. If they happen the caller should be changed so they never occur.
> 
> But perhaps I'm missing some use case...

I encountered a case while processing a Batch where I wanted to check if it had finished. I couldn't use `errors.Is` for the comparison because a new error instance was created each time. It seems to me it would be better to iterate over the array of Batch responses rather than the original input array, since we are dealing with database behavior rather than user data.

jackc:
Sorry, still don't understand. These errors should only occur due to a bug in the caller. I don't see how checking for those errors would ever be correct caller behavior.

If you are trying to use these errors as a sentinel value to read all results by calling `Exec()` /  `Query()` / `QueryRow()` on the `BatchResults` then I would suggest not doing that. It's usually cleaner to call those methods on the `QueueQuery`, then the handling of the query is in close proximity to the query rather than separated in the code.

PR #2642, opened 2026-08-28
Fix deallocation of failed digested prepares

### LINKED ISSUE #2640: stdlib: failed prepare is deallocated by SQL text instead of its server-side statement name — leaked statement, 42P05 re-poisoning, and errors surfaced on unrelated queries
## Version

pgx v5.10.0 (the relevant code is unchanged on master as of 2026-08-28).
Driven through `database/sql` (`jackc/pgx/v5/stdlib`), as used by GORM with
`PrepareStmt: true`. PostgreSQL 16.

## The mechanism

`stdlib.Conn.PrepareContext` calls `c.conn.Prepare(ctx, query, query)` — i.e.
`name == sql`. In `pgx.Conn.Prepare` that path derives a digest name for the
wire protocol:

```go
if name == sql {
    digest := sha256.Sum256([]byte(sql))
    psName = "stmt_" + hex.EncodeToString(digest[0:24])
    psKey = sql
}
...
sd, err = c.pgConn.Prepare(ctx, psName, sql, nil)
if err != nil {
    var pErr *pgconn.PrepareError
    if errors.As(err, &pErr) {
        c.failedDescribeStatement = psKey   // ← the SQL text, not psName
    }
    return nil, err
}
```

On the next `Prepare` on that connection, the cleanup introduced in 5.8.0
("Fix: Handle for preparing statements that fail during the Describe phase")
runs:

```go
if c.failedDescribeStatement != "" {
    err = c.Deallocate(ctx, c.failedDescribeStatement)
    ...
}
```

`Conn.Deallocate(ctx, name)` maps `name → sd.Name` via
`c.preparedStatements[name]` — but a FAILED prepare was never stored there, so
it falls through to `psName = name` and sends the protocol `Close` message
with **the full SQL text as the statement name**. No statement by that name
exists, closing a nonexistent statement is not an error, so the cleanup
"succeeds" while closing nothing.

## Consequences

When the original `PrepareError` had `ParseComplete == true` (Parse succeeded,
the error arrived during Describe — e.g. a `statement_timeout` or
`pg_cancel_backend` landing between Parse and Sync):

1. **The server-side statement `stmt_<digest>` leaks** — the deallocation that
   was supposed to remove it targets the wrong name.
2. **Any retry of the same SQL text on that connection is permanently broken**:
   `c.preparedStatements[sql]` is empty, so `Prepare` sends Parse for
   `stmt_<digest>` again → `42P05 duplicate_prepared_statement` → that is
   itself a `PrepareError`, so `failedDescribeStatement` is set again → the
   next cleanup no-ops again → loop, for the remaining lifetime of the
   connection. This is precisely the failure class the 5.8.0 fix was meant to
   close (#2223), still reachable through the stdlib path.
3. **The cleanup's own failure surfaces on innocent queries.** The deallocate
   round-trip (`Close` + `Sync` + read) runs at the START of whatever query
   next needs a prepare on that pooled connection. If it fails (context
   deadline, network hiccup), that unrelated query fails with

   ```
   failed to deallocate previously failed statement "SELECT ... <the OLD query's full SQL> ...": <cause>
   ```

   We hit this in production: an attendance-ingest query's prepare failure
   poisoned a pooled connection, and a completely unrelated background job
   that later borrowed the connection died with an error quoting the ingest
   query's SQL — genuinely confusing to debug from the victim's side.

## Where the fix probably belongs

`failedDescribeStatement` stores `psKey`, but the deallocation needs `psName`
(what was actually sent in the Parse message). Remembering the psName
alongside (or instead of) the psKey — and having the cleanup deallocate by the
server-side name — closes both the leak and the 42P05 loop:

```go
// on failure:
c.failedDescribeStatement = psName   // or a second field carrying psName
```

(The `QueryExecModeCacheStatement` path is unaffected: there `name` is
`stmtcache.StatementName(sql)` and `psName == psKey`, so the deallocation
already targets the right name. Only the `name == sql` route — the
stdlib/database-sql route — has the mismatch.)

Separately, it may be worth discussing whether a failed cleanup should fail
the innocent caller at all, or instead invalidate the connection (as other
unrecoverable states do) so the pool discards it.

## Repro sketch

Timing-dependent (needs the server error to land after Parse completes), so
this is probabilistic rather than a unit test:

```go
db, _ := sql.Open("pgx", dsn) // or GORM with PrepareStmt: true
db.Exec("SET statement_timeout = '5ms'")
// Prepare statements with large/slow-to-plan text under load until one
// PrepareContext returns a PrepareError with ParseComplete == true.
// From then on, on that pooled connection:
//   - re-preparing the SAME SQL fails with 42P05 (consequence 2)
//   - SELECT name FROM pg_prepared_statements shows the leaked stmt_<digest>
//     (consequence 1)
```

The naming mismatch itself (consequence 1) is directly visible without timing
tricks by inspecting what `Conn.Deallocate` sends when handed a key that is
not in `c.preparedStatements`: the `Close` message carries the raw key.

Happy to provide more detail from our production traces if useful.

AshSgDe29071999:
Fixes #2640

`database/sql` / GORM with `PrepareStmt: true` call `Prepare(ctx, sql, sql)`. In that path the wire name is `stmt_<digest>` while the client cache key is the SQL text.

On a Describe-phase `PrepareError`, `failedDescribeStatement` stored the cache key. The next `Prepare` on that connection then called `Deallocate` with the SQL text. A failed prepare is never in `preparedStatements`, so `Deallocate` sent `Close` for a name that does not exist, the leaked `stmt_<digest>` stayed on the server, and a retry of the same SQL returned `42P05` for the rest of the connection's life. Cleanup errors also surfaced on whatever query next borrowed the pooled connection.

Store `psName` (the identifier actually sent in Parse) so the follow-up `Close` targets the real server statement. The named-prepare path is unchanged because `psName == psKey` there.

The new test is the digest (`name == sql`) variant of `TestPrepareHandlesTimeoutBetweenParseAndDescribe`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -360,7 +360,8 @@ func (c *Conn) Prepare(ctx context.Context, name, sql string) (sd *pgconn.Statem
 	if err != nil {
 		var pErr *pgconn.PrepareError
 		if errors.As(err, &pErr) {
-			c.failedDescribeStatement = psKey
+			// psName is the identifier sent in Parse; psKey may be the SQL text.
+			c.failedDescribeStatement = psName
 		}
 		return nil, err
 	}
--- a/conn_test.go
+++ b/conn_test.go
@@ -532,6 +532,66 @@ func TestPrepareHandlesTimeoutBetweenParseAndDescribe(t *testing.T) {
 	require.NotNil(t, psd)
 }
 
+// https://github.com/jackc/pgx/issues/2640
+func TestPrepareDigestedNameDeallocatesFailedDescribe(t *testing.T) {
+	// Not parallel because it is a timing sensitive test.
+	//
+	// stdlib/database/sql calls Prepare(ctx, sql, sql), so the server-side
+	// name is stmt_<digest> while the client cache key is the SQL text.
+	// Cleanup after a Describe-phase failure must Close the digest name.
+
+	config, err := pgx.ParseConfig(os.Getenv("PGX_TEST_DATABASE"))
+	require.NoError(t, err)
+
+	var faultyConn *faultyconn.Conn
+	config.AfterNetConnect = func(ctx context.Context, config *pgconn.Config, conn net.Conn) (net.Conn, error) {
+		faultyConn = faultyconn.New(conn)
+		return faultyConn, nil
+	}
+
+	ctx := context.Background()
+	conn, err := pgx.ConnectConfig(ctx, config)
+	require.NoError(t, err)
+	defer closeConn(t, conn)
+	require.NotNil(t, faultyConn)
+
+	pgxtest.SkipCockroachDB(t, conn, "Induced error does not occur on CockroachDB")
+
+	_, err = conn.Exec(ctx, "set statement_timeout = '100ms'")
+	require.NoError(t, err)
+
+	faultyConn.HandleFrontendMessage = func(backendWriter io.Writer, msg pgproto3.FrontendMessage) error {
+		if _, ok := msg.(*pgproto3.Describe); ok {
+			time.Sleep(200 * time.Millisecond)
+		}
+		buf, err := msg.Encode(nil)
+		if err != nil {
+			return err
+		}
+		_, err = backendWriter.Write(buf)
+		return err
+	}
+
+	sql := "select $1::varchar"
+	psd, err := conn.Prepare(ctx, sql, sql)
+	var pgErr *pgconn.PgError
+	require.ErrorAs(t, err, &pgErr)
+	require.Equal(t, "57014", pgErr.Code)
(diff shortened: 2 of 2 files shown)


eliranbz:
Thanks for the PR, but I could have also let CC fix this in a second - I want to get feedback and see if this issue goes deeper or is this an issue on my implementation rather than a bug.

@jackc this seem to solve the issue, but would be happy to hear your thoughts on the issue. i have created https://github.com/jackc/pgx/pull/2644 for this.  this PR stops pointless cleanups from being scheduled at all, with tests that prove both the original issue and this

jackc:
Closed in favor of #2644

Issue #1667, opened 2023-06-30
Lazy transaction BEGIN


redbaron:
**Is your feature request related to a problem? Please describe.**
 On a high latency connections every additional roundtrip between client and server is undesirable. `Conn.BeginTx` starts transaction by `Conn.Exec`ing `BEGIN` command, which internally waits for result. This wait time is a wasteful and app can't do much about it.

**Describe the solution you'd like**
Delay `BEGIN` and sendit it together with first query in transaction, essentially making it lazy. Alternatively utilize Pipeline/Batch to send it in `BeginTx`, but don't wait for it's result until first query in transaction is issued.

**Describe alternatives you've considered**
App can chose to use Batch and manually roll `BEGIN`/`COMMIT` into the batch, but then it loses convenience of `Tx` API.

jackc:
👍 for the idea.

But not sure about what interface would make sense or exactly how to implement it. This behavior might need to be opt-in. It might be tricky to transparently implement lazy behavior if the first action is `Tx.SendBatch` or `Tx.LargeObjects` for example.

Also, if minimizing round trips are critical, then the commit is just as big a problem. Ideally, there would be a way to bundle the final query and the commit together as well.

redbaron:
> But not sure about what interface would make sense or exactly how to implement it.

BeginTx creates Batch with one command: `BEGIN` and saves `BatchResult` in `Tx`. On  the first query attempt it calls `BatchResult.Close()` and proceeds to executing query as usual.


> It might be tricky to transparently implement lazy behavior if the first action is Tx.SendBatch or Tx.LargeObjects for example.

I am not familiar with protocol or pgx internals, why do you think it might be tricky?

> This behavior might need to be opt-in.
 Less knobs is better. Only user visible change should be that now `Query` can return errors related to transaction creation. AFAIK correct use of `BEGIN` never fails I don't think any app will see this change. You have support large userbase so maybe knob is a safer bet.

redbaron:
> Also, if minimizing round trips are critical, then the commit is just as big a problem. Ideally, there would be a way to bundle the final query and the commit together as well.

We can add special query flag like `QueryCommit` which when present changes behaviour like following:
- query is added to new Batch together with `COMMIT`. BatchResult is saved in `Tx`
- batchresult.Exec() is called and result is returned as result to app
- on `Tx.Commit()` batchresult.Exec() & batchresult.Close() is called and result from Exec it returned to the app

jackc:
[...]

PR #2649, opened 2026-09-08
Add Batch.BeginTx and BatchResults.Tx

timofurrer:
Implements the interface suggested in https://github.com/jackc/pgx/issues/1667#issuecomment-5574730641. Supersedes #2637.

## What this adds

```go
// BeginTx queues the query that begins a transaction with txOptions as the first
// query of b. It must be called before any query is queued.
func (b *Batch) BeginTx(txOptions TxOptions)

// Tx returns the transaction begun by Batch.BeginTx, or nil if the batch did not
// begin one.
func (BatchResults) Tx() Tx
```

```go
b := &pgx.Batch{}
b.BeginTx(pgx.TxOptions{IsoLevel: pgx.RepeatableRead})
b.Queue(fenceCheck, shardID, fenceID).QueryRow(func(row pgx.Row) error { return row.Scan(new(int)) })

br := pool.SendBatch(ctx, b)
tx := br.Tx()
defer tx.Rollback(ctx)
if err := br.Close(); err != nil {
	return err
}
// ... the rest of the unit of work on tx ...
return tx.Commit(ctx)
```

## Design notes

Points that go beyond the sketch in the issue, each covered by a test:

* **`Tx()` is available before `Close`**, so `defer tx.Rollback(ctx)` works exactly as with `Conn.BeginTx`. The transaction must not otherwise be used until `Close` has been called; doing so fails with the usual conn busy error rather than corrupting the connection.
* **A failed begin leaves a closed `Tx`.** After `Close`, if the connection is not in a transaction because the begin query failed or the batch was never sent, the `Tx` is closed and `Rollback`/`Commit` return `ErrTxClosed`. Without this, `Commit` after a failed begin would return nil, since PostgreSQL answers a stray `COMMIT` with a warning and a `COMMIT` tag. A failing later statement leaves the connection in a failed transaction and the `Tx` open, so `Rollback` works as usual.
* **`Pool.SendBatch` hands the connection to the `Tx`.** `poolBatchResults.Close` releases the connection today, and `Conn.Release` destroys a connection that is inside a transaction. For a batch that began a transaction the connection is now released when the `Tx` is committed or rolled back, including when that `Tx` is already closed. `pgxpool.Conn.SendBatch` is unchanged: the caller owns the connection.
* **`Tx.SendBatch` rejects a batch that begins a transaction**, since otherwise two `Tx` values would refer to one server transaction. Easy to drop if you would rather leave this to the server's warning.
* **The begin query is an ordinary queued query.** Its result is the first result of the batch, so positional `br.Exec()`/`br.Query()` callers see it. Callbacks and `Close` are unaffected.
* `Batch.BeginTx` panics when a query is already queued.

`Conn.SendBatch` wraps the existing results in a small `txBatchResults` when the batch began a transaction; none of the per-`QueryExecMode` paths changed.

## Compatibility

`Tx()` is added to the `BatchResults` interface, as anticipated in the issue. Custom implementations, including mocks, must add it. The in-tree implementations in `pgxpool` are updated.

## Not included

The commit side. Queuing `commit` as the last query of a batch sent on the `Tx` already leaves `Tx()` closed after `Close`, so a deferred `Rollback` costs nothing, but `ErrTxCommitRollback` detection and a `Batch` API for it deserve their own change.

## Testing

New tests in `batch_tx_test.go` and `pgxpool/batch_results_test.go` cover commit, rollback, transaction options, a custom commit query, a failing statement after begin, a failing begin, a batch that is never sent, positional results, the panic, the `Tx.SendBatch` guard, and pool connection ownership in the success, failure and failed-begin paths. They run across all `QueryExecMode`s and pass against PostgreSQL 14, 16 and 17 and CockroachDB. The full suite passes against PostgreSQL 17.

## AI disclosure

Per CONTRIBUTING.md: this change was developed with AI assistance (Claude Code, Fable 5.1), including the review of the suggested interface, the implementation and the tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -62,6 +62,21 @@ func (qq *QueuedQuery) Exec(fn func(ct pgconn.CommandTag) error) {
 // unnecessary network round trips. A Batch must only be sent once.
 type Batch struct {
 	QueuedQueries []*QueuedQuery
+
+	txOptions *TxOptions
+}
+
+// BeginTx queues the query that begins a transaction with txOptions as the first query of b, saving the round trip that
+// [Conn.BeginTx] spends on it. It must be called before any query is queued and panics otherwise. Like any queued query,
+// the begin query has a result: it is the first result of the batch. The transaction is available from
+// [BatchResults.Tx] once b has been sent.
+func (b *Batch) BeginTx(txOptions TxOptions) {
+	if len(b.QueuedQueries) != 0 {
+		panic("Batch.BeginTx must be called before any query is queued")
+	}
+
+	b.Queue(txOptions.beginSQL())
+	b.txOptions = &txOptions
 }
 
 // Queue queues a query to batch b. query can be an SQL query or the name of a prepared statement. The only pgx option
@@ -113,6 +128,12 @@ type BatchResults interface {
 	// Close is safe to call multiple times. If it returns an error subsequent calls will return the same error. Callback
 	// functions will not be rerun.
 	Close() error
+
+	// Tx returns the transaction begun by [Batch.BeginTx], or nil if the batch did not begin one. It is available before
+	// Close is called so that a Rollback can be deferred as usual, but the transaction must not otherwise be used until
+	// Close has been called. If the batch fails before the transaction is begun, the transaction is closed by the time
+	// Close returns, and its methods return an error where errors.Is(ErrTxClosed) is true.
+	Tx() Tx
 }
 
 type batchResults struct {
@@ -259,6 +280,10 @@ func (br *batchResults) Close() error {
 	}
 
 	return br.err
+}
+
+func (br *batchResults) Tx() Tx {
+	return nil
 }
 
 func (br *batchResults) earlyError() error {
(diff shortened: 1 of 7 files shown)


jackc:
This review was written by **Codex, OpenAI’s AI coding agent**, after reviewing commit `549f0ce3`.

I found two issues:

1. **[P2] Repeated `Close` accesses a connection after it returns to the pool.** In `txBatchResults.Close`, every call reads `TxStatus()`. After `br.Close()` followed by `tx.Commit()`, another goroutine may acquire that connection. A deferred second `br.Close()` then races with the new borrower’s queries. I reproduced this with the Go race detector. Since `Close` explicitly permits repeated calls, transaction finalization should happen only once.

2. **[P2] Pool acquisition failure returns a nil transaction handle.** If acquisition times out, `Pool.SendBatch` returns `errBatchResults`, whose `Tx()` returns nil even when `Batch.BeginTx` was requested. Consequently, the PR’s example pattern—`tx := br.Tx(); defer tx.Rollback(ctx)`—panics before the caller can inspect the batch error. I reproduced this with an exhausted single-connection pool. Failed transactional batches should retain a closed transaction handle, consistent with the documented behavior when a batch is never sent.

The existing `pgx` and `pgxpool` test suites pass with `-race` against PostgreSQL 18. Targeted reproduction tests confirm both issues.

timofurrer:
@jackc thanks for the review. I've pushed an updated revision. PTAL

PR #2656, opened 2026-09-18
pgconn: support hostaddr in connection strings

### LINKED ISSUE #2655: Mismatch with libpq when hostaddr is specified in connection strings
**Describe the bug**

Excited to see `pgx`'s efforts to improve compatibility with `libpq` connection strings!

But there are still some vital mismatches.

**Expected behavior**

See https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-HOSTADDR

``` shell
$ psql "postgresql://someuser@hostname.not.used:54320/somedb?sslmode=require&host=hostname.for.sni&hostaddr=192.168.1.100"
Password for user someuser:
psql (18.4)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_128_GCM_SHA256, compression: off, ALPN: postgresql)
Type "help" for help.

somedb=>
```

**Actual behavior**

When specified, `hostaddr` is not obeyed and treated as `unrecognized configuration parameter`.

**Version**
 - Go: `go version go1.26.8 linux/amd64`
 - PostgreSQL: `PostgreSQL 18.4 on x86_64-pc-linux-musl, compiled by gcc (Alpine 15.2.0) 15.2.0, 64-bit`
 - pgx: `v5.11.0`

**Additional context**

I have digged a bit in `pgconn/config.go`. It seems to take some time to add new functionalities around `hostaddr`.

Leewwp:
## Summary

`hostaddr` names the address to dial while `host` keeps its meaning for server identity — SNI and certificate verification. pgx did neither: the value fell through to the run-time parameters, so the server rejected it with `unrecognized configuration parameter "hostaddr"`.

```shell
psql "postgresql://someuser@hostname.not.used:54320/somedb?sslmode=require&host=hostname.for.sni&hostaddr=192.168.1.100"
```

This treats `hostaddr` as a connection string field alongside `host` and `port`, and dials it instead of resolving the host name.

## Changes

- `pgconn/config.go`: `hostaddr` is added to `notRuntimeParams`, so `ParseConfig` consumes it instead of forwarding it to the server. `Config` and `FallbackConfig` gain a `HostAddr` field.
- As with `port`, a single address applies to every host, and any other count must match the host count. Otherwise parsing fails with `could not match N host addresses to M hosts`.
- `buildConnectOneConfigs` dials `hostaddr` directly when it is set, skipping name resolution, and keeps `host` as the `originalHostname` used for identity.
- A `hostaddr` is always a TCP connection, so the Unix-domain-socket TLS exemption no longer applies when one is present.

`hostaddr` without `host` is a complete target in libpq: there is no name to resolve and none to present. The default socket directory is therefore not substituted for the missing host name, which would otherwise become the TLS server name — `Config.Host` is left empty. Only a host name the user actually wrote counts, so `PGHOST` and a `host` from a service file still apply.

For the connection string above the parsed config is now:

```
Host="hostname.for.sni"  HostAddr="192.168.1.100"  Port=54320
dial 192.168.1.100:54320   originalHostname "hostname.for.sni"   TLS ServerName "hostname.for.sni"
```

## Tests

- `pgconn/config_test.go`: `TestParseConfigHostAddr` (including the issue's URL), `TestParseConfigHostAddrWithoutHost`, `TestParseConfigHostAddrWithHostFromEnvironment`, `TestParseConfigHostAddrCountMismatch`.
- `pgconn/pgconn_private_test.go`: `TestBuildConnectOneConfigsUsesHostAddrWithoutResolving` asserts the dial address, the retained host name, and that `LookupFunc` is never called when `hostaddr` is set.

## Verification

```
go build ./...                        ok
go vet ./pgconn/                      clean
gofumpt -extra -l pgconn/             clean
go test ./pgconn/ -run HostAddr -v    5 tests pass
go test ./pgconn/                     identical failure set before and after
```

The failures in `./pgconn` are the ones that need a running PostgreSQL server, which this machine does not have. They are present on `master` as well, and the failing test names are identical before and after the change (96 either way).

Fixes #2655


ur4t:
This PR works as expected for the example case from the issue.

